PR #318, opened 2014-12-22
Rename 'Concurrent builds' setting to 'Concurrent jobs'

aprescott:
My understanding is that this setting is actually a value used to control job concurrency, not build concurrency, so this clarifies that by referring to jobs instead of builds.

I realise there are other places this is used, such as `maximum_number_of_builds` in the CLI and elsewhere, but this change was more within my grasp. :)


joshk:
Great spotting!

Thanks for the feedback and PR!

PR #330, opened 2015-02-16
Set page titles relevant to page is loaded

drogus:


app/routes/ssh-key.coffee
```diff
@@ -3,6 +3,8 @@
 `import TravisRoute from 'travis/routes/basic'`
 
 Route = TravisRoute.extend
+  titleToken: 'Ssh Keys'
```
simi: SSH

app/routes/application.coffee
```diff
@@ -12,6 +13,14 @@ Route = TravisRoute.extend
   activate: ->
     @get('stylesheetsManager').disable('dashboard')
 
+  title: (titleParts) ->
+    if titleParts.length
+      titleParts = titleParts.reverse()
+      titleParts.push('Travis CI')
```
simi: What about to reuse `config.defaultTitle` here?

PR #331, opened 2015-02-17
Signin btn

lislis:


app/styles/app/components/buttons.sass
```diff
@@ -0,0 +1,81 @@
...
+  text-decoration: none
+  border: 1px solid #DDD
+.show-more:hover
+  background-color: #40454F
+  color: $white
+  border: 1px solid #40454F
+  cursor: pointer
+
+
+#top #navigation // bad style, we really need to get rid of the ids in the styles
```
simi: wondering why

lislis: Styling with ids makes it super hard to overwrite styles later on or in different contexts. Even in this case I needed to use two ids to overwrite styles for links in the navigation.

PR #332, opened 2015-02-17
remove unnessary(?) things from charmscout script

lislis:
where does the charm key come from?


joshk:
I think it comes from the ENV, @drogus can you confirm this?

simi:
@lislis it is defined at https://github.com/travis-ci/travis-web/blob/master/config/environment.js#L51

joshk:
Soooooo, do we still need this PR?

lislis:
@joshk probably not, since #371 removes it completely

joshk:
Should we close this PR?

lislis:
do eeeeet

PR #334, opened 2015-02-23
Allow cancel if there are cancelable jobs

drogus:


joshk:
Looks amazing to me! Thanks for fixing this! Merge and deploy at will!

PR #335, opened 2015-02-24
Anchor btn to log

lislis:


joshk:
Is everything good to merge in and deploy?

joshk:
We should remove the 'remove log' button, and possibly 'download log' button, when there is no log.

PR #336, opened 2015-02-24
Favicon changing

drogus:
This still needs some work, but it works already.

TODO:

- [x] write some tests
- [x] for now I just generated 1 pixel colored data-uris for states, I need to get proper icons
- [x] test on all the browsers (changing favicon may not work at all on IE and safari, or at least I've heard that's the case)


app/mixins/build-favicon.coffee
```diff
@@ -0,0 +1,36 @@
...
+    images = {
+      red: 'data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAYAAAAfFcSJAAAADUlEQVR42mP4z8DwHwAFAAH/VscvDQAAAABJRU5ErkJggg==',
+      yellow: 'data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAYAAAAfFcSJAAAADUlEQVR42mP4/5ThPwAHrwLktvrIVgAAAABJRU5ErkJggg==',
+      green: 'data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAYAAAAfFcSJAAAADUlEQVR42mOwv276HwAE7wJLzZ3V0gAAAABJRU5ErkJggg==',
+      gray: 'data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAYAAAAfFcSJAAAADUlEQVR42mMoqyj7DwAFMAJkMSbkxgAAAABJRU5ErkJggg=='
+    }
+
+    @setFavicon(images[color])
+
+  setDefault: ->
```
joshk: if we do this it means we need to remember to change it if we change our default favicon, could we just load our favicon instead?

app/mixins/build-favicon.coffee
```diff
@@ -0,0 +1,36 @@
...
+
+  init: ->
+    @faviconManager = new FaviconManager()
+
+    @_super.apply this, arguments
+
+  setFaviconForState: (state) ->
+    color = colorForState(state)
+
+    images = {
```
joshk: You can find all the favicons here: https://github.com/travis-ci/travis-web/tree/master/public

joshk:
I'M SO EXCITED!

app/mixins/build-favicon.coffee
drogus: Changing url doesn't work as good as setting a data uri (it was sometimes changing and sometimes stayed unchanged when I was testing on chrome). I'll check if we can generate data-uris based on images, it should be easy with the tool we use for CSS already.

joshk:
SUPER AWESOME!

BanzaiMan:
:confetti_ball:

PR #343, opened 2015-03-20
Refactor jobs list and add some unit tests for it

drogus:
This is a refactoring of jobs list, it includes:

* changing jobs list and jobs item into components
* adding unit tests for both components
* cleaning up templates a bit


drogus:
@lisa I was wondering if we can get rid of some of the excess markup for icons? Is it because you want to stay compatible with older browsers?

lislis:
@drogus we can get rid of some markup, I guess. I wanted to have one class that defines each icon and then have other elements that need that icon extend the icon class. But when writing the markup it was usually faster for me to just write a span with the icon class.

drogus:
@lislis cool, we can tackle that later

PR #349, opened 2015-04-24
Update Ember to 1.11.3

drogus:


package.json
```diff
@@ -14,31 +14,30 @@
   },
   "repository": "",
   "engines": {
-    "node": "0.10.33"
+    "node": "0.10.32"
```
nikolas: Is there a reason for this downgrade? I'd think it would be fine to use v0.10.38 here.

drogus: I had some problems with compiling dependencies on heroku on newer versions, it's probably not that hard to fix, but I'm too tired to fight with npm any more (upgrading ember-cli and ember versions is usually 80% of the time for npm issues and 20% for actual code changes, or at least it feels like that).

PR #358, opened 2015-06-10
Removed .DS_Store from index and added it to .gitignore

wbrefvem:


GrahamCampbell:
:+1:

GrahamCampbell:
That file does NOT belong in there. It's such a common error people make. That file belongs in your global system gitignore file.

PR #360, opened 2015-06-16
System files do not belong in the .gitignore file

GrahamCampbell:


drogus:
I'm sorry, but I don't see any problem with `.DS_Store` being there. Yes, it's a system file, but it doesn't hurt to have it in `.gitignore`.

PR #368, opened 2015-07-06
bump node version on travis

tchak:
It will remove deprecation warnings during build time


drogus:
After updating to the newest version of ember-cli it suddenly started working on heroku on 0.12.x, but I already changed it in master. Thanks for the contribution anyway!

PR #375, opened 2015-07-17
Handle job:queued events

svenfuchs:
Add `job:queued` as a known Pusher event.


svenfuchs:
Fairly sure the CI failure doesn't come from this change.

PR #377, opened 2015-07-18
Language setting matches outcome

leksak:


BanzaiMan:
Thanks for opening this PR, but this PR is incorrect.

It is true that the document is wrong; `C` is actually normalized to `c`, so the build runs with the correct script. But `language: ruby` correctly runs the job as a Ruby script, so it doesn't work as an example here.

A good example instead is something like `language: Erlang` which would cause an Erlang job to run on a Ruby image.

BanzaiMan:
Do you mind opening a new PR with more appropriate change? Thanks!

leksak:
I'm afraid I don't understand (not unsuprising as I have yet to progress far past the getting-started page). Let me see if I understand you correctly, today the documentation states the following (verbatim),

"If you set language: ``C``, for example, your project will be considered a Ruby project."

The setting is then normalized to "language: ``c``", but this setting (as per your explanation) would have no affect on the following statement? Namely that the project is still considered a Ruby project?

You say the following,
A good example instead is something like ``language: Erlang`` which would cause an Erlang job to run on a Ruby image.

And so I do not understand what the more appropriate change would be.

Kind regards,
leksak

PR #379, opened 2015-07-23
received state animation

lislis:
After toying around with svg animations for the received (and other) state I switched to css animations as they're not as performance heavy and work better considering cross browser.
This is only for the jobs view and does not effect any other received state icons and is, in theory, mergable.


joshk:
Awesome! Would you mind adding an animated screenshot here too?

lislis:
okwsz

PR #388, opened 2015-08-10
No builds/ not activated

lislis:
Adding a notice on repos that are not activated and on activated repo without builds.
Trigger build via API worked at one point but broke, so I took it out and just link to the docs instead. So this is shippable :)


joshk:
Sorry, could you add a screenshot as well?

lislis:
YES!
Not active
screen shot 2015-08-11 at 13 23 17

No Builds
screen shot 2015-08-11 at 13 23 27

Builds but no PRs

screen shot 2015-08-11 at 13 23 39

joshk:
This is awesome!

Thank you!

lislis:
I just noticed that repository should be lower case in no-pr as well

PR #391, opened 2015-08-17
Fixes for notactive

lislis:


app/styles/app/layouts/missing-notice.sass
```diff
@@ -1,9 +1,14 @@
...
 
-  a
+  p a
+    color: $grey-dark
+    text-decoration: underline
+
+  a.button
     font-size: 14px
 
+
```
GrahamCampbell: ?

lislis: What's the question?

GrahamCampbell: Why adding the extra newline?

lislis: Ah, my bad. Didn't realize that. :+1:

PR #392, opened 2015-08-25
Authenticate owner request

lislis:
@drogus so I tried to hack the auth header in there, but I still get a 404. What am I doing wrong?


drogus:
The problem is that `$.get` won't accept headers (if you look at the network tab, you'll see that it's appended to the URL as a query param, like `?headers=....`). Changing it to `$.ajax` should fix things.

There is also another place like that: https://github.com/travis-ci/travis-web/blob/dd09b105ee483148e450cc6204160b5df5d163e2/app/routes/owner/repositories.coffee#L13

PR #393, opened 2015-08-25
Fix account header name

lislis:
I basically get the name from the hook list, but that's not very nice. There should be a way to get it from the sidebar. @drogus halp!


drogus:
The problem is that `$.get` won't send headers properly (if you look at the network tab, the header will be appended as a query param, like `?headers=....`). If you use `$.ajax` instead of `$.get`, it should work properly. There is also another similar call here: https://github.com/travis-ci/travis-web/blob/dd09b105ee483148e450cc6204160b5df5d163e2/app/routes/owner/repositories.coffee#L13

drogus:
In the `app/templates/account.hbs` template, we have access to the currently loaded account, because it's set as a model in the route here: https://github.com/travis-ci/travis-web/blob/dd09b105ee483148e450cc6204160b5df5d163e2/app/routes/account.coffee#L16

So you could do something like:

```coffee
accountName: (->
  @get('model.name')
).property('model.name')
```

The other problem could be that not everyone has a name set up, so I would also add a fallback for login as we do in other places:

```coffee
accountName: (->
  @get('model.name') || @get('model.login')
).property('model.name', 'model.login')
```

PR #395, opened 2015-09-25
Change Rst ro RST

Ladsgroup:
RST is abbreviation of reStructuredText.
Using Rst is plain wrong, it would be good to use RST or reStructuredText.


lislis:
Hey @Ladsgroup, thank you so much for the PR!

PR #401, opened 2015-10-15
Broadcasts 

lislis:
So, this is kind of working.
Getting broadcasts from v3, filtered by expiration date.
When you click on the 'x'  save the id in localStorage and then the broadcast won't be shown anymore.
The only thing is that this doesn't update the list of broadcasts (refresh needed). @drogus a hint for me?
Edit: apparently I also lost some icons along the way....


app/controllers/top.coffee
```diff
@@ -12,10 +17,59 @@ Controller = Ember.Controller.extend
...
+              Ember.Object.create(broadcast)
+            ).reverse()
+          
+          if array.length
+            broadcasts.set('lastBroadcastStatus', array[0].category)
+        broadcasts.set('content', array)
+        broadcasts.set('isLoading', false)
+
+      broadcasts
+  ).property('broadcasts')
```
drogus: This should probably not depend on itself, I think that it's better to only depend on `auth.signedIn` - if this changes, we should probably refresh the list.

lislis: I don't get the "depend on itself". What do you mean?

PR #406, opened 2015-10-26
toggle token on profile page

lislis:


app/templates/profile/accounts.hbs
```diff
@@ -9,7 +9,26 @@
...
+
+        {{#if isVisible}}
+          <strong>{{auth.currentUser.token}}</strong>
+        {{/if}}
+        <a {{action 'tokenVisibility'}} class="profile-token-toggle tooltip--profile {{if isVisible 'is-visible'}} dropdown">
+          <span class="tooltip-bubble">
+            {{#if isVisible}}hide token{{else}}show token{{/if}}
+          </span>
+          <span class="icon-eye">
+            <svg x="0px" y="0px" viewBox="2 -1 17 17" width="20" height="18" xml:space="preserve">
```
drogus: I think we should move svgs to helpers or components to not be so verbose in the templates. Maybe even with the span?

lislis: Yea, makes sense. Rather a helper or a component? I used components before especially if I needed to change the svg depending on data. But I don't know if that's the right way?

drogus: I think that component will be easier to use, ie. you can put a template there. Helper is probably more lightweight, but I doubt it's worth to use it here (unless we need hundreds of these svgs or something like that ;) )

PR #408, opened 2015-11-13
add osx_image to keys-map

lislis:
screen shot 2015-11-13 at 13 12 46
Looks like this

### DIFF
--- a/app/utils/keys-map.coffee
+++ b/app/utils/keys-map.coffee
@@ -26,6 +26,7 @@ languageConfigKeys = {
   lein:        'Lein'
   compiler:    'Compiler'
   crystal:     'Crystal'
+  osx_image:   'Xcode'
 }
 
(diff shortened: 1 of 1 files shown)


simi:
Is that compatible with http://www.apple.com/legal/intellectual-property/guidelinesfor3rdparties.html?

lislis:
@simi I believe so. Do you think otherwise?

simi:
I don't know. I had an idea to ask and ensure that's in your mind also.

lislis:
The guideline says to only use Apple product names when the software is actually using it or is compatible with it. I think that's the case. Plus, 'Xcode' is written the way they want it to be written :D 
Thanks for checking in on that anyway :+1:

simi:
:+1: OK, thanks for clarification.

PR #409, opened 2015-11-17
Remove deprecations

lislis:
Edit: from my side, this is also good to go :)


drogus:
I'm closing this, because it's opened against web-on-v3 (which was a right move,  I just didn't manage to merge it before merging web-on-v3 to master), I'll open another one against master.

PR #417, opened 2015-12-09
Update the picture and title for Henrik

henrikhodne:
Looks like the image has to be modified a bit to be made circular, and shrunk a little bit.


henrikhodne:
Actually, it looks like the image is displayed correctly as-is:

2015-12-09-17-49-52_255x373

lislis:
Jupp, css should take care of the border-radius and stuff now.

PR #418, opened 2015-12-10
Mobile log

lislis:
for https://github.com/travis-pro/team-teal/issues/618

Edit: This is working as good to merge. It's not the "fullscreen" solution discussed in the ticket, but as close as I can get right now. I'd like to revisit this in the future

### DIFF
--- a/app/components/log-content.js
+++ b/app/components/log-content.js
@@ -224,7 +224,7 @@ export default Ember.Component.extend({
     },
 
     toggleLog() {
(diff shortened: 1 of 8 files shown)


lislis:
rebased!

drogus:
I merged it and I like the implementation for its simplicity, but in the future we should probably avoid fetching and rendering the log in case it's not displayed. I'm not sure how much work would it be, though, so it's better to have this than nothing.

lislis:
ok!

PR #423, opened 2016-01-01
Author/Committer Avatar Bug in Build List

thriqon:
When commits have an author different from the committer, the association between avatar and name in the builds list is wrong ("authored" by Josh Kalderimis, committed by me):

old-state

For reference, this build can be found here: [/thriqon/travis-demo-rep/builds](https://travis-ci.org/thriqon/travis-demo-rep/builds).

This only occurs in the build list, not in the build detail view.

With this PR, the calculation is moved to the model and reused in both instances. The third commit additionally drops some unneeded imports of `gravatarImage` which I found while scanning for more avatar calculations. After the change, the build overview looks like this:

new-state

### DIFF
--- a/app/components/branch-row.coffee
+++ b/app/components/branch-row.coffee
@@ -1,5 +1,4 @@
 `import Ember from 'ember'`
-`import { gravatarImage } from 'travis/utils/urls'`
 `import { githubCommit as githubCommitUrl } from 'travis/utils/urls'`
 `import TravisRoute from 'travis/routes/basic'`
 `import config from 'travis/config/environment'`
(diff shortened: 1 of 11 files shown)


drogus:
Thank you!

I merged this manually, cause I wanted to do one cosmetic change (rename `avatarUrlOrGravatar` to just `avatarUrl`) and I needed to remove an addition of committer to builds list. It was clearly a bug to display author's avatar and committer name, but we intentionally only wanted to display one avatar and name there. In order to do that we would have to add information on which person is a commiter and which person is an author and I think it wouldn't fit nicely there. That said, I'll discuss this with our designer. Thanks again for the contribution.

drogus:
One correction - I noticed that I made a mistake, ie. `authorAvatarUrl` is already a property of the model, so I reverted this change and merged your original commit.

joshk:
This is really cool! Thank you!

@saltinejustine @drogus could you please review this PR?

joshk:
Sorry, you totally did this before I posted my comment!
Silly me :)
High 5!

PR #424, opened 2016-01-04
add aria-labels to jobs-item

lislis:


joshk:
screenshot?

lislis:
@joshk there is no visual difference, it's just an attribute for screenreaders to use
screen shot 2016-01-04 at 14 04 42

I'll keep an eye on these things in the future, we should have this in all the templates imo

joshk:
ahhhhhh, sweet!

lislis:
rebased!

PR #426, opened 2016-01-05
add link to docs on requests if branch is gh-pages

lislis:
screen shot 2016-01-05 at 11 41 24
screen shot 2016-01-05 at 11 41 26


joshk:
:+1:

lislis:
rebased!

app/components/requests-item.js
```diff
@@ -37,6 +37,10 @@ export default Ember.Component.extend({
   message: function() {
     var message;
     message = this.get('request.message');
+
+    if (message === 'github pages branch') {
```
drogus: I think it would be better to create a property `isGHPages`, which depends on `message`. Setting other properties inside other properties is generally discouraged as in such case you depend on `message` being called to have `isGHPages` correctly set.

PR #433, opened 2016-01-13
Update Ember.js to 2.2.0 and Ember Data to 2.3.0

drogus:


drogus:
This PR also contains commits from #432, it will show only update commits once I merge #432.

PR #435, opened 2016-01-14
Leverage <time> element

jasonkarns:
Sample of using the `time` element instead of `abbr` for displaying timestamps.

The `time` element can also represent durations. So the sample used here is suboptimal. Rather than the duration time lising the `lastBuildStartedAt` time in the `datetime` attribute, it ought to be a [valid `duration` value](https://www.w3.org/html/wg/drafts/html/master/infrastructure.html#valid-duration-string). However, I didn't see any existing helpers for formatting according to a machine-readable duration value.

Ideally, the `time` element would be used in other areas where times/durations are displayed. This pr is meant as a starting point for discussion.

### DIFF
--- a/app/templates/components/build-header.hbs
+++ b/app/templates/components/build-header.hbs
@@ -67,17 +67,17 @@
   <ul class="list-icon">
     <li title="{{format-duration elapsedTime}}" class="commit-stopwatch">
       <span class="icon-stopwatch" aria-hidden="true"></span>
-      <span class="label-align">Elapsed time {{format-duration elapsedTime}}</span></li>
+      <span class="label-align">Elapsed time <time datetime="PT{{elapsedTime}}S">{{format-duration elapsedTime}}</time></span></li>
     {{#unless isJob}}
       {{#if item.isMatrix}}
         <li title="{{format-duration item.duration}}" class="commit-clock">
           <span class="icon-clock" aria-hidden="true"></span>
-          <span class="label-align">{{#if item.isFinished}}Total time{{else}}Running for{{/if}} {{format-duration item.duration}}</span></li>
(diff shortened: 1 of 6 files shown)


jasonkarns:
It appears that most `duration` properties are stored as integer seconds representing the duration. Which means the durations can be properly marked up using the duration microsyntax from ISO 8601. Additional instances of build times (startedAt, finishedAt, duration) have been marked up using `time` with appropriate `datetime` attributes.

lislis:
I like the idea! Let me read up on the time element a bit and I'll get back to you.

drogus:
Thanks for the contribution. It looks good with one exception - I don't think that `duration` should be represented as `<time>`, because according to [description on MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/time) the `<time>` element "represents either a time on a 24-hour clock or a precise date in the Gregorian calendar (with optional time and timezone information).".

jasonkarns:
The MDN description is lacking. [Per the spec](https://html.spec.whatwg.org/multipage/semantics.html#the-time-element:valid-duration-string), duration is a valid value.

drogus:
Thanks for the link, sounds good!

Could you rebase? I'll merge right away.

jasonkarns:
rebased for great justice

jasonkarns:
I should note that I made the changes via markup only and haven't visually inspected anything regarding layout. For the most part, most of the `time` elements were added so it shouldn't affect anything. However, in a couple places, I swapped the existing `span` for `time`. All the classnames remained, so it *should* affect anything. But if the styles were dependent on those elements being `span`s and not simply leveraging the classnames, there may be issues.

drogus:
I think we don't usually use element types, but I'll check it visually after merging, thanks for the work on this.

PR #436, opened 2016-01-14
Fix grammatical error for finished job duration

jasonkarns:
While a job is running, on the job status view, the duration is labeled: "Running for x sec". When the job finishes, 'Running' is swapped for 'Total time'. But "Total time for x sec" doesn't make any sense. the " for" should be part of the conditional with "Running"


lislis:
Hey @jasonkarns! Thank you for this!
awesome

PR #437, opened 2016-01-19
make a user-avatar component

lislis:
delete some, add some, fix some tests


app/components/user-avatar.js
```diff
@@ -0,0 +1,40 @@
...
+
+  tagName: 'span',
+  classNames: ['avatar'],
+
+  gravatarUrl: function() {
+    if (!this.get('isAuthor') && !this.get('isComitter')) {
+      if (this.get('user.type') === 'organization') {
+        return this.get('user.avatarUrl');
+      } else {
+          return gravatarImage(this.get('user.email'), 36);
```
drogus: Can't we use `user.avatarUrl` here as well? I thought that we now save `avatarUrl` from GitHub for all type of accounts.

drogus: I've checked it myself and it seems that we don't even have `avatar_url` column for users, I thought it was fixed at some point.

drogus:
I really like the idea behind this, but I'd change the implementation a little. There is a lot of logic involved in figuring out what we should display for an avatar. I think that a better approach would be to pass url instead of user object (which is also a bit confusing, because in this case a user object can be also commit.

For example after you add an `avatarUrl` property to the user, which would do gravatar id calculations, you could do component that could be used like:

```hbs
{{user-avatar url=commit.authorAvatarUrlOrGravatar name=commit.authorName}}
{{user-avatar url=user.avatarUrl name=user.fullName}}
{{user-avatar url=account.avatarUrl name=account.name}}
```

lislis:
@drogus that does make a lot more sense. I'll update this

lislis:
@drogus wdyt? If that looks good then I'll rebase.

drogus:
Looks good apart from an `avatarUrl` property on the user model. At the moment there's no such property, so you need to add it and make it calculate gravatar url based on e-mail.

lislis:
Updated!

app/models/user.js
```diff
@@ -132,5 +133,9 @@ export default Model.extend({
     user = JSON.parse(this.get('sessionStorage').getItem('travis.user'));
     user[name.underscore()] = this.get(name);
     return this.get('sessionStorage').setItem('travis.user', JSON.stringify(user));
-  }
+  },
+
+  avatarUrl: function() {
+    return gravatarImage(this.get('email'), 36);
+  }.property()
```
drogus: this should depend on `email`, so if `email` ever changes it will get recomputed. That's probably not really that important at the moment, because email won't rather change without reloading, but it will be more future proof.

lislis:
updated!

drogus:
Looks good!

lislis:
ahahaha, now I didn't rebase :P  
But thanks :)  
wheee

Krinkle:
The pseudo-avatar and real avatar aren't aligned properly. There is a white edge along the bottom. 

(Chrome 51.0.2704.106 stable, OSX, Retina)

screen shot 2016-07-04 at 18 04 46

lislis:
Hey @Krinkle thanks for spotting and reporting this!
I'll open an internal ticket.

lislis:
Hey again @Krinkle!
I just deployed a fix for the avatar bug. Please let me know if you experience any further issues.

Krinkle:
@lislis Thanks, looks great!
screen shot

PR #444, opened 2016-01-31
Remove initializer deprecation warnings

clekstro:
The previous initializer API was deprecated in Ember 2.1 (See
http://emberjs.com/deprecations/v2.x/), so this updates the initializer
signatures to accomodate that change.
Also, remove the registered deprecation from the workflow.


drogus:
Thanks!

PR #449, opened 2016-02-02
Enterprise ui fixes

lislis:


app/routes/home-pro.js
```diff
@@ -1,5 +1,12 @@
 import TravisRoute from 'travis/routes/basic';
+import config from 'travis/config/environment';
 
 export default TravisRoute.extend({
-  needsAuth: false
+  needsAuth: false,
+
+  activate: function() {
```
drogus: I think that a better place to fix this behaviour would be to change our `Location` handler. At the moment we handle home pages with `if(config.pro)` check: https://github.com/travis-ci/travis-web/blob/master/app/utils/location.js#L21-L31

I think that if we avoid logic for homepage for enterprise altogether, it will result in cleaner code, so we could change [this statement](https://github.com/travis-ci/travis-web/blob/1d0fb6468e83f6b99d16f15e1bd2a095e67c6380/app/utils/location.js#L22) to `if(location.pathname === '/' && !config.enterprise)` and [this statement](https://github.com/travis-ci/travis-web/blob/1d0fb6468e83f6b99d16f15e1bd2a095e67c6380/app/utils/location.js#L37) to `if (!config.enterprise && (logicalPath === '/repositories' || logicalPath === '/home' || logicalPath === '/home-pro'))`. With these conditions we would run regular URL handling for enterprise, so no homepage would be displayed initially.

app/templates/account.hbs
```diff
@@ -2,25 +2,21 @@
...
-      <a class="btn btn-activated" href={{billingUrl}}>
-        Subscription active!
-      </a>
-    {{else}}
-      {{#if subscribeButtonInfo.education}}
-        <a class="btn btn-activated" href={{billingUrl}}>
-          Educational account!
-        </a>
+  {{#unless config.enterprise}}
+    {{#if config.billingEndpoint}}
```
drogus: All of the places that already have checks for things like `billingEndpoint` could be fixed by not setting the defaults in `config/environment.js`: https://github.com/travis-ci/travis-web/blob/1d0fb6468e83f6b99d16f15e1bd2a095e67c6380/config/environment.js#L51

In order to not set these things we could do `if (process.env.TRAVIS_PRO && !process.env.TRAVIS_ENTERPRISE) {` [here](https://github.com/travis-ci/travis-web/blob/1d0fb6468e83f6b99d16f15e1bd2a095e67c6380/config/environment.js#L40).

lislis:
@drogus can you have a quick look if it's cool now?

joshk:
How's this looking for merging?

lislis:
I think it's good, but Piotr didn't have time to look at it. I think I'll just merge it. It works, so there is that :D

drogus:
Sorry, I have totally missed that earlier. It looks good! (better later than never I guess ;) )

PR #451, opened 2016-02-04
Adding some more "about us" info to /about

solarce:
I got a DM from a friend interested in the job asking if we had
an "about page" that was more than just the headshots, which he
thought were super cool. I took the language from what we've been
including in jobs ads, as a start. I thought I'd submit it for
tweaking and review by those with more design sense than me ;)

### DIFF
--- a/app/templates/team.hbs
+++ b/app/templates/team.hbs
@@ -12,7 +12,20 @@
           {{team-member member=member}}
         {{/each}}
       </ul>
-    </div>
+      <header class="section--center team-header">
+        <h1 class="h1--green">Travis CI is:</h1>
+        <p class="text-big">A continuous integration platform.</p>
+        <p class="text-big">We run people's tests, deploy their code and help them be confident about their work.</p>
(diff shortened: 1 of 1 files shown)


saltinejustine:
Hey @solarce. I super appreciate you going whipping this draft together, it's a good idea. It's such a good idea, in fact, we're already working on it :) However the copy is coming from Mr. @roidrage himself.

We'll be putting up a more condensed version of the [Company Values](https://github.com/travis-pro/how-we-work/issues/8) we've been working on for months and the plan is to put it on a separate page including a team group photo (probably from BCN offsite) and leave the team individuals on another page entirely.

I'll let everyone in Europe have a looksie at this even though we won't be able to merge this pull request, and perhaps we can add that page sooner rather than later.

lislis:
Would it be ok with y'all if I close this for now? As explained this will come 🔜 ish

saltinejustine:
Yes — let's close this. It'll happen when the values are ready

PR #458, opened 2016-02-15
Remove click handler overriding default link behavior

clekstro:
An [issue](https://github.com/travis-ci/travis-ci/issues/5181) was reported in
which the repository sidebar links were not functioning as expected. I've
tracked it down to the fact that a click handler was registered and
automatically performing a redirect to the repo route.

As reported, this breaks CMD+Click behavior (as the click handler was still
called and navigating when the user did not expect it). It also breaks normal link
behavior as it would take the user to a URL they did not intend to visit.

This commit simply removes that click handler. I checked the previous
coffee version of the file to gain more context on why it was initially
added, but unfortunately, it was simply added when the original
component was created, so I'm not sure what its use was.

If there is known behavior that this would break, we should probably add some
additional test coverage around it, and I could take another crack at a fix.

### DIFF
--- a/app/components/repos-list-item.js
+++ b/app/components/repos-list-item.js
@@ -23,10 +23,5 @@ export default Ember.Component.extend(Polling, {
         scrollTop: 0
       }, 200);
     }
-  },
-
-  click() {
-    this.scrollTop();
-    return this.get('routing').transitionTo('repo', this.get('repo.slug').split('/'));
   }
 });
--- a/app/mixins/scroll-reset.js
+++ b/app/mixins/scroll-reset.js
@@ -1,0 +1,8 @@
+import Ember from 'ember';
+
+export default Ember.Mixin.create({
+  activate: function() {
+    this._super(...arguments);
+    window.scrollTo(0,0);
+  }
+});
--- a/app/routes/repo.js
+++ b/app/routes/repo.js
@@ -1,8 +1,9 @@
 import TravisRoute from 'travis/routes/basic';
 import Repo from 'travis/models/repo';
+import ScrollResetMixin from 'travis/mixins/scroll-reset';
 import Ember from 'ember';
 
-export default TravisRoute.extend({
+export default TravisRoute.extend(ScrollResetMixin, {
   store: Ember.inject.service(),
 
   titleToken(model) {


lislis:
Hey @clekstro! Thank you for looking into this. 
This click handler was a fix where users complained that if they clicked on a link in the sidebar the page would transition but not scroll up to where the information is at (unlike eg on the build history). Any ideas on how to achieve that?

clekstro:
I think the easiest way would be to hook into the destination routes' `activate` hook, which is invoked whenever that route is entered. I've generally created a mixin in the past so it can be used wherever it's needed.  Something like this:

```js
export default Ember.Mixin.create({
  activate() {
    this._super(...arguments);
    window.scrollTo(0, 0); // or desired way of scrolling
  }
});
```

Then just extend your routes as normal. If we know exactly which routes this should apply to (looks like the `repo` route from the code, but could be more), I'd be happy to make the necessary changes.

lislis:
awesome

clekstro:
Hey @lislis, thanks for getting this merged in! I was about to comment again on [this issue](https://github.com/travis-ci/travis-ci/issues/5181) but don't believe the changes from this PR are live yet.  According to the logs [here](https://travis-ci.org/travis-ci/travis-web/jobs/109616665#L1599), it looks like the deploy was a success, but I'm not seeing the changes I would expect to (and see locally with current master). Is there a way to confirm (e.g. via a git SHA) what version of the code is live in production?

lislis:
@clekstro you're right, the changes are not live yet. Busy day :D Gimme a few minutes and they shall be up!

clekstro:
No rush! Thought they were automatic and something was wrong ;)

lislis:
and :ship:ed

drogus:
I think that this will not work properly, because `activate` hook is fired only when the route is entered from another route. If you transition to the same route, just with other arguments (for example you transition from `/travis-ci/travis-web` to `/travis-ci/travis-core`, the hook won't fire). A better place to do it would be the `beforeModel` hook as it is guaranteed to run on each model change (initial and all subsequent changes).

clekstro:
@drogus Good catch. I would have expected url changes to fire that event, but based on the docs it's clearly more nuanced than that. Do you want to push a commit directly to address this or should I submit another PR?

drogus:
@clekstro I pushed the change in 55005b1, thank you for all the work on this!

clekstro:
@drogus np!

PR #459, opened 2016-02-16
update footer

lislis:
screen shot 2016-02-16 at 19 13 02
looks like this
@saltinejustine I went for 40% opacity for the logo as 30% looked too light. but we can of course change that


saltinejustine:
@lislis A++ looks great to me!!

PR #460, opened 2016-02-16
Remove legacy binding deprecation warnings

clekstro:
This removes the warnings and should clean up the test output (especially in CI) quite a bit.


lislis:
dancing

PR #462, opened 2016-02-17
Add R language key for matrix column

jimhester:
I believe this should fix the "no language set" message seen at https://travis-ci.org/jimhester/travis-test/builds/109884131

Let me know if there is something else which needs to be done


lislis:
@jimhester that is totally true! Thank you for fixing this!  
party

PR #464, opened 2016-02-23
Cronjobs

Nef10:
This add the GUI to travis-ci/travis-api#225

- Cron jobs can be configured in settings of a repo 
- A tab next to Pull Requests which shows only cron builds
- Cron jobs are included in the Build History tab
- Cron jobs are included in the branch-row 

Settings:
screenshot 2016-02-23 14 32 26


app/models/repo.js
```diff
@@ -131,17 +131,31 @@ Repo.reopen({
...
+      repository_id: id
+    }, function(cron) {
+      return cron.get('branch.repoId') === id;
     });
-    return builds;
   }.property(),
 
+  sortedCrons: Ember.computed.sort('crons', function(a, b) {
+    return a.get('branch.name') > b.get('branch.name');
+  }),
```
Nef10: Unfortunately this currently only work after e.g. adding a cron job. We think it could be because `branch` will be loaded async but we did not figure out if or how we could use a `then` inside the sort method.

app/adapters/cron.js
```diff
@@ -0,0 +1,44 @@
+import Ember from 'ember';
+import ApplicationAdapter from 'travis/adapters/application';
+
+export default ApplicationAdapter.extend({
+
+  findRecord(store, type, id, record) {
+    return this.ajax(this.urlPrefix() + '/cron/' + id, "GET");
```
drogus: I think that should be the default behaviour for adapter (and the same goes for all the other methods here), have you tried to run the code without specifying these?

app/components/add-cron-job.js
```diff
@@ -0,0 +1,59 @@
...
+      repo_id = this.get('branches').toArray()[0].get('repoId');
+      branch = this.get('selectedBranch') ? this.get('selectedBranch') : this.get('branches').toArray()[0];
+
+      store.filter('cron', {
+        repository_id: repo_id
+      }, function(c) {
+        return c.get('branch.repoId') === repo_id && c.get('branch.name') === branch.get('name');
+      }).then(function(existing_crons){
+        if(existing_crons.toArray()[0]){
+          store.unloadRecord(existing_crons.toArray()[0]);
```
drogus: Why is it needed to unload records here?

app/components/cron-job.js
```diff
@@ -0,0 +1,30 @@
+import Ember from 'ember';
+
+export default Ember.Component.extend({
+  classNames: ['settings-cron'],
+  isDeleting: false,
+  actionType: 'Save',
+  store: Ember.inject.service(),
+
+  disable_by_build: function(key) {
```
drogus: We use camel case for function and variable names (or at least we try ;) )

app/adapters/cron.js
Nef10: the default is `crons` instead on `cron`

app/components/add-cron-job.js
Nef10: Only one cronjob can exist per branch. If you add a new one, the API will delete the old one for this branch (if existing). Here we unload the old cron (which will be deleted in the API), so the autorefesh works.

drogus: Ah, makes sense.

aioue:
+1

app/adapters/branch.js
```diff
@@ -0,0 +1,16 @@
+import Ember from 'ember';
+import ApplicationAdapter from 'travis/adapters/application';
+
+export default ApplicationAdapter.extend({
+
+  query(store, type, query) {
+    var repo_id = query['repository_id'];
+    delete query['repository_id'];
```
backspace: This is more a stylistic thing, but note that you could use `query.repository_id` here, it’s more conventional.

app/adapters/cron.js
```diff
@@ -0,0 +1,44 @@
...
+    return this.ajax( this.urlPrefix() + '/v3/repo/' + repo_id + '/crons', "GET", query);
+  },
+
+  ajaxOptions(url, type, options) {
+    var hash = this._super(url, type, options);
+
+    hash.headers['accept'] = 'application/json';
+    hash.headers['Travis-API-Version'] = '3';
+    hash.headers['Content-Type'] = 'application/json';
+
```
backspace: This header situation is also specified in `adapters/v3`, maybe this could inherit from that, or they could share these headers somehow? But it only happens twice that I can see, and like [Sandi Metz says](http://www.sandimetz.com/blog/2016/1/20/the-wrong-abstraction):
> prefer duplication over the wrong abstraction

So we can come back to this for refactoring once we have proper tests.

app/components/add-cron-job.js
```diff
@@ -0,0 +1,59 @@
+import Ember from 'ember';
+import config from 'travis/config/environment';
+
+export default Ember.Component.extend({
+  classNames: ['form--cron'],
+  store: Ember.inject.service(),
+
+  reset() {
+    return this.setProperties({
```
backspace: Is this `return` superfluous? Not worth blocking over, but having a return suggests it’s a promise or something, which it isn’t.

app/components/add-cron-job.js
```diff
@@ -0,0 +1,59 @@
...
+
+  actions: {
+    save() {
+      var cron, self, store, repo_id, branch;
+
+      if (this.get('isSaving')) {
+        return;
+      }
+
+      self = this;
```
backspace: A nice thing about the default Ember CLI configuration is that it comes with Babel, which lets us use things like the “fat arrow” syntax that preserves `this` in a callback. So below you could have `.then((existing_crons) { … this.set ` and be addressing the proper `this`. I see `self`s scattered around the codebase so this isn’t worth holding up a merge, but it’s something you can take advantage of in the future!

app/components/add-cron-job.js
```diff
@@ -0,0 +1,59 @@
...
+          self.set('isSaving', false);
+          return self.reset();
+        });
+      });
+    }
+  },
+
+  intervals: function() {
+    return ['monthly', 'weekly', 'daily'];
+  }.property()
```
backspace: I should have said this earlier, but anyway, I just joined this project, so many of my comments thus far have been more general Ember Way things that aren’t necessarily backed up by the codebase. Another trend within Ember world is that people are moving away from the “prototype extensions” syntax and toward a more plain Javascript syntax. It’s a bit more verbose, but also more explicit. So this could be rewritten:

```
intervals: Ember.computed(function() {
  return …
})
```

_However_, does this even need to be a computed property? Could it just be this?

`intervals: ['monthly', 'weekly', 'daily']`

app/components/branch-row.js
```diff
@@ -6,15 +6,15 @@ import config from 'travis/config/environment';
 export default Ember.Component.extend({
   routing: Ember.inject.service('-routing'),
   tagName: 'li',
-  classNameBindings: ['build.last_build.state'],
+  classNameBindings: ['branch.last_build.state'],
```
backspace: I’m generally trepidatious regarding giant renames without proper tests to ensure functionality has been preserved! But you do seem to have covered all the uses.

app/components/cron-job.js
```diff
@@ -0,0 +1,30 @@
...
+    }
+    return value + 'if no new commit after last cron build';
+  }.property('cron.disable_by_build'),
+
+  actions: {
+    "delete": function() {
+      if (this.get('isDeleting')) {
+        return;
+      }
+      this.set('isDeleting', true);
```
backspace: If you do more Ember work in the future, I recommend checking out [`ember-concurrency`](http://ember-concurrency.com/) which is a beautiful addon that simplifies a lot of this ceremony around limiting concurrent tasks and much much more.

app/controllers/builds.js
```diff
@@ -13,7 +13,7 @@ export default Ember.Controller.extend({
     var id, number, type;
     id = this.get('repo.id');
     number = this.get('builds.lastObject.number');
-    type = this.get('tab') === "builds" ? 'push' : 'pull_request';
+    type = this.get('tab') === "builds" ? 'push' : this.get('tab').substr(0, this.get('tab').length-1);
```
backspace: This change is inscrutable to me. My guess is that it’s about stripping a pluralisation? In cases like this, I would make it more explicit at the expense of being more verbose:

```
const tabName = this.get('tab');
const singularTab = tabName.substr(0, tabName.length);
```

Then people looking at the code don’t have to mentally run it to figure out what it’s for. This is my preferred approach for semi-opaque snippets of code that aren’t worth a block comment.

backspace:
I got hung up on the `models/repo` changes, so I made a [deployment of this pull request](https://hpi-cron.test-deployments.travis-ci.org/). It’s the same environment as travis-ci.org. Hopefully that works for testing things out?

backspace:
ACTUALLY maybe that deployment is broken, so I put it on [staging.travis-ci.org](https://staging.travis-ci.org/).

app/models/repo.js
```diff
@@ -130,18 +130,47 @@ Repo.reopen({
...
+    });
+    array = ExpandableRecordArray.create({
+      type: 'build',
+      content: Ember.A([])
+    });
+    array.load(builds);
+    id = this.get('id');
+    array.observe(builds);
+    return array;
+  }.property(),
```
backspace: I’m pausing here because I need to understand better what is happening here and it’s the end of the day for me! I’ll return.

backspace: My understanding is that this code sets up a pseudo-relationship between a `repo` and its `crons`. Normally one would do this with `hasMany` but that appears to not be possible. So this technique observes all known `build`s within the Ember Data store and selects the ones that belong to this `repo`.

Given that `filter` no longer exists on the Ember Data store, it looks like [the Ember Way is to use computed properties with `peekAll`](https://github.com/ember-data/ember-data-filter#i-dont-want-to-enable-this-feature-due-to-its-heavy-churn). I imagine that would look like this:

```
builds: Ember.computed(function() {
  return this.store.peekAll('build');
}),

crons: Ember.computed('id', 'builds.@each.repo', function() {
  return this.get('builds').filterBy('eventType', 'cron').filterBy('repo.id', this.get('id'));
})
```

I’m not quite sure about that, but see how it’s more concise? It seems easier to understand to me, and it’s more idiomatic.

I’m not sure what the role of the `event_type` and `repository_id` filters are versus the `function(b)` filter, they appear to be identical? To me, this is the kind of code that pretty much demands an automated test.

app/models/repo.js
```diff
@@ -130,18 +130,47 @@ Repo.reopen({
...
-    builds.then(function() {
-      return builds.set('isLoaded', true);
+  }.property(),
+
+  cronJobs: function() {
+    var id = this.get('id');
+    return this.store.filter('cron', {
+      repository_id: id
+    }, function(cron) {
+      return cron.get('branch.repoId') === id;
```
backspace: With my above suggestion, I think this code could be similarly simplified. But maybe some attention around naming would be useful as it’s confusing to me what the difference between a `cron` and a `cronJob` would be…?

app/models/repo.js
```diff
@@ -130,18 +130,47 @@ Repo.reopen({
...
     });
-    return builds;
   }.property(),
 
+  sortedCrons: Ember.computed.sort('cronJobs', function(a, b) {
+    return a.get('branch.name') > b.get('branch.name');
+  }),
```
backspace: And if `cron`s vs `cronJob`s are a needed distinction then this should probably be `sortedCronJobs`, right?

app/models/repo.js
backspace: I see now that this almost totally mirrors the implementation of `pullRequests`. I understand more now why you chose to implement it this way, since it’s a pattern that already exists. If it works, you can just keep it the way it is and I’ll refactor this and the `pullRequests` code to be more intelligible and concise. I’m just getting to know the codebase so sorry for complaining about code you’re just inheriting!

app/routes/settings.js
```diff
@@ -21,6 +21,36 @@ export default TravisRoute.extend({
     return repo.get('envVars.promise');
   },
 
+  fetchCronJobs() {
+    var repo = this.modelFor('repo');
+    var apiEndpoint = config.apiEndpoint;
+
+    return $.ajax(apiEndpoint + "/v3/repo/" + repo.get('id'), {
```
backspace: Elsewhere I’ve seen an `ajax` service used for Ajax calls, do you know whether that should be used here? I believe it’s used because of browser incompatibilities, though I don’t know much more than that.

app/templates/builds.hbs
```diff
@@ -4,7 +4,7 @@
     {{#each builds as |build|}}
       {{builds-item build=build}}
     {{else}}
-      {{no-builds repo=noticeData isPR=displayPullRequests isBranch=displayBranches}}
+      {{no-builds repo=noticeData isPR=displayPullRequests isBranch=displayBranches isCron=displayCrons repo=repo}}
```
backspace: You have `repo=` twice here. Is one of them more accurate?

app/templates/components/branch-row.hbs
```diff
@@ -50,99 +50,23 @@
...
-            <span class="dropup-item">View all builds</span></a>
-          </li>
-        </ul>
-      </div>
-    {{/if}}
-
-  {{/if}}
-{{/unless}}
-</div> --}}
-
```
backspace: 👍 on deleting unused code; Git remembers everything if we ever do need it!

app/templates/components/repo-show-tabs.hbs
```diff
@@ -27,6 +27,15 @@
       {{/link-to}}
     {{/if}}
   </li>
+  {{#if repo.slug}}
+    {{#if repo.cronJobs.length}}
```
backspace: Did you find that the `.length` was necessary? I believe that an empty array is “falsy” in Handlebars.

backspace:
Okay, I’ve completed my review. I do think a few improvements are necessary for this to be merged. But it’s close! If you want any further feedback or to pair/screenshare, let me know.

Nef10:
@backspace Thanks for this complete review!
As you may have noticed this is the first ember project for all of us.
We are going to adress the issues after fixing that crons are not shown on the settings page.

backspace:
You’re welcome! Well done for a first Ember effort. It’s especially challenging because the `travis-web` codebase has a lot of workarounds for things that weren’t well-supported in older versions of Ember, so there’s idiosyncratic code that has been superseded with newer patterns, but no one’s had time to get older things up to date.

The crons not showing seems like a significant problem! Hopefully soon we’ll get back to having proper acceptance tests, which would have caught such a regression.

app/routes/settings.js
Nef10: as far as i know the service can only be used for `v2` queries

app/models/repo.js
Nef10: Maybe `cronBuilds`?
But to be consistent `pullRequests` should than be renamed, too...

app/models/repo.js
Nef10: Yes, we orientated our code to the existing code.

app/routes/settings.js
backspace: ah indeed, good to know!

Nef10:
@backspace I have addressed / commented on all issues.
I would prefer if you could to the `pullRequests` and `crons` refactoring later because you are the ember expert. In my quick try it seems that `return this.store.peekAll('build');` is not fetching builds from the server and the suggested `crons` function is not working.

backspace:
ah yes `peekAll` isn’t supposed to fetch from the server, so I misunderstood that suggestion as a replacement for `fetch`. When we have tests I’ll be able to more thoroughly clean up that code you were emulating.

Thanks for addressing my suggestions, I’m going to merge this now! 🎉

aioue:
when will this be released into the travis-ci.com interface?

backspace:
Hey @rkh, do you mind giving @aioue some details on how this will roll out? Not sure what, if anything, has been communicated publicly about this.

app/models/repo.js
drogus: Just for the record, I tried `peekAll` approach, but it didn't work for some reason (ie. it wasn't properly updated when new builds were coming in). I didn't have time to look into it further, so we may try it at some point. On the other hand I think that we will be able to use `hasMany` when we have proper pagination metadata from V3, so I'm not sure how much value is in refactoring it now.

rkh:
@backspace hasn't been announced yet at all.

@aioue we're planning to roll out to .com this week, but as on .org, we will have it feature flagged in the beginning to see how it impacts load on our sysyem.

PR #465, opened 2016-02-25
[Refactor] Remove *Binding properties, inject user into relevant components

clekstro:
These bindings can be replaced wholesale with the more idiomatic
alternative: aliases.

In addition, avoid passing in user to components where it can be injected directly.

One of the perceived downsides of dependency injection can be
that it can make debugging feel more difficult because it's not
immediately clear where the value is coming from, which the explicit
variant we previously used does not suffer from.

While explicitness is often a virtue, it comes at the cost of increased
noise that pervades multiple layers of components. I'd argue this makes
the parent components more difficult to understand, given they are
littered with unnecessary references to data they themselves do not
need.

My only concern with this PR is that, without acceptance tests in place, that I may
have inadvertently broken something. I've done extensive smoke testing to try and
compensate for that, but it's possible I've overlooked something.

### DIFF
--- a/app/components/build-repo-actions.js
+++ b/app/components/build-repo-actions.js
@@ -1,7 +1,9 @@
 import Ember from 'ember';
 import RepoActionsItemComponentMixin from 'travis/mixins/repo-actions-item-component-mixin';
 
+const { alias } = Ember.computed;
+
 export default Ember.Component.extend(RepoActionsItemComponentMixin, {
-  item: Ember.computed.alias('build'),
+  item: alias('build'),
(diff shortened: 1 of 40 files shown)


clekstro:
@drogus This is now rebased, should be good to go.

lislis:
@clekstro do you want to rebase this again? Would be super :dope:!

clekstro:
I looked at this again this afternoon. There are some broken tests that I need to fix. Can hopefully get this all wrapped up tomorrow.

PR #467, opened 2016-03-02
Add pendo.io integration

drogus:


joshk:
Hey @drogus 

I think setup might be misspelt? https://github.com/drogus/ember-cli-pendo/blob/master/app/initializers/setup-pendo.js#L25

drogus:
Good catch, thanks! That should not affect the functionality, though, so I hope that what you're seeing on pendo's website is what you would expect to see.

PR #472, opened 2016-03-07
Add activate button on 'not-active' repo page

drogus:


drogus:
This is ready feature wise, but I have no idea if we want any better UI or if it can be just an additional green button:

saltinejustine:
It looks like some top margin/padding changes between the two views? Am I imagining that?

Also I think perhaps both buttons shouldn't be green. On the `This is not an active repository` view, we should maybe change the `Read the Docs on Getting Started` button to grey and keep `Activate` green since that is probably more important and should have focus.

lislis:
I think we only put in the "read the docs" link as button because we didn't have the activate button yet. Which we now have, so we could potentially remove the docs button altogether?

@saltinejustine also I think the images have different heights and that's why there is this jumping. I'll take a look at that

saltinejustine:
Alrighty — if that is the case, let me know and I can change it so they're consistent.

drogus:
> I think we only put in the "read the docs" link as button because we didn't have the activate button yet. 

For me having a docs link there would be nice anyway (just like there's also a docs button on the "no builds" page). There's also a link to profile page, which I guess could be removed. The only caveat is that once you activate the repo, there's no way to disable it using the same view and then you don't know where to go. With the current text you would probably guess that if you can also enable a repo on profile page, that would be a place to disable it. That said, I'm fine with removing the text, I'm not sure how much this explanation applies to users' behaviour.

Alternatively we could also add a deactivate button on the "no builds" page, so if someone clicks it by mistake, or just goes to a repo that shouldn't be active, it's already there. But again, I'm not sure how useful that would be and it shouldn't rather block this PR.

joshk:
I think we should remove the 'getting started' section when the repo isn't active, but show it once you have activated the page.

As for not showing a 'deactivate' link, I think this is fine as this improved interaction is about improving the onboarding. How often are users going to activate and then deactivate? 

The plan is to add activate and deactivate buttons to the settings page once we are on V3.

saltinejustine:
I think the deactivate button makes sense because I can see a use case where someone might click activate on accident and want to deactivate. But that button should be red as a warning and I agree with Piotr that it shouldn't block the PR 👍

joshk:
I'm still 👎  on adding a deactivate. I would be fine with adding a sentence about where to go to deactivate a repo. I would be more than happy to add deactivate button usage tracking to confirm if this button would be used or not. I guess A/B testing to a certain degree. I also think adding a red button with the green button will fight for attention. Anyhow, as this is not being added right now, it is a moot discussion.

saltinejustine:
@joshk cool if it's not really on the radar right now we can leave the discussion for another time all together.

lislis:
I took the liberty to move some stuff around. Looks like this now:
screen shot 2016-04-26 at 18 37 54

any takers?

lislis:
Also fixed the jumping. Should we just go for this and add a other buttons later?

saltinejustine:
@lislis looks good — one thing, remove the colon after below.

lislis:
Merging this! So exiting!

yiss, you

PR #477, opened 2016-03-14
Fixed the full color logo downloads and links

saltinejustine:
Our links for the full-color logos were not correct. This fixes the .eps and .png download files.


lislis:
\o/

PR #478, opened 2016-03-15
Permissions mixin

lislis:
Put permission checks into a mixin and use it for repo tools and dashboard menu.


lislis:
@drogus can you take a look and tell me if that makes sense?

PR #483, opened 2016-04-05
fix image format for POD

mithun:
Some POD parsers recognize only lowercase format names.

For e.g., pod2markdown fails to generate a markdown file with the status image if the format is 'HTML'


lislis:
Very nice, thank you!  
yay

PR #484, opened 2016-04-05
pass through Env.pusher.channelPrefix for enterprise

svenfuchs:
this fixes live log updates on enterprise (also, removes one trailing space, ha!)


lislis:
Awesome!
dog on a board

PR #487, opened 2016-04-10
Ja adding igor and carmen

saltinejustine:
@lislis wanna take a look at this, I got a conflict with `home.hbs` initially but I just removed it since it should affect this. But take a look and merge if you think everything is fine! Thanks!


lislis:
@saltinejustine looks like you didn't check out from master but the other branch we merged last week. A rebase should fix that. I can help you with that :)

PR #488, opened 2016-04-12
Upgrade Ember-CLI to 2.4.3

clekstro:
This does not affect our Ember/Ember-Data versions. Those will need some code changes to remove deprecations and figured this could go out on its own.


clekstro:
This may also help a little with the issue we've been encountering with longish production builds. Here are a few local benchmarks:

Before:
screenshot 2016-04-12 10 41 21

After:
screenshot 2016-04-12 10 48 09

backspace:
I too am scandalised by the huge build time. I know Ember CLI creator (and sharer of my hometown) Stefan Penner likes to improve performance and is sometimes lacking in public repositories to work with. So he may be interested in collaborating on this, as he‘s quite experienced in tracking down performance problems.

Is it true that deprecations need to be fixed to update Ember and Ember Data? I do try to address them when updating so it’s less work when they are finally removed, but there‘s currently no plan for Ember 3 and maybe it’ll never happen! I could look at updating in a separate pull request if you aren’t already doing that.

drogus:
Regarding build time, you could also try removing an addon that we use and that I wrote - ember-cli-inline-images. I doubt that they're written in the best way possible, ie. maybe I've done something really stupid performance wise.

clekstro:
@backspace Sorry if I was vague. The deprecations wouldn't technically keep us from upgrading. But if we don't fix them up front, they'd require us to either silence the errors via [config](https://github.com/travis-ci/travis-web/blob/master/config/deprecation-workflow.js) -- which feels like temporary whack-a-mole -- or to deal with the noise. My preference would be to just solve those errors in the same upgrade PR instead of postponing the work. That said, we may have other reasons to upgrade which might trump everything I just mentioned.

@drogus Thanks for the tip. I'm just digging in now and am trying to find some resources on how best to diagnose these issues.

backspace:
Running off `master`, my first build time for production was 190s and it was 165s after removing `ember-cli-inline-images` from `package.json`. So that did help, a bit.

drogus:
@backspace one thing to keep in mind is that ember-cli  could have used some cached files, so I'd also try removing `dist` and `tmp` dirs before each try. I'm not sure if that affects times at all, but I think it's worth checking.

testem.js
```diff
@@ -1,9 +1,17 @@
-{
+/*jshint node:true*/
+
+var launchInCI = function() {
+  if (!process.env.TRAVIS || process.env.TRAVIS_PULL_REQUEST) {
```
drogus: Nice, I didn't know it can be a js file.

clekstro: Yeah, this is actually new in Ember-CLI, wasn't possible before. :surfer:

backspace: That’s new as of [2.4.0](https://github.com/ember-cli/ember-new-output/commit/66d108894c9612ffe0e29ec36c749bedd21e5aeb). Useful!

PR #493, opened 2016-04-13
Update Ruby version

backspace:
Heroku deployments occasionally fail, such as [this one](https://dashboard.heroku.com/apps/travis-web-staging/activity/builds/6198e1c3-71f3-44cd-95e2-dc5c952faf06). The log says that “Only the most recent version of Ruby 2.1 is supported on Cedar-14” which according to the [linked document](https://devcenter.heroku.com/articles/ruby-support#supported-runtimes) is currently 2.1.10.

It’s strange because Heroku deployments mostly succeed, so maybe Cedar 14 isn’t always used? Regardless, it seems harmless to update to prevent these intermittent failures.


clekstro:
Based on @backspace's comments, it's clear that this is an improvement over the current state of things. Out of curiosity, is there something stopping us from upgrading to the most recent Ruby version/Cedar stack? Ruby 2.1 is growing a little long in the tooth.

drogus:
I think that while we upgrade we could go on the newest ruby version. The app serving HTML files is rather simple, so I don't think that can break anything.

lislis:
all the way up to 2.3!!!

backspace:
I’ve overwritten the commit to update to Ruby 2.3.0. I tried running the tests for `waiter/` and there were 14/20 failures. However, the same was true when I ran the tests on `master`. Have these tests fallen out of date?

drogus:
I'll check that. Also, I think that while we're at fixing deploys, maybe it would be nice to actually split the ruby and ember apps? I think that the problem above was caused by me when I was rewriting tests to use ember-try and then I totally forgot to add ruby app here as well. The ruby app changes very infrequently, so it also doesn't make sense to run its tests whenever anything in ember app changes. On top of that it would also make the .travis.yml more complicated.

I think how it could work is this:

* we create a new repository with a simple ruby app that knows how to serve assets prepared with ember-cli-deploy 
* this is the thing that is actually deployed to heroku
* travis-web stays in current repository and ember-cli-deploy puts things on S3

wdyt?

backspace:
Yes, this plan makes sense to me, @drogus. I’ll start a fresh and simple repository with basic tests and we can add functionality as needed.

backspace:
The repository in question: [`travis-web-index`](https://github.com/travis-ci/travis-web-index).

drogus:
I pushed fixed specs to `fix-waiter-specs` branch.

PR #495, opened 2016-04-13
Upgrade Ember to 2.4.5 and remove all application deprecations

clekstro:


clekstro:
The branch name of this PR is now slightly misleading, as there's more going on. The recent commits upgrade us to Ember 2.4.5 and remove all related deprecations.

PR #496, opened 2016-04-14
Deploy with ember-cli-deploy

backspace:
This is a work in progress. For now, I’ve entirely bypassed the existing Ruby server and replaced it with a tiny Sinatra application from the `ember-cli-deploy` documentation. This means the `meta` tag environment variable substitution of `waiter/lib/travis/web/app.rb` doesn’t happen, so I expect the deployment to be non-functional. My goal for now is to get the assets and `index.html` deployed via the `.travis.yml` deployment script.


backspace:
@clekstro, @lislis, and @drogus (not sure who else might care?): I have partial success at creating a deployment for every in-repository pull request. My aim is that people within the organisation can much more easily demonstrate their proposed changes. Work remains to ensure deployments only happen for desired PRs and that the deployment functions as it should. Some cleanup and improvements are also needed.

You can see below that the PR checks now have a deployment link:

image

My user image appears there because I had to use my own token to call the status API. Maybe @travisbot is the answer here? I’ll look into this.

### Branch name as subdomain?

The deployment link currently looks like this:
https://travis-web-production-next.herokuapp.com/?index_key=6b0aac9f171c2f65c6ce2312250e1b28

I consider this less-than-desirable because the `index_key` parameter disappears immediately when you start using the application, so you can no longer track what deployment you’re visiting. Worse in our case is that when you try to log in to Travis via GitHub, you are forwarded back to the domain with no `index_key`, so you won’t return to the proper deployment.

I’ve previously solved this problem by using a subdomain that contained a normalised version of the branch name. For this pull request, it would be `ember-cli-deploy.staging.travis-ci.org` (or whatever appropriate host). I like this because you can glance at the URL and know what deployment you’re working with. It also solves the forwarding-after-login problem. You lose the capability of using a particular `index_key` for whatever commit you’re interested in viewing, but I’ve never found myself needing that; if we did, we could add support for an `index_key` subdomain as well.

### SSL and subdomains

I’m anticipating some wrinkles regarding SSL and DNS. The last time I set this kind of thing up, the SSL certificates didn’t support unlimited subdomains, so we had to use a hyphen instead. Hopefully things are more flexible here. I assume SSL is a requirement.

### Environment variables

I used environment variables on the [`travis-web` Travis settings route](https://travis-ci.org/travis-ci/travis-web/settings). Needed were `REDIS_URL`, `AWS_KEY`, `AWS_SECRET`, and `GITHUB_TOKEN` (which I mentioned above). Is this the standard way to add environment variables now? Is there some automation necessary? I remember having encrypted variables in `.travis.yml` in the past, has that fallen out of favour?

### Separating the Redis/`index.html` server

I used the existing `waiter` Rack server, with slight modifications suggested by @drogus, to serve Ember’s `index.html` file. There’s some code in there that substitutes Ember environment variables to get everything set up.

Since the deployment of assets happens to S3 and `index.html` goes into Redis, there’s no need to have this server in the same repository as `travis-web`. I propose creating a `travis-web-index` repository for the Redis front-end, I can strip out a lot of the asset handling while I’m at it. Some complexity is still required if we want things like badge images to work, but maybe that’s not important unless this deployment strategy ultimately completely replaces the Rack server.

### Environments to support

I copied the Heroku environment variables of `travis-web-production` to the `travis-web-production-next` server I’m using as the Redis front end. So the Rack server is using org/production environment variables to connect to the relevant services. My current thinking on this is that we might as well have a deployment link for all four environments: org/production, org/staging, com/production, and com/staging. Any potential problems with that that I’m missing?

There could be four Redis instances, but there could also just be four Heroku Redis front ends all pointing to one Redis. There could also just be _one_ Heroku Redis front end that did the environment variable substitution differently.

### Possible Travis features to simplify this

@joshk encouraged me to think of ways this process could be smoothed for other `ember-cli-deploy` users. He talked about using GitHub deployment features and/or `dpl` for some aspects. I’ll consider this as I refine things.

I mentioned that my current plan as discussed with @drogus is to only support this type of deployment for internal pull requests, which mitigates the security concern of malicious PRs that appear to be legitimate instances of Travis. @joshk suggested that it _would_ be nice for people outside the organisation to get this sort of deployment, so maybe we can think of a way to make this work with no risk of malicious impersonation.


whew! A lot of text, sorry. I can proceed with cleanup tasks but some of the above I’ll have to reach out for help on.

backspace:
Progress report:

`ember-cli-deploy` collaborator @lukemelia informed me of [`ember-cli-deploy-github`](https://github.com/seawatts/ember-cli-deploy-github), which takes care of calling the status API with a URL for the deployment, which would be nice to get rid of some of the shell script hackery as removed in the [commit where I added the addon](https://github.com/travis-ci/travis-web/pull/496/commits/1aeb2f901683e4abb151ea59a95a69e7c2c9be2c). Sadly, it seems to hardcode a particular format for the index key, so it may not be flexible enough to accommodate the branch-name-as-subdomain idea. I may be able to make a PR to add that possibility.

So the deployment link in the status listing is currently broken, I’ll come back to that on Monday.

I got a certificate and subdomain for deployments and amended the Rack server to accept the index key as a subdomain. So you can visit the latest deployment here; when you log in, you are properly forwarded back to the desired deployment:
https://8709b4f2d1f4c2d6ef0cda751406755b.test-deployments.travis-ci.org/

My next goal is to use the branch name as the subdomain. It’s not immediately clear to me that it’s even possible to trivially find the branch name in the Travis environment. It’s not an environment variable, that I can see? I can always fall back to using the GitHub API to access the branch name via the pull request ID, but that’s more laborious than I’d like.

backspace:
There’s now a [working deployment link](https://ember-cli-deploy.test-deployments.travis-ci.org/) that uses the branch name as the subdomain. 🎉 

I’ve started [`travis-web-index`](https://github.com/travis-ci/travis-web-index) to split off the server that delivers `index.html` from Redis. It’s not yet serving at travis-web-production-next.herokuapp.com but I have the beginnings of a test suite so I should be able to get it working for the basics tomorrow. Then we can 👋  to the `/waiter` directory!

There are two uses of the GitHub API’s `pulls` endpoint that are pretty messy right now to obtain the branch name and the latest commit to the pull request:

1. `$TRAVIS_COMMIT` gives the SHA of a merge commit (I think?) which won’t work for the call to the status API, so I get it out of the GitHub API instead and store it as `$TRAVIS_PULL_REQUEST_COMMIT`.
2. There’s no environment variable for the branch name, so I also get it from the API and store it as `$TRAVIS_PULL_REQUEST_BRANCH`.

My method of doing so involves `grep` and `cut` and is maybe brittle. It would be ideal if these were just environment variables, but in the meantime maybe I should use `jq` or the like.

backspace:
So! I think this is ready. I made some executive decisions to limit scope:

1. Only the production-org environment is supported on the Heroku Ruby Redis front-end, for now.
2. `ember-cli-deploy` deployments are _only_ for pull requests right now.

But I think this is good for now! Thoughts, @clekstro and @drogus?

clekstro:
I think this is awesome and am super happy to leverage this ASAP.  I think we should document how to use this in the README, squash and 🚢

drogus:
It looks great!

Regarding your question about env vars in settings - yes, this is currently a preferred way to set environment variables. The only caveat is that they're global across all of the builds, so maybe there're cases when you may need to use secure env vars in `.travis.yml`, but we haven't had any problems like that so far.

The only problem that I have with merging this is that as far as I understand it will be deployed for all of the PRs. In order to make it safe I would add a `TRAVIS_SECURE_ENV_VARS` check there in order to limit it only to our PRs.

Regarding the branch, I think that the GitHub call is fine for now, but I would also look on what we would need to change to get `TRAVIS_PULL_REQUEST_SOURCE_BRANCH` or something along the lines working out of the box.

backspace:
Okay, I added a check for `TRAVIS_SECURE_ENV_VARS` to `.travis.yml` and fixed a typo. Merging!

PR #503, opened 2016-04-19
Add Buck to team page

backspace:
Here is my new submission.


clekstro:
This looks pretty great

backspace:
Hey, @saltinejustine, how do these assets look for sizing etc?

(Note that the test failures are spurious, the result of canceled builds for UI experimentation)

saltinejustine:
@backspace looks good! just one thing, the portrait img should be sized down to 226x226px which is our retina size. Gif size is good. If you fix that, you may deploy that to the team page!

backspace:
226px it is! 🎉

PR #504, opened 2016-04-19
More lookup fixes

clekstro:


app/models/broadcast.js
```diff
@@ -1,41 +0,0 @@
-import Ember from 'ember';
```
drogus: I think that this is still being used, we display broadcasts in the top bar.

app/services/ajax.js
```diff
@@ -62,24 +62,12 @@ export default Ember.Service.extend({
       options.contentType = options.contentType || 'application/json; charset=utf-8';
     }
     success = options.success || (function() {});
-    options.success = (data, status, xhr) => {
```
drogus: I think that we misunderstood each other when we discussed this. This code is still needed until we switch calls that may have flashes in the response. So a proper fix for now would be to inject `flashes` service here and do:

```
       if (data != null ? data.flash : void 0) {
         this.get('flashes').loadFlashes(data.flash);
       }
       if (data != null) {
         delete data.flash;
       }
```

And the same should go to the error callback.

app/models/broadcast.js
clekstro: I wasn't able to find any of the methods (e.g. `isSeen` or `setSeen`) being used anywhere else in the app. The rest of the app appeared to use the storage service. I will double-check though.

app/services/ajax.js
drogus: Or, alternatively, we could pair on removing all of the ajax calls that needs it and then you could leave it removed here.

clekstro: Yeah, let's have a look at it together :thumbsup:

drogus: I've checked the API and unfortunately there's a bug that may prevent us from implementing this, so for now I'd leave `loadFlashes` here.

app/models/build.js
```diff
@@ -129,7 +129,7 @@ Build.reopen({
   },
 
   restart() {
-    return this.get('ajax').post("/builds/" + (this.get('id')) + "/restart");
+    return Ember.$.post(`/builds/${this.get('id')}/restart`);
```
drogus: I missed this, it should stay `ajax`, because otherwise will fail. This is an interesting issue, though, the new test didn't catch it, because in Ember CLI Mirage we don't check any headers (like: api version, authentication etc)

app/utils/repo-actions-item-component-mixin.js
```diff
@@ -21,18 +25,36 @@ export default Ember.Mixin.create({
     return this.get('item.canRestart') && this.get('userHasPermissionForRepo');
   }.property('userHasPermissionForRepo', 'item.canRestart'),
 
+  displayFlashError(status) {
+    let type = this.get('type');
+    if (status === 422 || status === 400) {
+      this.get('flashes').error(`This ${type} can't be cancelled`);
```
drogus: This message will not work for restarting, we should either get back to do it inline or parametrise it somehow.

app/models/build.js
backspace: oh yeah, this ended up being a red herring when debugging why the test wasn’t working. It can be rolled back.

clekstro: indeed. good catch. didn't mean for this to make it in.

app/utils/repo-actions-item-component-mixin.js
clekstro: Is this because the `/restart` endpoint won't return an error code currently?  Can we simply not invoke the function in that case, or is there something else that should be done?

drogus: No, sorry, I just mean that the message says "can't be cancelled".

clekstro: Duh. 👍

backspace:
Is this ready now, @clekstro?

clekstro:
@backspace Yeah, should be g2g. I was just waiting to let @drogus take a look if he wanted before merging.

PR #506, opened 2016-04-20
Disable Sauce Connect

backspace:
It’s not currently being used and it’s limiting the number of concurrent builds, so this disables it until we want to start using it again.


backspace:
Ref [@drogus on Slack](https://travisci.slack.com/archives/teal/p1461148398000121)

PR #507, opened 2016-04-20
Profile route acceptance test

backspace:
This is not ready, it’s just one assertion to start with.

### DIFF
--- a/app/templates/account.hbs
+++ b/app/templates/account.hbs
@@ -59,7 +59,7 @@
     </ol>
   </section>
 
-  <div>
+  <div id="administerable-hooks">
     <ul class="profile-hooklist">
       {{#each hooks as |hook|}}
         {{hooks-list-item hook=hook}}
--- a/mirage/config.js
+++ b/mirage/config.js
@@ -40,6 +40,11 @@ export default function() {
     return { hooks: hooks.map(hook => hook.attrs) };
   });
 
+  this.put('/hooks/:id', (schema, request) => {
+    const user = schema.hook.find(request.params.id);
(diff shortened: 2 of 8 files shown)


backspace:
image

My suggestions as to what assertions are good for this test:

* the user’s name
* their organisations and repository counts
* their repository list and whether they are enabled

backspace:
In addition to my screenshot annotations, I added assertions about:
* revealing the user’s token
* the presence of “unadministerable” hooks (where `!admin`)

I imagine we will eventually want to extract the `localStorage`-related boilerplate currently in the module setup, but I’m fine with copying it around for now.

Do you think this is adequate coverage for this route, @drogus?

backspace:
hmm, now I’m wondering whether there should be testing around what happens if you enable/disable a hook…???

backspace:
Okay, I added assertions against toggling hooks. I don’t really like the way the tests reach into the Mirage database _by index_ to find the post-update attributes on hooks. But I don’t know what would work better… this is my first time using Mirage and that’s [one of two explicitly-recommended techniques](http://www.ember-cli-mirage.com/docs/v0.2.0-beta.8/acceptance-testing/#asserting-a-server-call-was-made-in-a-test).

drogus:
@backspace I think that the toggle test is fine for now. In the future (with V3) we will be using repositories endpoint to handle the enabling and disabling them, so I wouldn't want to put too much pressure on hooks.

I think we can merge it, but I think we could abstract the auth code soon after that. It seems like a simple helper would be enough, but we can discuss it together with @clekstro

PR #508, opened 2016-04-21
Home page acceptance tests

clekstro:
This is one of those cases of making the change easy so you can make the easy change. The initial acceptance test for org showed little difficulties, but I ran into some heavy resistance when attempting to stub config flags for pro.  Simply overwriting a config value in a beforeEach block of an acceptance test was not enough, as it somehow gets overridden over time (value was fluctuating, which to me made absolutely no sense).

I also looked into [ember-config-service](https://github.com/gmurphey/ember-config-service), figuring we could wrap config in a service and then register mock services in each of our tests to overwrite the values we care about. I decided against it, however, given 
1) we could probably write something similar ourselves and 
2) that manner of stubbing would introduce a lot of noise and ceremony.

Enter [ember-feature-flags](https://github.com/kategengler/ember-feature-flags).  This is something I feel like we could leverage in other ways (e.g. the debug toggle I've mentioned this week), and it has a solid testing story that should make testing various product versions much more straightforward. Also, it's excellently maintained. I'm not yet to the point of actually writing my desired test, but that should be relatively straightforward.


app/controllers/repos.js
```diff
@@ -170,7 +170,7 @@ var Controller = Ember.Controller.extend({
           return reposRecordArray;
         };
 
-        if(Config.useV3API) {
+        if(this.features && this.features.useV3API) {
```
drogus: Shouldn't it be `this.get('features.useV3API')`?

app/models/log.js
```diff
@@ -21,7 +23,7 @@ var Request = Ember.Object.extend({
   },
 
   handle(body, status, xhr) {
-    if (config.pro) {
+    if (config.featureFlags.pro) {
```
drogus: Unsure if it's needed now, but in the future we could make the `Request` object container aware, so we can use a service.

app/controllers/repos.js
clekstro: I think both syntaxes are allowed, I'd just need to be consistent.

app/models/log.js
clekstro: Yeah, I've tried to keep the changes as small as possible to avoid getting overwhelmed. Definitely think this should happen in the future though.

app/routes/home-pro.js
```diff
@@ -2,5 +2,5 @@ import TravisRoute from 'travis/routes/basic';
 import config from 'travis/config/environment';
 
 export default TravisRoute.extend({
-  needsAuth: false
+  // needsAuth: false,
```
drogus: We can remove this.

app/routes/main.js
```diff
@@ -3,18 +3,28 @@ import config from 'travis/config/environment';
 
 export default TravisRoute.extend({
   renderTemplate() {
-    $('body').attr('id', 'home');
```
drogus: I really don't like this style of differentiating between layouts, but I think that we need to leave it here. We rely on the id of a body element in a few places and if it doesn't get set here, we may end up with an unexpected result. We don't use `#home` anywhere in the sass from what I see, but we use `#not-found` (among others). So for example when you go to "not found" page and transition to home page, you would end up with the wrong id.

clekstro:
This has languished and largely been replaced by other efforts. May cherry-pick some pieces here or there in the future.

PR #510, opened 2016-04-21
Move debugging logging behind feature flag

clekstro:
This removes the console clutter we have in our tests (as well as
production). The development environment would continue to contain the logged
information.

My plan for production was to create a debug button in the future that we could
activate at will to debug an issue, so we only have the logging as
needed.


clekstro:
Going to take a fresh crack at this later.

PR #511, opened 2016-04-21
Avoid building phantom.js from source

clekstro:
This doesn't appear to be necessary in other projects I've encountered
recently, so removing it to see if it makes any difference. If nothing
breaks, it shortens our build config at the very least.


backspace:
It looks to me like it was downloading a binary and not building from source, but whatever, might as well use the local instance if it’s already there!

PR #513, opened 2016-04-22
Use dot reporter within CI for shorter logs

clekstro:
This will still show a failure/stacktrace at the end of our builds, meaning it simply cuts down on the noise
of our passing tests

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,7 +41,7 @@ install:
   - bower install
 
 script:
-  - ember try:one $EMBER_VERSION
+  - ember try:one $EMBER_VERSION --- ember test --reporter dot
 
 notifications:
   campfire:


clekstro:
Here's a preview, though you can obviously just check out the builds themselves:
screenshot 2016-04-22 02 17 17

backspace:
I have a mild preference for more verbose logging where each test is shown, but not enough to insist upon it. Do you have a preference, @drogus? I’ll assign this to you for the final decision.

clekstro:
To @backspace's point, I'm generally only interested in the output when something goes wrong, which this reporter preserves. Half of our tests at this point are also just `jshint` checks. Here's a screenshot of what that looks like (sorry, should have attached before):

screenshot 2016-04-22 11 39 00


That said, it's really a matter of opinion/taste, so I'm fine if we decide to keep it.

drogus:
I prefer the dots, I always find it hard to find a reason a test suite failed with a default output.

PR #516, opened 2016-04-24
Remove campfire notifications

clekstro:
I am pretty sure this is just an artifact of a bygone era. @drogus can you confirm?

Also, curious what you and @backspace think about setting up slack notifications as an alternative?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,10 +41,6 @@ install:
 script:
   - ember try:one $EMBER_VERSION --- ember test --reporter dot
 
-notifications:
-  campfire:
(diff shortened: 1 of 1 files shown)


backspace:
I generally prefer email over Slack because it’s easier to use “mark unread” to store things for future attention, which is my workflow. But Slack is fine too, I can just mute channels if I need to.

drogus:
Honestly I wouldn't set any notifications that way. The way it works when you use `notifications` key is that Travis CI will send a notification on a given medium for each build that matches the settings (I think the default is: any failed build and any build that succeeds after previous failures). This seems unnecessary as we already send notification to the author and committer anyway.

PR #517, opened 2016-04-25
Add sentry integration

clekstro:
This will only log to Sentry in production-like environments (staging and PRs *will* log).
In development, we will receive a console error that Sentry *would have* been notified.

In parallel with this, I have integrated Sentry to post to a #production-web-errors room in
Slack. Email notification is currently disabled, as I do not wish to spam the entire org whenever
something is undefined.

In the future, we might consider having different channels for non-production environments, but that can be addressed later.

### DIFF
--- a/app/components/builds-item.js
+++ b/app/components/builds-item.js
@@ -6,6 +6,11 @@ export default Ember.Component.extend({
   classNameBindings: ['build.state'],
   classNames: ['row-li', 'pr-row'],
 
+  didInsertElement() {
+    this._super(...arguments);
+    throw new Error("Error");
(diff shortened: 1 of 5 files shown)


clekstro:
So it looks like simply enabling source maps within the production builds does not magically make our stack traces particularly helpful. Given that this is currently a stop gap until we improve our acceptance test coverage, I think this can go out as is.

I'm going to create a new issue to track improving the stack traces, but I think this has a lower priority than the test coverage at this point. For future reference, there is an ember-cli-deploy-sentry plugin that should automate this for us.

PR #518, opened 2016-04-25
Fix Sass extend problems

backspace:
A newer version of node-sass was released and it caused build errors related to these `@extend`s. This removes two uses of `@extend` that didn’t actually extend anything that existed. The other change is syntactic: a missing space was causing the parent selector to have nothing defined within it.


backspace:
The HPI people ran into this problem this morning and prompted this fix. So thanks to them for reporting it, otherwise we would have eventually run into failures on Travis and locally if we ever updated `node-sass`.

PR #519, opened 2016-04-25
Extract test session setup and teardown

backspace:
In preparation for my next acceptance test, I extracted a `storeUserSession` test helper and added the clearing of `localStorage` to the `afterEach` of all acceptance tests. This will reduce boilerplate.

This is pretty simple so I’ll probably just merge after the tests have passed, but let me know if you’d prefer a name other than `storeUserSession` or anything else?

### DIFF
--- a/tests/.jshintrc
+++ b/tests/.jshintrc
@@ -49,7 +49,8 @@
     "currentRouteName",
     "jQuery",
     "Visibility",
-    "Travis"
+    "Travis",
+    "storeUserSession"
   ],
   "node": false,
   "browser": false,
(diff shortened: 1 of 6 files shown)


drogus:
Looks good, although I'm not sure about the name. Store user session names the function after it's implementation and not the intent. We will be probably changing auth in the future and this may change and the name will not make sense anymore (for example we were discussing using cookies if we want to preload data in travis-web or use fast-boot once it's able to rehydrate). What do you think about something like `signUserIn(currentUser)` o something similar?

PR #522, opened 2016-04-26
Use absolute paths for team assets

backspace:
As [pointed out by @BanzaiMan](https://travisci.slack.com/archives/teal/p1461614758000435), team asset links were broken in PR deployments. Will this fix it?


backspace:
This isn’t working as I expected… I’ll try to figure out a solution. It’s fine for PR deployments to be broken, but if we eventually move to this style of deployment for production, we’ll have to have an answer.

backspace:
I made another effort at this yesterday but it appears that dynamically-constructed asset paths aren’t rewritten. This doesn’t seem high-priority as it doesn’t affect production deployments, though it’s sad that PRs for team page changes won’t render properly 😢

PR #531, opened 2016-05-02
More package upgrades

clekstro:
This bumps pusher to the most recent version, so will want to test this extensively to make sure it doesn't bork stuff


backspace:
I’m pretty sure this is long superseded so I’m closing it. Let me know if I’ve done this in error though!

PR #534, opened 2016-05-06
Move Help Scout beacon to navigation

backspace:
People have been complaining about the Help Scout beacon button. This is an experiment to move it to be a link-like button in the top navigation.

### DIFF
--- a/app/app.js
+++ b/app/app.js
@@ -82,6 +82,10 @@ var App = Ember.Application.extend(Ember.Evented, {
 
     if (window.HS) {
       HS.beacon.ready(function() {
+        HS.beacon.config({
+          autoInit: false,
+          modal: true
+        });
         HS.beacon.init();
       });  
     } else {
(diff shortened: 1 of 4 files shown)


joshk:
can you include a screenshot of what this looks like?

generally I'm not in favour of this. I would like to keep this button at the bottom right. Before we decide though, I would like to see if we can reduce the size of the button by 30%, move it to the bottom right, and instead add a margin to the build/job view (and log) which means there is no overlap, similar to what Joe whipped up: https://files.slack.com/files-pri/T024GG3SR-F16MBD0C9/job__30611.1_-_travis-pro_chirp-com-production_-_travis_ci_2016-05-06_11-43-07.png

Having this button at the bottom right is a sudo standard as this is how people expect intercom to be used, for example. If you use inbox (inbox.google.com) they use the button at the bottom right, plus margin, for composing a new email.

backspace:
Yeah, I only made this experimental PR so I could deploy to staging, since I don’t know how to get it running on my own machine.

backspace:
Closing in favour of travis-pro/team-teal#848

PR #536, opened 2016-05-10
Sentry app version

clekstro:
We need to set the release version explicitly on the global Raven object to have errors marked correctly. Opening a PR so I can test without impacting staging.


clekstro:
Merged this functionality already. Closing.

PR #537, opened 2016-05-16
Replacing old assets with optimized ones

saltinejustine:
This Pull Request will resolve some performance issues we have on our team page. Almost all the gifs have been optimized to be significantly smaller and portraits were also optimized to dump wasted kb. This will hopefully result in improved performance especially on mobile devices.


lislis:
:dope:  
dance

saltinejustine:
Bonus points to you @lislis for the seinfeld gif 😉

PR #538, opened 2016-05-17
Fix fingerprinting for Enterprise

clekstro:
Not sure if this is ideal or not, but this fixes https://github.com/travis-pro/travis-enterprise/issues/116.  We're now able to build and deploy enterprise again.

### DIFF
--- a/ember-cli-build.js
+++ b/ember-cli-build.js
@@ -14,14 +14,18 @@ module.exports = function(defaults) {
       extensions: ['js', 'css', 'png', 'jpg', 'gif', 'map', 'svg']
     };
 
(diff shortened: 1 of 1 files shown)


backspace:
It doesn’t seem harmful to me. How is Enterprise deployed?

clekstro:
It has its own setup. I'll probably just merge this tomorrow, I'm just confused as to why this was necessary in the first place.

backspace:
But does it use `ember-cli-deploy`? I know nearly nothing about Enterprise, I’m just wondering about future possibilities of simplifying and collapsing differences.

PR #540, opened 2016-05-18
Settings acceptance test

backspace:
This is in progress. I expect it to fail!


backspace:
Okay, it’s passing now. I have assertions for passive viewing of all “General Settings”. To do:

* [ ] environment variables
* [ ] crons
* [ ] setting of all the above
* [ ] cleanup/correction of assumptions in mock API specifications

backspace:
To @clekstro, @lislis, and @drogus (maybe), in my [most recent commit](https://github.com/travis-ci/travis-web/pull/540/commits/d19d366b74bba565be02e1402055f1bc2568373b) which adds an assertion that changing the concurrent build setting saves properly to the server, I changed the implementation of the `limit-concurrent-builds` component in a way I’m not sure about.

It previously used a `key-up` event to trigger saving the new setting after every key press, which was causing some trouble for me when trying to trigger a change in an acceptance test. Now it uses a `change` event, which will only save the setting after focus has left the field. Does this seem like a problem? It seems to me that saving only once is actually preferable, but maybe I’m missing something…?

I’ve got passing tests for the major functionality of the settings page, now I just need to do some cleanup.

lislis:
@backspace if I remember correctly, we used `key-up` because `change` caused troubles when trying to enter a two digit value because it would try and save the first digit and then overwrite it with the second one. Does that make sense? Can you test for that?

backspace:
hmm, well I’m glad you remember something about it, @lislis! I I can’t really figure out why using `key-up` would not cause overwriting but `change` would. Sadly I’ve found automated testing of these kinds of interactions not very representative of what it’s like for actual users. But for this maybe I’ll just manually test in various browsers? It definitely works for me in Chrome with saving the entire number and not one individual digits, but I can make sure that’s the case in other browsers.

backspace:
After a little detour to fix the PR deployments, which broke because Heroku inexplicably changed the Redis URL 😠, I’ve tested the `change`-based field in Chrome, Firefox, and Safari, and it worked as expected in each. You have to tab out of the field to save it, I wonder if that’s a problem? Hopefully not…

I’m going to proceed with cleanup and hopefully finish this soon.

backspace:
I realised during cleanup that I was missing assertions for SSH keys (and whatever the `/repo/:id/keys` endpoint is), but I’m going to merge this without that because I’ll be off next week and I need to put this aside to create a deployment with Segment instead of Pendo. The routes related to these missing tests are marked with `FIXME`s.

In case you’ll be doing more acceptance test work before your time off, @clekstro, some notes:

I amended your recent change to `turnIntoV3` to remove the `payload.models` addition, instead having the callers access that property themselves. More on that at bb132a2.

Mirage is pretty new to me, but I think specifying relationships as in your [change to `repository.js`](https://github.com/travis-ci/travis-web/blob/256a415e4d36587177c31b630895bdef50f152b8/mirage/models/repository.js) doesn’t do anything unless you use things like `repository.createBuild` to define related models. I think what I’ve been doing with Mirage routes doesn’t fully exercise Mirage as intended; certainly the entire serialiser concept isn’t even being used! As I write more acceptance tests, I’ll hopefully see how those can be integrated, as it would avoid [this kind of awkward mapping](https://github.com/travis-ci/travis-web/blob/4e989fb4d6c15ff4111de33ea663e06d3c42ac6c/mirage/config.js#L111).

I have a `TODO` here where I have to hack around `turnIntoV3`’s lack of relationship-handling. I don’t yet know how relationship serialisation should be handled. Probably with a Mirage serialiser, customised for the V3 API.

Yeah. I’m not in love with everything in this PR but as we grow the acceptance tests, I think we can improve on some of my decisions here when the overall needs are more clear.

PR #543, opened 2016-05-26
Update repo.hbs

Elektron1c97:
In every other title GitHub is written with a capital H so why not use this in the repo's title :stuck_out_tongue:

(sorry for the useless build :grin:)


lislis:
Good catch! Thanks!
awesome

PR #544, opened 2016-05-27
Experiment with Segment integration

backspace:
As requested by @joshk in a call this morning, issue forthcoming.

I’ve removed Pendo and very naïvely added an initial tracking event for the application route. I added a `SEGMENT_STAGING_KEY` to the [Travis settings](https://travis-ci.org/travis-ci/travis-web/settings) that should work for the deployment, but this will probably need further thought.

Let’s see what happens!

### DIFF
--- a/app/mixins/track-current-page.js
+++ b/app/mixins/track-current-page.js
@@ -1,0 +1,15 @@
+import Ember from 'ember';
+
+// Taken from https://github.com/poteto/ember-metrics/issues/43#issuecomment-194690887
+
+export default Ember.Mixin.create({
+  metrics: Ember.inject.service(),
+  trackCurrentPage: Ember.on('activate', function () {
+    Ember.run.scheduleOnce('afterRender', this, () => {
+      const page = document.location.pathname;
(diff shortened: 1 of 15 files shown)


backspace:
I have `page` and `identify` events triggering, which you can view at the [debugger page](https://segment.com/travis-ci-staging/sources/javascript/debugger). I don’t know how you want Segment set up, @joshk; it suggests either adding “integrations” to route data toward or a “warehouse” to store it. The three routes current triggering `page` events are:
* `repo`
* `repo/pull_requests`
* `repo/settings`

The `identify` event _should_ be triggered when the application loads and you’re signed in, or when you sign in post-load.

There’s a lot of flexibility in what can be tracked here, so just let me know what you want.

I’d like some assertions in acceptance tests that events are being tracked, which I’ve seen in a previous Ember application. We don’t have many acceptance tests yet, but I could add them to what’s currently present.

backspace:
I’m going to close this as it seems to not be a priority at this time. I can salvage some of its code if it becomes a priority in the future.

PR #547, opened 2016-05-30
Cron interval

Nef10:
Instead of the next enqueuing the concrete interval is now shown:

pasted image at 2016_05_30 11_34 am

This depends on travis-ci/travis-api#266.

### DIFF
--- a/app/components/cron-job.js
+++ b/app/components/cron-job.js
@@ -8,7 +8,7 @@ export default Ember.Component.extend({
   actionType: 'Save',
   store: service(),
 
-  enqueuingText: function(key) {
+  intervalText: function(key) {
     var interval = this.get('cron.interval');
     var creationTime = new Date(this.get('cron.created_at'));
     var hours = creationTime.getHours();
--- a/app/components/cron-job.js
+++ b/app/components/cron-job.js
@@ -10,7 +10,7 @@ export default Ember.Component.extend({
 
   isOverdue: function(key) {
     var time = this.get('cron.next_enqueuing');
-    if (time + 1 > Time.now){
+    if (time + 1 > Date.now()){
       return true
     }
     return false
--- a/app/components/cron-job.js
+++ b/app/components/cron-job.js
@@ -9,37 +9,50 @@ export default Ember.Component.extend({
   store: service(),
 
   intervalText: function() {
+    function timeOfDay(creationTime) {
+      var hours = creationTime.getHours();
+      var ampm = hours >= 12 ? 'pm' : 'am';
+      hours = hours % 12;
+      hours = hours !== 0 ? hours : 12;
+      return hours + ' ' + ampm;
+    }
+
+    function dayOfWeek(creationTime) {
+      var weekday = new Array(7);
+      weekday[0] = "Sunday";
+      weekday[1] = "Monday";
+      weekday[2] = "Tuesday";
+      weekday[3] = "Wednesday";
+      weekday[4] = "Thursday";
+      weekday[5] = "Friday";
+      weekday[6] = "Saturday";
+      return weekday[creationTime.getDay()];
+    }
+
+    function dayOfMonth(creationTime) {
+      var post = 'th';
+      var day = creationTime.getDate();
+      if ( day === 1 ){
+        post = 'st';
+      } else if ( day === 2 ){
+        post = 'nd';
+      } else if ( day === 3 ){
+        post = 'rd';
+      }
+      return day + post;
+    }
+
     var interval = this.get('cron.interval');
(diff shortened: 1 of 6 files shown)


app/templates/components/cron-job.hbs
```diff
@@ -1,6 +1,10 @@
 <div class="cron-job-value branch-name">{{cron.branch.name}}</div>
 <div class="cron-job-value interval">{{cron.interval}}</div>
-<div class="cron-job-value next-enqueuing">{{pretty-date cron.next_enqueuing}}</div>
+{{#if isOverdue}}
+  <div class="cron-job-text next-enqueuing">{{"soon"}}</div>
```
clekstro: If we're just returning a string here, no need for the `{{ }}` syntax. We can just do:

```
<div class="cron-job-text next-enqueuing">soon</div>
```

app/components/cron-job.js
```diff
@@ -9,8 +9,8 @@ export default Ember.Component.extend({
   store: service(),
 
   isOverdue: function(key) {
```
clekstro: These computed properties don't appear to be using the `key` being passed in. Can they be removed?

app/components/cron-job.js
```diff
@@ -8,12 +8,39 @@ export default Ember.Component.extend({
...
+        weekday[5] = "Friday";
+        weekday[6] = "Saturday";
+        text = 'Every ' + weekday[creationTime.getDay()] + ' at ' + creationTime.getHours();
+        break;
+      case 'daily':
+        text = 'Every day at ' + creationTime.getHours();
+        break;
     }
-    return false;
+    return text;
```
clekstro: It might be nice to break this up into some smaller functions to make it a little more readable.

tests/acceptance/repo-settings-test.js
```diff
@@ -111,13 +111,11 @@ test('view settings', function(assert) {
...
     assert.equal(settingsPage.crons(0).branchName, 'daily-branch');
-    assert.equal(settingsPage.crons(0).interval, 'daily');
-    assert.ok(settingsPage.crons(0).nextEnqueuing.indexOf(':19'), 'expected cron next enqueuing to match');
+    assert.ok(settingsPage.crons(0).nextEnqueuing.indexOf('Every') === 0, 'expected cron next enqueuing to match');
     assert.ok(settingsPage.crons(0).disableByBuildText.indexOf('Even') === 0, 'expected cron to run even if no new commit after last build');
 
     assert.equal(settingsPage.crons(1).branchName, 'weekly-branch');
-    assert.equal(settingsPage.crons(1).interval, 'weekly');
-    assert.ok(settingsPage.crons(1).nextEnqueuing.indexOf(':11'), 'expected cron next enqueuing to match');
+    assert.ok(settingsPage.crons(1).nextEnqueuing.indexOf('Every') === 0, 'expected cron next enqueuing to match');
```
clekstro: Can we distinguish between the assertion messages in these two cases?  Maybe something like "Shows weekly/daily enqueuing text" or something to that effect?

clekstro:
Except for the few things that could use a little cleanup, code looks fine to me 👍 .

Nef10:
@clekstro adjusted according to your comments

app/components/cron-job.js
```diff
@@ -8,6 +8,46 @@ export default Ember.Component.extend({
   actionType: 'Save',
   store: service(),
 
+  intervalText: function() {
```
renee-travisci: I would split this up into 3 helper methods for extracting the time, calculating the day of the month, and getting the day of the week. Then have the procedural function that is putting together the text variable, call your helper functions. The encapsulation will help test and isolate logic, which is easier to maintain for 3 reasons: 

1. Smaller procedural functions are easier to read and understand for other developers (or yourselves) in the future. 

2. You can unit test the isolated logic (it's easier to cover edge cases of smaller logic).

3. You can replace the logic more easily if needed in the future. For example if we wanted to support another language we could edit the function that grabs weekday names, and add new tests for just that logic without changing the larger procedural function or its integration tests.

renee-travisci:
As an FYI there is a large js library that does this type of work that you may want to look at for reference called later.js (https://bunkat.github.io/later/). I don't think we would want to bring in such a large dependency for something simple like this, but it might be a good learning tool to look at that code.

app/components/cron-job.js
clekstro: @renee-travisci Thanks for such clear instructions. I had mentioned breaking this up before, but this is much more actionable.

chistopher:
changed text 
image

chistopher:
reopen

PR #552, opened 2016-06-06
Clarify cron time UI

Nef10:
Reopen of #547
@carlad

### DIFF
--- a/app/components/add-cron-job.js
+++ b/app/components/add-cron-job.js
@@ -54,6 +54,6 @@ export default Ember.Component.extend({
 
   intervals: ['monthly', 'weekly', 'daily'],
 
-  options: ['Only run if no new commits', 'Always run']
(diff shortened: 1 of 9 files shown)


carlad:
The position of the interval message should align more closely with the interval box below it. This will make it easier for the user to read. This wording will reduce the amountof text needed and hopefully mitigate possible problems on smaller screens:

"Enqueues each Monday after 4pm"
"Enqueues the 1st of every month after 5pm"
"Enqueues each day after 5pm"


Changing the commit options to this would also be good:
image

Note: Use "Options" as a label instead of "Constraints" and use "no new commits". The options text above should be shrortened to read:
- "only if no new commit"
- "always run"

carlad:
All looking good. Only thing we need to do is make "Always run" the default option.

PR #557, opened 2016-06-06
Settings acceptance test: keys

backspace:
This is a continuation of my [previous acceptance test](https://github.com/travis-ci/travis-web/pull/540), where I noticed when I wrapping up work on it that I’d missed assertions for SSH keys. It’s still somewhat mysterious to me how all this works and what the two different key endpoints are about, but hopefully I’ll learn that during this exercise.


backspace:
Okay, I believe this is sufficient coverage for the settings page! 🎉

PR #559, opened 2016-06-07
dropdown menu only lists cron-free branches

chistopher:
This PR limits the availabe branches in the dropdown box to branches without crons. 
(branches on which you can create a new cron job)

To edit a cron job a user currently creates a new cron on the same branch thus overriding the old one. That is unintuitive and not documented. 

image

### DIFF
--- a/app/controllers/settings.js
+++ b/app/controllers/settings.js
@@ -9,7 +9,7 @@ export default Ember.Controller.extend({
         return branch.get('name') === cron.get('branch').get('name');
       });
     });
-  }.property('model.cronJobs','cronJobs','this.model.cronJobs'),
+  }.property('model.cronJobs.jobs.@each'),
 
   freeSortedBranches: Ember.computed.sort('freeBranches', function(a, b) {
     if(a.get('defaultBranch')) {
--- a/app/controllers/settings.js
+++ b/app/controllers/settings.js
@@ -2,19 +2,18 @@ import Ember from 'ember';
 
 export default Ember.Controller.extend({
   envVars: Ember.computed.filterBy('model.envVars', 'isNew', false),
-  freeBranches: function() {
+
+  branchesWithoutCron: Ember.computed('model.cronJobs.jobs.@each', function() {
     var cronJobs = this.get('model.cronJobs.jobs');
     var branches = this.get('model.branches').filter(function(branch) {
       return branch.get('exists_on_github');
     });
     return branches.filter(function(branch) {
-      return ! cronJobs.any(function(cron) {
-        return branch.get('name') === cron.get('branch').get('name');
-      });
+      return ! cronJobs.any(cron => branch.get('name') === cron.get('branch.name'));
     });
-  }.property('model.cronJobs.jobs.@each'),
+  }),
 
-  freeSortedBranches: Ember.computed.sort('freeBranches', function(a, b) {
+  sortedBranchesWithoutCron: Ember.computed.sort('branchesWithoutCron', function(a, b) {
     if(a.get('defaultBranch')) {
       return -1;
     } else if(b.get('defaultBranch')) {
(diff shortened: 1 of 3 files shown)


lislis:
What's the status of this? ping @chistopher @backspace

backspace:
hmm, not sure, @lislis. If you want this merged, @chistopher, you can assign me to it and I’ll review.

chistopher:
@backspace I can't assign people to this PR but a review would be great.

backspace:
oh I’m sorry, I forgot people outside the organisation can’t assign PRs! I’ll review this later today.

app/controllers/settings.js
```diff
@@ -2,16 +2,19 @@ import Ember from 'ember';
...
-    if(a.get('branch.defaultBranch')) {
-      return -1;
-    } else if(b.get('branch.defaultBranch')) {
-      return 1;
-    } else {
-      return a.get('branch.name') > b.get('branch.name');
-    }
-  }),
-  branches: Ember.computed.sort('model.branches', function(a, b) {
+  freeBranches: function() {
```
backspace: It’s my preference to use property names that clearly describe what they refer to. When you look at this here, it’s easy enough to decode that `freeBranches` is about crons, but in the template, you can see `freeSortedBranches` without any explanation of what is “free” branches. So what do you think of a name like `branchesWithoutCron` or something else that explains it’s cron-related?

app/controllers/settings.js
```diff
@@ -2,16 +2,19 @@ import Ember from 'ember';
...
+    var cronJobs = this.get('model.cronJobs.jobs');
+    var branches = this.get('model.branches').filter(function(branch) {
+      return branch.get('exists_on_github');
+    });
+    return branches.filter(function(branch) {
+      return ! cronJobs.any(function(cron) {
+        return branch.get('name') === cron.get('branch').get('name');
+      });
+    });
+  }.property('model.cronJobs.jobs.@each'),
```
backspace: Ember land is moving away from this computed property syntax and more toward the `Ember.computed`-type used elsewhere. It seems to be widely agreed that this syntax is confusing, and it requires “prototype extensions” that are falling out of favour. So my advice is to change this to just `Ember.computed('model.cronJobs.jobs.@each', function() ` up on the line where the property is declared. But this syntax appears elsewhere in the codebase so it’s not something I’m insisting upon, just a direction I’d like the codebase to move.

app/templates/components/add-cron-job.hbs
```diff
@@ -3,9 +3,7 @@
     <span class="label">Branch</span>
     {{#x-select value=selectedBranch required="true"}}
       {{#each branches as |branch|}}
-        {{#if branch.exists_on_github}}
-          {{#x-option value=branch}}{{branch.name}}{{/x-option}}
-        {{/if}}
+        {{#x-option value=branch}}{{branch.name}}{{/x-option}}
```
backspace: I love that you were able to simplify in this way. Moving the filtering outside the template is very appealing to me.

backspace:
Sorry for not getting to this yesterday, I was migrating to a new machine and everything dragged on dreadfully. But I’ve made some comments now, @chistopher, let me know if you have any thoughts!

app/controllers/settings.js
```diff
@@ -2,16 +2,19 @@ import Ember from 'ember';
...
-  branches: Ember.computed.sort('model.branches', function(a, b) {
+  freeBranches: function() {
+    var cronJobs = this.get('model.cronJobs.jobs');
+    var branches = this.get('model.branches').filter(function(branch) {
+      return branch.get('exists_on_github');
+    });
+    return branches.filter(function(branch) {
+      return ! cronJobs.any(function(cron) {
+        return branch.get('name') === cron.get('branch').get('name');
+      });
```
backspace: If you wanted to get more ES2015 and Ember-pro with this you could shorten to this:

```
return !cronJobs.any(cron => branch.get('name') === cron.get('branch.name'));
```

Concise and elegant!

chistopher:
Thanks for the help @backspace. I adapted the PR accordingly.

backspace:
Excellent, @chistopher! Merging now.

PR #560, opened 2016-06-07
Remove reference to campfire on empty-repos page

igorwwwwwwwwwwwwwwwwwwww:
We used to use campfire, but now we no longer do.


backspace:
👍🏻

PR #562, opened 2016-06-08
Add branches to requests page

lislis:
Sooo, I'm not sure why branches are currently not displayed on the requests page. Everything looks fine, but it's just not there?

screen shot 2016-06-08 at 16 00 37
This ^ is the view in question

### DIFF
--- a/app/components/requests-item.js
+++ b/app/components/requests-item.js
@@ -39,10 +39,6 @@ export default Ember.Component.extend({
     }
   }.property('request.isAccepted'),
 
-  hasBranchName: function() {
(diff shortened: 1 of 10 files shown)


lislis:
Update: there might be an inconsistency between v2 and v3 about the use of `branch` and `branch_name` that could potentially cause the troubles. Further investigation needed.

lislis:
So, it looks like this is really an issue with API not returning actual branch names for requests. If I manipulate the data locally it works just fine. Gonna open a ticket for API

PR #563, opened 2016-06-08
Branches route acceptance test

backspace:
Here is a beginning for an [acceptance test for the branches route](https://github.com/travis-ci/travis-web/pull/563). I’m hoping this one is less involved!

### DIFF
--- a/app/components/branch-row.js
+++ b/app/components/branch-row.js
@@ -49,9 +49,12 @@ export default Ember.Component.extend({
             array.push({});
           }
         }
-        lastBuilds.set('count', response['@pagination'].count);
-        lastBuilds.set('content', array);
-        return lastBuilds.set('isLoading', false);
+
+        Ember.run(() => {
+          lastBuilds.set('count', response['@pagination'].count);
+          lastBuilds.set('content', array);
+          lastBuilds.set('isLoading', false);
+        });
       });
     }
     return lastBuilds;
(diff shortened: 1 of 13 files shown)


backspace:
Hopes dashed!

It’s clear we need to improve the mock API so it properly serves related resources and includes pagination data. My latest assertion addition hardcodes a `@pagination.count` value that’s used by the component that fetches the builds for a branch (😳); probably most of the acceptance tests won’t use pagination at all, but it can be included in the mock API pretty easily, I suspect.

lislis:
<3 your mock data :)

backspace:
Extraneous comment to link to the [meta-issue for acceptance tests](https://github.com/travis-pro/team-teal/issues/687) because my edit to the initial comment didn’t do so 💔

backspace:
Well! I was wondering whether in this PR I’d get to a more general mock V3 API solution, but I only got partway. I’ve learned more about Mirage and think within the next acceptance test PR, I’ll write a V3 serialiser that will take care of most resource types.

You can see the beginnings of it in this PR, with the `user` serialiser, and particularly the `commit` serialiser. The latter just inherits Mirage’s `RestSerializer` and uses its built-in flags to control what is serialised.

I tried for a while to figure out how to get the other serialisers similarly simplified, but for now they are still pretty custom. I feel certain there’s a way to define something like a computed property on a Mirage model, so a `branch` can have a `lastBuild` that’s just the last item in the associated `builds` list, but my query on the Ember community Slack went unanswered and my experimentations led nowhere. `Object.defineProperty` seems like a good lead but I’m not sure where to put it to have it apply to all models.

It irks me to have that `FIXME` left in but I think I’ll be able to remove it soon!

PR #564, opened 2016-06-08
Remove unused route

backspace:
Since I enabled SSH keys in the test environment, I was seeing a lot of deprecation warnings about this. It appears to be unused? Do you agree, @drogus?


lislis:
I think it's fine to remove the route 👍 . It's a leftover from when each setting option had it's own route. We changed it to have all settings under `/settings`

backspace:
Excellent, thank you @lislis!

PR #568, opened 2016-06-10
Remove unneeded branch line for cache

acorncom:
Per https://github.com/travis-ci/travis-ci/issues/3039#issuecomment-119302093 it looks like the md5deep branch has been deployed, so this line is not needed ...


backspace:
Nice catch, thank you!

PR #569, opened 2016-06-10
Adding Anna's entry and assets to team page

saltinejustine:

PR #571, opened 2016-06-10
Adding Anna's entry and assets to team page

sinthetix:
This is #569 but just undoes the revert I did in #570. Sorry again about being too excited before considering possible consequences! 😬


lislis:
robert redford

PR #572, opened 2016-06-12
Make jobs tooltip wider

san650:
before
jobs-tooltip-before

after
jobs-tooltip-after


lislis:
Awesome! Thanks for catching that!
high five

PR #573, opened 2016-06-14
change from HS beacon to modal

lislis:
The HS beacon was not flexible enough styling-wise.
We're switching to the modal that let's us style an element
of our choosing to open the form.

Also updated the mascot in the noscript tag

### DIFF
--- a/app/app.js
+++ b/app/app.js
@@ -59,12 +59,7 @@ var App = Ember.Application.extend(Ember.Evented, {
   },
 
   setupBeacon() {
-
-    if (window.HS) {
-      HS.beacon.ready(function() {
-        HS.beacon.init();
-      });
-    } else {
+    if (!window.HS) {
       initHsBeacon();
     }
   },
@@ -83,6 +78,7 @@ var App = Ember.Application.extend(Ember.Evented, {
           return "private-" + channel;
         }
       });
+      
     }
     return Travis.pusher.subscribeAll(channels);
   },
--- a/app/index.html
+++ b/app/index.html
@@ -20,7 +20,7 @@
   <body>
     <noscript>
       <div style="width: 60%; margin: auto;">
-        <img src="/images/travis-mascot-150.png" alt="Travis CI mascot" style="float: left; margin: 1em 3em;">
+        <img src="/images/logos/TravisCI-Mascot-1.png" alt="Travis CI mascot" width="200" style="float: left; margin: 1em 3em;">
         <div>
           <h1>Hey there!</h1>
           <p>Looks like you have JavaScript disabled.</p>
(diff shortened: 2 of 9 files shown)


app/styles/app/modules/helpscout-link.sass
```diff
@@ -0,0 +1,21 @@
...
+  display: none
+  padding: 0.7em 1em 3.3em
+  z-index: 99
+  font-size: 16px
+  border: 2px solid $oxide-blue
+  border-radius: 2px
+  background-color: white
+  color: $oxide-blue
+  transform: translateY(2.7em)
+  will-change: transform
```
backspace: ooh, I didn’t know about this!

lislis: it's pretty neat. I draws the element on a new layer so when animation happens only that element needs to be repainted

app/app.js
```diff
@@ -22,6 +22,7 @@ var App = Ember.Application.extend(Ember.Evented, {
     }
 
     this.on('user:signed_in', function(user) {
+      $('body').addClass('signed-in');
```
backspace: It’s my impression that using jQuery to add/remove classes is counter to “the Ember Way”, because you can use a more declarative approach, tying the existence of the class to whatever data you‘re checking. Then there’s no need to manually add/remove classes, they’re just explicitly connected to the data model.

I’ve seen such uses of jQuery throughout the codebase though, as I’m adding acceptance tests, so it’s not a departure from the existing style. But eventually I’d like to move away from this pattern.

Just a note, not something I think needs to change in this PR!

app/index.html
```diff
@@ -36,5 +36,7 @@
     <script src="assets/travis.js"></script>
 
     {{content-for "body-footer"}}
+
+    <a href="mailto:support@travis-ci.com" title="Ask Travis CI support for help" id="helpscout-link" class="helpscout-link">Do you have a question?</a>
```
backspace: I’d usually expect this to be a component, which would let you wrap up the click-handling and show/hiding in a more Ember-y fashion. Buuuut maybe I am too focused on Ember conventions??? 😬

backspace:
Thus ends my review. If you are interested in component-ising the Help Scout trigger, I think that would be a more Ember-y approach. We could pair on that if you think that’s worthy, @lislis.

backspace:
Yes, I like this change, @lislis!

It’s true that I would usually test for this kind of thing, but I don’t think we have good infrastructure in place for setting com-vs-org in the test environment, so for now I think it’s okay to leave off.

lislis:
thanks @backspace I'll rebase and 🚢 then!

PR #577, opened 2016-06-17
Pull requests acceptance test

backspace:
This is another step in the [ongoing saga of adding acceptance tests](https://github.com/travis-pro/team-teal/issues/687).

It looks like this will overlap significantly with my previous pull request for the repository branches route. So I hope to extract some shared functionality. My other goals in this pull request are to:

* create a true V3 Mirage serialiser
* figure out how to get Mirage models to have dynamic properties ([more](https://github.com/travis-ci/travis-web/pull/563#issuecomment-226842717))
* maybe? extract fixtures or common setup code for Mirage factories

### DIFF
--- a/app/templates/components/builds-item.hbs
+++ b/app/templates/components/builds-item.hbs
@@ -34,7 +34,7 @@
       {{/link-to}}
     {{/if}}
   </h3>
-  <div class="row-item">
+  <div class="row-item row-commit">
     <a href="{{urlGithubCommit}}" title="See the commit on GitHub">
       <span class="icon-github"></span>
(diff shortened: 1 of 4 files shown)


backspace:
I think that’s it for the assertions. I could add more to show an in-progress build but I can’t even see in production how that works, so I won’t for now. I think it wouldn’t be much different than just checking the class to be `started`.

I declared above that I want to make a true V3 Mirage serialiser, but this route appears to not use V3 at all, or very little. But I can still do some mock API refactoring here.

backspace:
ACTUALLY I’m going to just merge this and carry out the refactoring in a separate pull request, since it’s not specific to the pull requests route.

PR #578, opened 2016-06-20
Refactor acceptance test mock API and data

backspace:
I’m taking a break from adding [acceptance tests](https://github.com/travis-pro/team-teal/issues/687) to refactor the mock API as it stands.

Goals:
* move more of it to use Mirage serialisers
* extract fixtures or somehow consolidate setup data currently heavily duplicated between tests

The Mirage serialiser trouble I foresee is having serialisers for both V2 and V3. 😳


backspace:
I likely won’t be able to make progress on this for a while, as I’m going to be offline for travel to and attendance of Wicked Good Ember. The current state is that I have separate V2 and V3 Mirage serialisers that are combined by a `TravisSerializer` that I plan to eventually make the default serialiser. However, I feel like I’m fighting the object model for Mirage serialisers and maybe this isn’t the best way. I could go into detail here but I suspect no one is reading!

Some of the Mirage routes are now “shorthands” which is nice, and my eventual goal for most routes.

Maybe I’ll see the creator of Mirage at the conference and can get some insight, because my request on the Ember Community Slack was presumably too esoteric for the #ec-mirage channel!

drogus:
> I could go into detail here but I suspect no one is reading!

Someone *is* reading, but with a slight delay!

I'm not really experienced with Mirage serialisers yet, but I think that we don't really have to make it work with both payloads. When I ship V3 usage for repos then we will need to return V3 for repos and V2 for pretty much everything else. Then the plan is to switch next endpoints to V3 and then we can switch Mirage serialisers at the same time.

I'm not sure how it will work with embedded records, though - is the problem laying here? If that's true (ie. you would like to use serialisers to serialise embedded records as well), I will need to have another look at that.

backspace:
haha good to know, 👋!

The Mirage serialiser system is actually pretty excellent, now that I’ve looked into it. It does indeed serialise embedded records, or sideloaded records, which is the trouble I was running into, because my weird prototype overrides for the V2 vs V3 serialisers weren’t working when serialising child records.

I did eventually get an answer from the creator of Mirage in the Ember Community Slack; he suggested using [this method to use a custom serialiser](http://www.ember-cli-mirage.com/docs/v0.2.x/route-handlers/#serialize). I only had a short time to try that out before I finished my work week, since I was heading south for the conference I’m now at. I didn’t get it working in that time, but I think I should be able to when the conference is over this week. Also, the creator is here and I hope to talk to him to confirm my plan with that method.

backspace:
I _just_ talked to Mirage’s creator :tada: and he confirmed my understanding of how to get the V2 vs V3 alternate serialisers working. It seems best to just construct separate trees for V2 and V3, with the custom serialisers for each inheriting from the base whenever `include`s are needed and the like. I’ll bless V3 as the default serialisers and when something uses V2, I’ll use `this.serialize` and pass in `v2branch` or whatever.

Also, for the question of the `last_build` that I currently have a custom `serialize` method for, he said that [`include`](http://www.ember-cli-mirage.com/docs/v0.2.x/serializers/#include) lets you pass a method rather than an array, which should let me need less customisation. (I think it’s undocumented so I’ll probably make a PR to add that.)

backspace:
Well! I’m not that satisfied with this refactoring, I’d like to have gotten further into full Mirageification, but it’s gone on pretty long already, and after this most recent brutal merge, I’m just going to move on. It seems like somewhat of a waste to work too much on cleanly handling V2 and V3 simultaneously.

I didn’t end up extracting the duplicated factories either but I really just _have_ to merge because I can’t risk another dreadful set of conflicts with `master`!

backspace:
oh yeah, I was happy to have [fully removed `turnIntoV3`](https://github.com/travis-ci/travis-web/pull/578/commits/4fc66d878ff809f8f4c7841dc6b420ec156828d0) in my refactoring, but it came back in from the merge with `master`, in the `repository` serialiser. I’m too DONE with this branch to figure out how to get rid of it for now, but eventually I think that serialiser should be merged with the `v3` one.

backspace:
ALSO of note is that this removes three mis-committed files:

* https://github.com/travis-ci/travis-web/blob/3f798567d0294b833b9cff0332cd812d23f552da/flow.txt
* https://github.com/travis-ci/travis-web/blob/3f798567d0294b833b9cff0332cd812d23f552da/repos_response.json
* https://github.com/travis-ci/travis-web/blob/3f798567d0294b833b9cff0332cd812d23f552da/then-change-tabs.txt

PR #581, opened 2016-06-27
Change to absolute paths

backspace:
This is a potential fix for [this](https://github.com/travis-pro/team-teal/issues/980).

### DIFF
--- a/app/templates/logo.hbs
+++ b/app/templates/logo.hbs
@@ -12,13 +12,13 @@
       <div class="logo-logos">
         <figure>
           <div class="logos--light">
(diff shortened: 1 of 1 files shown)


backspace:
This _didn’t_ fix the problem… on `master`, the first logo URL starts with `..`. So we thought removing the unnecessary `..`s might fix it. But after deploying this, the first logo had a leading `.` instead! v strange

image

This only happens in production and not staging, presumably because staging doesn’t use a CDN for assets. We were unable to reproduce locally.

backspace:
😆 it turns out I actually had a leading `.` on the `img` in question. Now that I removed that, the [logo loads in production](https://travis-ci.org/logo)! 🎉

PR #584, opened 2016-06-29
Changed Fritz's job title and animated gif, still dying laughing

saltinejustine:


BanzaiMan:
That is one epic GIF.

saltinejustine:
We'll drop this, see https://github.com/travis-pro/design-tasks/issues/35 for reasoning!

PR #590, opened 2016-07-07
Cron builds and build history acceptance tests

backspace:
Another entry in the [ongoing saga](https://github.com/travis-pro/team-teal/issues/687).

I mostly copied the pull requests acceptance test and page object for this, since the route is almost the same. Since this is so easy and the forthcoming build history route will also be similar, I’ll think about how to extract some duplication.


backspace:
Soooo I threw in an acceptance test for the Build History route because it’s almost identical to the Crons route. Now there are three routes that are near-identical. See [this diff](https://www.diffchecker.com/tkhqkez1) of the Build History vs Crons acceptance test files.

This feels weird to me but I’m not sure how to address it.

Maybe… I can just combine all three tests into one file, so the setup is shared, and then just make assertions about which builds show on which tabs. I think that’s a good answer…

backspace:
Okay, done!

image

🎉🎉🎉

PR #591, opened 2016-07-07
Remove v3 repo feature flag and conditionals

clekstro:
Given the recent rollout of v3 (this time w/o issues), it seems a good opportunity to remove some of the indirection and noise the feature flag conditionals are causing


backspace:
image

Always a delight!

PR #592, opened 2016-07-07
Move permissions logic from util to service

clekstro:
The service previously just wrapped the util, and all of the invocations currently reference the service.  This just gets rid of the middle man


tests/unit/services/permissions-test.js
```diff
@@ -1,12 +1,87 @@
...
+
+  assert.ok(service.hasPermission(1));
+  assert.notOk(service.hasPermission(2));
+
+  service.set('currentUser', null);
+  assert.notOk(service.hasPermission(1));
 });
 
-// Replace this with your real tests.
-test('it exists', function(assert) {
```
backspace: 😆

PR #596, opened 2016-07-11
Replace use of hooks across the application

clekstro:
Now that v3 repos has been officially released, we can remove the use of the `/hooks` endpoint entirely, and all of the code that relied on it.


drogus:
This should wait for https://github.com/travis-ci/travis-api/pull/297

drogus:
> This should wait for https://github.com/travis-ci/travis-api/pull/297

Just to clarify, when I wrote that I meant that deploying this should wait for the API fix, not that the work should be put on hold.

clekstro:
Yep, makes sense. I'll hopefully be able to wrap this up tomorrow and will keep an eye on https://github.com/travis-ci/travis-api/pull/297.  Thanks for the heads up @drogus!

app/components/repository-toggle.js
```diff
@@ -6,13 +6,13 @@ export default Ember.Component.extend({
...
+  active: alias('repo.active'),
   click() {
     this.sendAction('onToggle');
-    let hook = this.get('hook');
-    return hook.toggle().then((function() {}), () => {
-      this.toggleProperty('hook.active');
-      return this.sendAction('onToggleError', hook);
+    let repo = this.get('repo');
+    return repo.toggle().then((function() {}), () => {
+      this.toggleProperty('repo.active');
```
drogus: I think that this could be toggled in `toggle` in repo now.

clekstro:
Blocking this until we've added `push` permissions to the v3 repo payload.

- [ ] Use `store.filter` to filter by repos we have `push` access to.

app/components/repository-toggle.js
clekstro: Good point.

drogus:
@clekstro have you created a ticket for adding a new permission? I would like to suggest a bit different approach, because technically what you're aiming for is not a push permission, but the info if a user is a collaborator (or I'm not sure how to call it better). So, if you're, for example, added as a collaborator to a private repository with only read access, you should still see the repository on the sidebar. So when reporting this to API I think that we need something like `@is_collaborator: true`. Does it make sense?

clekstro:
@drogus I haven't yet. I think the approach you're suggesting is good, but I'm wondering if we should still add `push` and `pull` as they're currently used within the app?

drogus:
@clekstro I don't think that this would be a good way to go. The whole idea to introduce `@permissions` in the API was based on a problem with a disconnect between the API and the client. So let's say that you define that only people with push permissions can restart a build. If you return just `push`/`pull`/`admin` permissions from API, the client also needs to know that restart is possible only if you have `push` permission. Having each action defined in permissions makes it easier to keep client and server in sync - client doesn't care what actual permissions does a user have on GitHub, it only cares for what API tells it. So API becomes the source of truth.

That's why I think that we should open separate issues for `is_collaborator` and for new permissions for ssh keys (like `manage_ssh_key: true/false`).

drogus:
And also for restart/cancel I guess

clekstro:
Thanks @drogus. First, I totally agree that we can separate these into different issues. Also, I really appreciate the additional explanation around the permissions. For instance, the `read` permission sounded very similar to what we might expect from Github -- and our code currently leverages `push` and `pull` -- so I wasn't aware of the bigger picture here.  It definitely makes sense for clients not to have to care about those details and define their abilities themselves out of one (or more) Github permissions.

Are you aware of where we are with identifying which new permissions we'd like to expose, i.e. are there others beyond `manage_ssh_key` that come to mind?

backspace:
Does this seem salvageable, @clekstro?

clekstro:
@backspace Yeah, it's just blocked atm until the new permissions have been added to v3. The current status is obviously quite a ways behind, but the substance shouldn't be hard to apply to master later.

carlad:
see: https://github.com/travis-pro/team-teal/issues/1845

A `delete_key_pair` permission has been added to v3 on the repository level which, along with the `create_key_pair` permission, will show whether a user can create or delete a key_pair (aka what we used to call ssh_key)

clekstro:
@drogus was kind enough to point out that shipping this in its current state is a bad idea given there is an implicit limit on the v3 `/repos` endpoint of 100 records and this work doesn't yet handle pagination.

carlad:
I had a chat today with Joe and Drogus about the `is_collaborator` check. We think the best way forward, because of how api v3 permissions are modelled, is to create an `is_collaborator` permission. It will basically check for pull/read access to a repository. We thought this the best approach as `is_collaborator` is a permission connected to a user rather than a repo, so would be better added as a permission.

carlad:
@drogus @backspace @clekstro Here's a PR that adds an `is_collaborator` permission to v3 api.

Can you test this branch of api with the work you're doing here and let me know if it's what you need.

clekstro:
Missing pagination, potential styling issues, but otherwise functional.

clekstro:
Pagination now working, but @drogus likely has a better idea of how to approach this. The design could use some work, given that we now show activateable/non-activateable repositories in the same list and need a way to better visually distinguish them.

Here's a screenshot to show what I mean:
screen shot 2017-08-31 at 4 35 53 pm

I haven't started looking into broken tests yet, likely just the hook logic that's failing due to the new layout.

clekstro:
I've fixed some of the styles to be more consistent with the previous implementation, though I do wonder if we may need even more contrast to avoid confusing users.

clekstro:
Here's what things look like currently:
screen shot 2017-09-12 at 12 03 36 pm

@ctrlaltjustine Do you have any suggestions?

ctrlaltjustine:
They're currently listed in alphabetical order. It seems to me the behavior and styles should be similar to what we have on Dashboard with "Active repositories" and "Inactive repositories" where they are broke up by those respective categories. 

This of course has the limitation/potential snag that in orgs (such as Travis CI's) where you have many activateable repositories, it may take a couple pages before you find a repo you were looking for, because it happens to be an non-activateable repo. But it's consistent with the way we break up the information in other parts of our UI.

activateable-repositories

clekstro:
@ctrlaltjustine My bad, I thought I had pinged you in the [Teal issue](https://github.com/travis-pro/team-teal/issues/578) regarding this where I [laid out my reasoning](https://github.com/travis-pro/team-teal/issues/578#issuecomment-326238500) for going with this approach.

> This of course has the limitation/potential snag that in orgs (such as Travis CI's) where you have many activateable repositories, it may take a couple pages before you find a repo you were looking for

Yep, totally. This was one of the main reasons I opted against this. The other is that the division, depending on how pagination is done (having two controls, one per section is not currently possible as we can't currently filter api requests by permissions), will potentially be very awkward  aesthetically. You might have ratios of 1/24 on page 1 and 24/1 on the next which will be visually jarring. Even having 0/25 would be possible, which is just another edge case to handle in terms of coming up with copy for that case.

> They're currently listed in alphabetical order. It seems to me the behavior and styles should be similar to what we have on Dashboard with "Active repositories" and "Inactive repositories" where they are broke up by those respective categories

I'm all for consistency, but I think we're trying to solve two different problems here. On the Dashboard, you want the "Active Repositories" first, because you're most interested in seeing the status of those repositories. So we show you the most relevant information first. In the case of activation/deactivation, you're just trying to find the repository as efficiently as possible. To me, the two groups feel like an artificial hurdle to accomplishing the main goal of this page. Also, given the dashboard is behind a feature flag, I don't think anything close to the majority of our users are aware of this inconsistency yet, though I'd have to look at the metrics to confirm.

ctrlaltjustine:
@clekstro that all makes sense.

Here are two options looking at this problem from that angle: 

- Create an additional icon (similar to our, cannot delete icon) with the cog that supports the dimmed style for non-activateable repos? The perceivable issue with this would be it's a bit hard to see dimmed down.

activateable-repositories-02

- Add a label next to non-activateable repos similar to cron jobs and trials. I'm not sure if that's the copy we'd want to use, but labels are something we're already using and are more explicit than icons.

activateable-repositories-03

clekstro:
@ctrlaltjustine Thanks for the feedback and the suggestions! I think I prefer the first one; the second feels a little busy with the additional label and like we're missing a chance to communicate our intent in the first two icons (the switch and the gear), though it does have the advantage of being more explicit. 

As this seems like it's going to be a more common issue moving forward now that V3 supports fine-grained permissions, it might make sense to come up with something unique across the app that implies you don't have the necessary permissions (a user avatar with a lock?), as this probably won't be the last case where we show a UI element that you can't interact with (but need to know that information instead of just hiding it)

Here's a screenshot showing a few more alternatives: a combination of the private/settings icons as a proof of concept (plz ignore my very basic SVG skills), the lock icon itself, as well as a grey-filled version of the settings icon:
 
screen shot 2017-09-14 at 3 26 41 pm

All that said, I'd be happy to defer to whatever approach you think best. Thanks!

ctrlaltjustine:
@clekstro thanks very much for mocking that up and the explanation. As much as I like the idea of using a cog with a lock, I'm afraid it's not super readable at such a small size. I've also tried to create my own version but with no success. With that in mind, I think the standalone lock icon is a good way to go. Thanks for this.

clekstro:
There were some diffs in the Percy snapshots regarding the orgs in the tests, but I wasn't able to reproduce in the app either locally or in PR envs, so probably just some test data issue that shouldn't hold up this PR #featurefreeze. Merging.

PR #597, opened 2016-07-11
Remove unused code

backspace:
I came across this while looking for code that would be improved by `ember-concurrency`. Could it truly be unused?? Thoughts, @lislis, @drogus, and/or @clekstro? I don’t see the `tiggerBuild` action being called anywhere, nor the `triggerBuild` non-action and the `canTrigger` property. A legacy of removed code perhaps?

### DIFF
--- a/app/components/branch-row.js
+++ b/app/components/branch-row.js
@@ -60,39 +60,7 @@ export default Ember.Component.extend({
     return lastBuilds;
   }.property(),
 
-  canTrigger: function() {
-    return this.get('permissions').hasPermission(this.get('branch.repository'));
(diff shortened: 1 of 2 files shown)


clekstro:
Yeah, this looks great 👍. Just as a heads up, I rebased master in and force pushed on this branch to ensure your fix from last night is included, as I wanted to check this out in the PR env.

drogus:
This was used on one of the versions of the dashboard, I think that it's safe to remove for now.

backspace:
Thanks for that, @clekstro. I’m glad the PR deployments are useful for others, it was a 😳 moment for me when I saw it wasn’t working.

PR #598, opened 2016-07-11
Add ember-concurrency

backspace:
So far, I’ve added `task`s in some obvious places, by searching for `this.get('is`, which usually signifies that a flag is being set and checked to prevent duplicate firing of asynchronous events, such as updates and deletes. You can see that `ember-concurrency` reduces the boilerplate around this pattern.

(Some of the line savings and shortening is because I’m replacing things with template strings too. 😁)

It would likely be useful elsewhere in the codebase, but for now I’ll just clean up some of the simple cases.

### DIFF
--- a/app/components/add-cron-job.js
+++ b/app/components/add-cron-job.js
@@ -2,6 +2,7 @@ import Ember from 'ember';
 import config from 'travis/config/environment';
 
 const { service } = Ember.inject;
+import { task } from 'ember-concurrency';
 
 export default Ember.Component.extend({
   classNames: ['form--cron'],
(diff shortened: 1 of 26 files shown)


backspace:
All the tests are passing, but I don’t have _that_ much confidence in the test suite to just deploy this and move on…

aaaand the test deployment won’t boot, nor did it when I deployed it to org-staging. So I’ll try to figure out what might be going wrong. Strange!

backspace:
After that now-resolved [side quest](https://github.com/travis-ci/travis-web/pull/599), I’ll try deploying this tomorrow morning so I can roll back if there’s drama.

clekstro:
Looks great!

PR #599, opened 2016-07-11
Debug site launch failure

backspace:
As I mentioned [here](https://github.com/travis-ci/travis-web/pull/598#issuecomment-231853066), `master` is failing to deploy. This is an experimental branch to see why this might be happening. Since the Javascript error relates to initialisers, I’m removing the one I added recently. This will cause the tests to fail, but I can force a deployment to staging.


backspace:
Yes, that worked. Merging this fix! `master` should always be deployable 😳

PR #600, opened 2016-07-12
Subscribe new repositories to pusher

drogus:


joshk:
will this work for Enterprise as well? (where all pusher communication should be via private channels?)

drogus:
@joshk oops, I haven't seen your comment when I clicked merge. I'm not sure to be honest, I wasn't aware that even public repos use private channes on enterprise, I'll double check it

joshk:
@drogus hey, sorry, did you see my comment above?

joshk:
Ah, ok. For reference, with Enterprise you can have a 'public' repo, but all communication from us should be 'private'.

joshk:
It might mean that you need to check if this is an Enterprise install.

drogus:
This should be fixed with https://github.com/travis-ci/travis-web/pull/601

PR #603, opened 2016-07-13
Enforce basic styles with ESLint

clekstro:
This adds ESLint to the project and converts our current code base over to following a very rudimentary set of linting rules.

While this is only a first pass -- we can continue to discuss and enhance this over time -- I think this is already worthy of integrating within the project (I fixed something like 300+ warnings).  Unless there are major concerns as to how those fixes were made, I'd suggest we address them in separate issues to avoid a lengthy back and forth with merge conflicts, as this touches most of the code in our project.

I should also note that this gave me the opportunity to see a lot of the code base I hadn't seen yet. I added `TODO`'s for items that I think we should address in the future.


app/adapters/application.js
```diff
@@ -45,12 +44,14 @@ export default ActiveModelAdapter.extend({
...
   handleResponse(status, headers, payload) {
     if (status > 299) {
+      // eslint-disable-next-line
       console.log("[ERROR] API responded with an error (" + status + "): " + (JSON.stringify(payload)));
     }
 
     return this._super(...arguments);
   },
 
+  // TODO: remove this now that this has been merged!
```
backspace: hehe I noted this myself in [iDoneThis](https://idonethis.com/done/28001896/)

backspace:
I didn’t survey all the changes, but I’m glad to see this, its absence has been bothering me.

I definitely agree regarding merging quickly and making additions later, I haaaate merge conflicts. I’m sorry for the inevitable conflicts that’ll happen when I merge the concurrency branch. It might be easier to just accept my branch’s version of conflicted files and make whatever changes ESLint demands after that, though you can do whatever works of course.

clekstro:
It looks like there are a few issues I've introduced with this work, especially as concerns restarting builds/jobs. I'll look into them more in the am:

- [ ] Add additional UI assertions to our acceptance tests so we have more confidence in our test suite (i.e. I broke sth but all our tests are passing)
- [ ] Fix state not updating after triggering build restart
- [ ] Fix flash message showing "Build has been restarted" when restarting job (not sure if this is only a problem on this branch or not)

My best guess here is that I followed the linter's suggestion precisely and that changed a function signature that's somehow borking things.

backspace:
Insomnia? 😆

PR #606, opened 2016-07-14
Fix Ember Inspector route listing

backspace:
As noted here: https://github.com/emberjs/ember-inspector/issues/521#issuecomment-190359087

The way the custom location implementation was installed in the router was causing the Ember Inspector to be unable to load the route hierarchy. A string is expected for the location implementation name, to be loaded from app/locations, according to Ember CLI conventions. This also lets the location itself load the auth service, via the `getOwner` API.


backspace:
Hey, @drogus, does this look like a reasonable change? It’s helpful to me and apparently [random others](https://github.com/emberjs/ember-inspector/issues/521#issuecomment-190359087) to have a working route inspector.

backspace:
Hey, @drogus, does this look like a reasonable change? It’s helpful to me and apparently [random others](https://github.com/emberjs/ember-inspector/issues/521#issuecomment-190359087) to have a working route inspector.

backspace:
I deployed this to both org-staging and com-staging. The routes for `/` were as expected both when logged in and not and the Ember Inspector route listing works. So I’m going to merge.

PR #616, opened 2016-07-19
Trying to see if Pusher works on .com PR envs with current master

clekstro:


clekstro:
It looks like we get some Pusher authentication errors. I'm not sure if we have to specify which domains to allow, but given this same code runs without issues in production, it'd be good to get total parity between environments. 

For what it's worth, com-staging appears to have similar issues. @drogus @backspace any ideas?

clekstro:
@backspace Just heard from @svenfuchs that we also see this from time to time in production, but very sporadically. My hope is that if we can track it down in this environment, then we can solve it there as well...

clekstro:
screenshot 2016-07-29 15 50 18

clekstro:
And the specific error message:
screenshot 2016-07-29 15 51 17

backspace:
hmm, I somehow missed this until now, sorry. My suspicion is also about permitted domains, but when I created an account on Pusher I couldn’t see anything about restricting domains, nor can I in the documentation. I do see things about `authEndpoint` being `/pusher/auth`, which it’s possible is misconfigured or not properly present on the PR deployments server (`travis-pro-web-production-next`). I was able to examine the Pusher configuration by putting a breakpoint in `getDefaultStrategy` in `/app/utils/pusher.js`.

Is there an open issue for this?

clekstro:
@backspace i think this is related to the `/pusher/auth` stuff I mentioned yesterday. Going to close this now that we no longer see it.

backspace:
Great, I’m glad it’s fixed!

PR #617, opened 2016-07-19
Don't add `private-` suffix manually to pusher channels

drogus:
We already have [code](https://github.com/travis-ci/travis-web/blob/f500c1eb225f087b9eed992ed970facacc71d611/app/utils/pusher.js#L73-L81) that automatically adds `config.pusher.channelPrefix`
to all pusher channels if it's set

### DIFF
--- a/app/components/hook-switch.js
+++ b/app/components/hook-switch.js
@@ -16,11 +16,7 @@ export default Ember.Component.extend({
       repoId = hook.get('id');
 
     return hook.toggle().then(() => {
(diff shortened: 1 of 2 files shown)


clekstro:
Can we do something similar in [app/components/not-active.js](https://github.com/travis-ci/travis-web/blob/master/app/components/not-active.js#L43-L49)?

drogus:
@clekstro it's in this PR already

clekstro:
Sorry, I'm blind. Disregard :)

PR #618, opened 2016-07-19
Add (nearly all of) Airbnb's ESLint rules

clekstro:
While this mostly works functionality-wise, this breaks a bunch of tests (and is more invasive than I'd like it to be). I'll probably end up breaking this up into several pieces, but haven't come up with the best approach yet.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,2 @@
+node_modules/
+bower_components/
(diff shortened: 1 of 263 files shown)


clekstro:
Going to close this and open up individual PR's for the smaller pieces.

PR #619, opened 2016-07-20
Add fastboot

clekstro:
Ember network is necessary because we do not have access to $.ajax() within node.
This provides a fetch() that is compatible with the browser API to make it run in both
the browser and node.

We could remove this if we ported all network calls to Ember Data in the future, but I do not
think we are ready for that yet (maybe never due to things like logs).

This is totally WIP and is not close to a running state, just wanted to get it out there.

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,6 +33,7 @@
     "ember-cli-deprecation-workflow": "0.2.3",
     "ember-cli-document-title": "0.3.1",
     "ember-cli-eslint": "1.7.0",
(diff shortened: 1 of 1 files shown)


drogus:
> We could remove this if we ported all network calls to Ember Data in the future, but I do not
think we are ready for that yet (maybe never due to things like logs)

Couldn't we also use `fetch()` for logs?

clekstro:
@drogus Definitely. I more meant that we could get rid of the ember-network dependency entirely if all calls were using Ember Data.

PR #623, opened 2016-07-22
Remove unused/invalid alt attribute for {{link-to}} helpers

clekstro:
I think this can be removed. It doesn't appear to be used, and if it is, seems like it's actually invalid HTML?

### DIFF
--- a/app/app.js
+++ b/app/app.js
@@ -8,10 +8,6 @@ import initHsBeacon from 'travis/utils/init-hs-beacon';
 
 Ember.MODEL_FACTORY_INJECTIONS = true;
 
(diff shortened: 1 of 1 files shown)


backspace:
It looks to me that [`top.hbs`](https://github.com/travis-ci/travis-web/blob/5a42ba9f90d38e262f93d9b2fe3727c05fc857ca/app/templates/top.hbs#L6) is the only place this is used, unless there’s a `link-to` on multiple lines that escaped my regex. Do you want to remove the `alt` attributes from that file too, @clekstro? Or I could do it also.

PR #632, opened 2016-07-27
Upgrade Ember-CLI to 2.7.0

clekstro:
The only thing to note here is the change from using `baseURL` to `rootURL`. It doesn't look like waiter touches that, meaning we shouldn't be impacted in production.


clekstro:
I'm seeing some odd behavior when refreshing the page in our PR environments. It looks like it is signing us out :\ for some strange reason. Going to see if this impacts other PR's (and so is specific to the env and not code).

clekstro:
This appears to impact #623 as well, so doesn't look like it has anything to do with these changes specifically. Going to test on staging and then deploy to production if everything looks kosher.

clekstro:
Looks good on staging. Merging.

PR #633, opened 2016-07-27
Fix path loaded for customers using .com on login

clekstro:


clekstro:
This appears to fix [this issue](https://github.com/travis-pro/team-teal/issues/1093). Merging.

PR #634, opened 2016-07-27
Enforce Airbnb ES6 styles for app/components

clekstro:


clekstro:
This doesn't seem to have caused any issues. Will merge and deploy in the am.

PR #638, opened 2016-07-29
Try out Glimmer 2 alpha

backspace:
I changed the `.travis.yml` so that a deployment will occur even when the tests fail. (For me, locally, it’s passing 373/498.)


backspace:
hah well _nothing_ renders. I’m not surprised, since this is the first release. Maybe next week I can look into it to see what might be wrong and report back to Ember people.

clekstro:
Ha, I was just doing this locally 👍

PR #639, opened 2016-07-31
Adding Lena to the teampage

saltinejustine:


lislis:
@saltinejustine there was an open PR for this https://github.com/travis-ci/travis-web/pull/574 that only waited for the assets :D But I can use this instead 💃

PR #646, opened 2016-08-05
Improve UX on profile page

lislis:
- link repos to Travis and not to Github
- refactor some component/template logic

### DIFF
--- a/app/components/hook-switch.js
+++ b/app/components/hook-switch.js
@@ -1,25 +1,24 @@
 import Ember from 'ember';
 
-const { alias } = Ember.computed;
-
 export default Ember.Component.extend({
   tagName: 'a',
   classNames: ['switch--icon'],
-  classNameBindings: ['active'],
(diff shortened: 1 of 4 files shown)


app/components/hook-switch.js
```diff
@@ -1,25 +1,24 @@
 import Ember from 'ember';
 
-const { alias } = Ember.computed;
-
 export default Ember.Component.extend({
   tagName: 'a',
   classNames: ['switch--icon'],
-  classNameBindings: ['active'],
-  active: alias('hook.active'),
+  classNameBindings: ['hook.active:active', 'disabled:disabled', 'disabled:inline-block'],
```
clekstro: I think we can shorten `'disabled:disabled'` to just `'disabled'`. If the value is a boolean (which we're implying when using the `:` syntax, then it will simply take the name of the property, so the `disabled` class would be applied.

clekstro:
@lislis 🚀

PR #647, opened 2016-08-08
Run ImageOptim against public/images/emoji

clekstro:
This saves about 30k, which is not a ton, but why not?


clekstro:
Here's the output from ImageOptim with quality at 80%. I honestly can't tell the difference when looking at the diffs.
screenshot 2016-08-08 14 40 43

lislis:
amaze ✨

clekstro:
@lislis Actually curious about something related. I noticed that the current size of the packaged emojis is 64x64, but we display that as 20x20 in the UI.

[Here's a build](https://travis-ci.org/travis-infrastructure/packer-build/builds/152576663) that shows emoji in the commit message.

Here's the original size:


I'm viewing this on a huge monitor, so I'm skeptical that we actually show larger versions of this. Do you think it'd be possible if we resized everything down to 20x20? I think that'd probably save even more space than the optimizations above.

clekstro:
Apparently the display of these images is broken within our PR envs for some reason. I can confirm that the images are part of the build, so not sure what's up. Probably has to do with the waiter config, but the badges are working so ❓

clekstro:
Just to give a visual idea of the differences here after resizing to 20x20 and optimizing:
screenshot 2016-08-16 13 48 36

And current production:
screenshot 2016-08-16 13 49 11

Here's the asset savings (output using `du -h`):
Master: 
`5.6M    public/images/emoji`
This branch:
`3.5M    public/images/emoji`

@lislis @saltinejustine Any concerns about this?

lislis:
Nice!
Finger guns!

PR #651, opened 2016-08-18
Fix profile page redirect by overriding basic route behavior

clekstro:
Given the invasiveness of the change, it would be good to overhaul this
as soon as possible.


clekstro:
This should address [this issue](https://github.com/travis-pro/team-teal/issues/1184).

PR #654, opened 2016-08-19
Fix broken search tab behavior

clekstro:
This fixes the issue, but also converts our requests to using ember-concurrency.


clekstro:
This appears to be working as expected and to fix https://github.com/travis-pro/team-teal/issues/1194, though the behavior of the .com PR environment is somewhat concerning. Lots of pusher errors (shouldn't at all be related to this work).

clekstro:
My only concern about this currently is what we're showing when we perform a search and show the first result.  Here's a screenshot of what we'd currently display.
screenshot 2016-08-22 11 48 05

I've compared this behavior with the state before this was broken, and it appears to be consistent with the behavior after we shipped "View Config". I think we can open up a new issue if this isn't what we'd ideally like to be showing to users.

I'm going to merge as this is a definite improvement over the current state of things.

PR #655, opened 2016-08-21
always encrypt websocket connection to pusher/slanger

rkh:
This is useful for running a local dev instance of web against a production enterprise setup.

/cc @bnferguson


waiter/lib/travis/web/app.rb
```diff
@@ -198,6 +198,7 @@ def set_config(string, opts = {})
       pusher['host'] = options[:pusher_host] if options[:pusher_host]
       pusher['path'] = options[:pusher_path] if options[:pusher_path]
       pusher['channelPrefix'] = options[:pusher_channel_prefix] if options[:pusher_channel_prefix]
+      pusher['encrypted'] = true
```
clekstro: This won't change how things work within .org, right?

rkh: `encrypted` is currently set to `nil`/`null` (due to not being set at all), which means it defaults to `true` on any web instance served over ssl (like on staging/production) and to `false` on any not running over ssl (ie, localhost), so running locally will now always use ssl, but since pusher supports both modes this should not cause issues.

The only scenario that doesn't work is running Slanger (used by Enterprise) locally without SSL and then running web against this. For Enterprise development running Slanger over SSL is an option though.

PR #657, opened 2016-08-22
Prevent `tokenVisibility` from bubbling.

rwjblue:
Prevent `return true;` which causes the action to be bubbled up to the routing sytem (and no other handlers exist to handle it).


backspace:
Excellent, thanks.

PR #659, opened 2016-08-25
some minor bug fixes

lislis:
- remove duplicate subject line from PR build header
- use darker color for sync button pro profile page


lislis:
sweet, thanks!

PR #665, opened 2016-08-30
User-facing feature flags

clekstro:
This will allow users to experiment with new features and give us feedback before we roll them out for everyone.


clekstro:
Here's a sample of the current UI I've hacked together (with a sample feature toggled):
featureflags

Would love to start getting feedback as it's functional at this point. Right now just having issues when deploying while Mirage is enabled in production mode.

backspace:
That UI looks fine to me! I love the pink header example 😆

clekstro:
Need to fix whatever tests are currently broken here as well as get the deployment issues resolved so we can start to merge @lislis' work in with this.

clekstro:
The deploy now works as expected as long as Mirage is disabled.  I wonder if this might have something to do with the host/endpoint config. However, it's strange that we don't even get to the point where the app boots.

clekstro:
Ok, the app currently boots, but we're not seeing the same behavior in PR envs that we see when running in production mode locally. The requests to `/features` aren't being properly intercepted by Mirage.

clekstro:
Closing in favor of #683.

PR #669, opened 2016-09-06
Reset log truncation flag after job restart

backspace:
Restarted jobs that formerly had truncated logs were showing as truncated immediately, even when they only had a few log lines showing. Log truncation was probably added before the
job reset UI, so when truncation broke, no handling was added for resetting the limited flag when a job was restarted.

Adding automated testing for this might be difficult, but I’ll try it in my separate pull request for overall truncation tests.


backspace:
Here‘s a GIF (with some waiting edited out) of it working on org-production:

untruncation

Merging.

PR #670, opened 2016-09-06
Add assertions for log truncation

backspace:
This is mysteriously broken, but hopefully I’ll figure that out soon.

### DIFF
--- a/app/components/log-content.js
+++ b/app/components/log-content.js
@@ -123,9 +123,9 @@ export default Ember.Component.extend({
           this.unfoldHighlight();
         }
       });
-      this.limit = new Log.Limit(Log.LIMIT, () => {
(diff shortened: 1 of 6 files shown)


backspace:
I’m not too excited about how I accomplished this: with the addition of a `waitForElement` helper. However, since the rendering of log elements happens outside Ember facilities, I couldn’t think of a better way that didn’t involve adding a hook inside the still-mysterious `Log` class, and it would have required a `wait`-y approach anyway.

I ended up removing the assertion that the truncation message disappears after a job is restarted because it was failing… I think some external event is necessary for the `log-content` component to be refreshed but I couldn’t figure out the specifics and I felt like this had dragged on too long already. :cry:

PR #672, opened 2016-09-07
remove notes.txt

lislis:


backspace:
yay, thanks! I noticed that early on but never got around to removing it 😁

clekstro:
Thanks @lislis!

PR #673, opened 2016-09-07
Feature debug job

lislis:
It's happening 🎊 
This includes icon and styles for the button (build buttons are changed as well). It is visible for jobs on .com.
It looks like this:
screen shot 2016-09-07 at 16 39 21

PR #674, opened 2016-09-07
Add backend debugging support

backspace:
This is an addendum to [@lislis’s design work](https://github.com/travis-ci/travis-web/pull/673). It’s a naïve implementation for now, it just copies the code for the job restart action. It also includes a commit that fixes some selectors for tests that broke due to the design adjustments.


joshk:
THIS IS AWESOME!

joshk:
Small bug I found: 

screenshot 2016-09-07 18 29 36

If the build only has one job, then only `Restart build` is shown.

backspace:
haha omgtoosoon 😂

The feature flag `proVersion` isn’t enabled in the current deployed version. I’ve just pushed a commit that I hope will fix that. You can manually enable the feature flag via the Ember Inspector but it’s pretty involved to describe so I won’t go into that unless this possible fix doesn’t work.

joshk:
hahaha, sorry for being so quick, it make me all happy and excited :)

backspace:
Some code review notes:

* I’m not happy about the `TRAVIS_PRO=true ` addition to the com-production test deployment script but it doesn’t seem worth changing how these flags are handled in the test deployments as I don’t expect that to last
* I made a “task group” for debug/restart because one precludes the other; I called them `restarters` which isn’t my favourite but other options didn’t satisfy me either and even the [example in the documentation](http://ember-concurrency.com/#/docs/task-groups) uses a noun for the group despite its members being verbs
* the call to the V3 API through the Ajax service includes a header to trigger it on V3, which looks unsightly to me, but doesn’t seem worth extracting as it’s the only such instance

Refactoring that came along for the ride:
* I found and deleted the `repo-actions` component, which seemed superfluous
* I converted some manual `and`-y computed properties to just use the built-in `and`

backspace:
@joshk just asked me IRL to deploy this so I’m merging 😆

amarkowitz:
Hello there.  I actually just noticed this feature in the wild and decided to try it out because I got so pumped to see it 😹  Logging in via ssh works great and it's super sweet to be able to this.  

That said, I ran into an issue where once I ran a job as a `Debug job`, there was no way to get back to a non-Debug job.  i.e.  I tried just restarting, I tried canceling all jobs for my build and re-running all of them via `Restart Build`, but the single job I did a `Debug job` run on previously stayed in debug mode with no way to get out of it.

Let me know if you want the url (rather not post publicly if possible) or what else I can do to help track this down for you.

backspace:
eep. Thanks for letting us know, @amarkowitz. Could you write in to support@travis-ci.org with the URL and we’ll look into it?

amarkowitz:
sure thing!

amarkowitz:
I received this response:
> Hello Adam,
> Thank you for writing in and for testing out the debug job feature. I'm sorry for the hassle.
> Unfortunately, at the moment, once a job has been restarted in debug mode, that state prevails and it's considered a debug job. 
> That said, and since this is a feature we're slowly rolling out, I'll make sure to pass your feedback along to the team to see what they think and how we can improve this behaviour and expectations in the future.
> ...
Just wanted to be sure you got the info you need if this is actually not working as intended.

/cc @backspace

PR #681, opened 2016-09-12
Remove profile controller dependencies

backspace:
As part of [this meta-issue](https://github.com/travis-pro/team-teal/issues/871), this aims to remove controller dependencies within the `profile` controller. I haven’t yet actually removed all the dependencies, I’m still trying to untangle everything.


backspace:
Ultimately, the `profile` controller seems entirely superfluous. It was formerly responsible for switching between User and Hooks tabs, but the user tab is apparently unused, its template missing. There was a dependency chain with the `account` route calling the `profile` controller to activate the hooks tab, which incidentally reloaded the hooks; now, the account route can just reload the hooks itself.

I’ve clicked around with this and have seen no breakage so it seems like it should be deployable but I’m kind of scared! Want to check it out too, @clekstro? It seems unlikely to be subject to annoying merge conflicts though so I can save it for next week otherwise.

clekstro:
@backspace Thanks for already picking this up! I'll have a look later today after I've gotten the dashboard/feature flags stuff taken care of.

clekstro:
Great work @backspace! I've played around with the app locally as well and haven't noticed anything out of place, so I'd be comfortable merging as-is. I'm around in case there's any fallout.

Related: I think we can remove `app/templates/accounts.hbs` as well, right?

backspace:
Good point re `accounts.hbs`, gone now. Merging!

PR #678, opened 2016-09-12
Show next and last run times for crons

aakritigupta:
This PR is a part of the a big [re-factoring of cron jobs](https://github.com/travis-ci/travis-api/pull/319) and it does the following:
- Show next and last run times for cron jobs
- Add tests for last and next runs
- Change text for cron job running condition
- Fix option string: `dont_run_if_recent_build_exists`
- [x] Awaiting @joshk review of https://github.com/travis-pro/team-teal/issues/1248

PR #687, opened 2016-09-14
Add org-staging PR deployments

backspace:
@aakritigupta had use for this. It may be Too Much to have four deployments per PR given how slow they can be. To discuss.

### DIFF
--- a/config/deploy.js
+++ b/config/deploy.js
@@ -1,4 +1,5 @@
 var VALID_DEPLOY_TARGETS = [
+  'org-staging-pull-request',
   'org-production-pull-request',
   'com-production-pull-request'
 ];
@@ -26,6 +27,12 @@ module.exports = function(deployTarget) {
   if (deployTarget === 'org-production-pull-request') {
     ENV.s3.bucket = 'travis-web-production-next';
     ENV.redis.url = process.env.ORG_PRODUCTION_REDIS_URL;
+  }
+
+  if (deployTarget === 'org-staging-pull-request') {
+    ENV.s3.bucket = 'travis-web-production-next';
+    ENV.redis.url = process.env.ORG_PRODUCTION_REDIS_URL;
+    ENV.redis.keyPrefix = `${process.env.TRAVIS_PULL_REQUEST_BRANCH}-staging`;
   }
 
   if (deployTarget === 'com-production-pull-request') {
--- a/config/deployment/update-github-status.sh
+++ b/config/deployment/update-github-status.sh
@@ -4,6 +4,10 @@ curl -X POST \
      -H "Authorization: token $GITHUB_TOKEN" \
      https://api.github.com/repos/travis-ci/travis-web/statuses/$TRAVIS_PULL_REQUEST_COMMIT
 curl -X POST \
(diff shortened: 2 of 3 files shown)


config/deploy.js
```diff
@@ -28,6 +29,12 @@ module.exports = function(deployTarget) {
     ENV.redis.url = process.env.ORG_PRODUCTION_REDIS_URL;
   }
 
+  if (deployTarget === 'org-staging-pull-request') {
+    ENV.s3.bucket = 'travis-web-production-next';
+    ENV.redis.url = process.env.ORG_PRODUCTION_REDIS_URL;
```
backspace: This instance of Redis stores the `index.html` file that links to the JS and CSS assets a particular deployment of `travis-web` is composed of. This is known as the “lightning” strategy of deployment for Ember CLI applications.

Since these are all ephemeral deployments produced by `travis-web` pull requests, it seems harmless to me to use the same instance of Redis. It’s really just a collection of HTML files.

More on this below.

config/deployment/update-github-status.sh
```diff
@@ -4,6 +4,10 @@ curl -X POST \
      -H "Authorization: token $GITHUB_TOKEN" \
      https://api.github.com/repos/travis-ci/travis-web/statuses/$TRAVIS_PULL_REQUEST_COMMIT
 curl -X POST \
+     --data "{\"state\": \"success\", \"target_url\": \"https://`echo $TRAVIS_PULL_REQUEST_BRANCH`-staging.test-deployments.travis-ci.org\", \"description\": \"Visit a org-staging deployment for this commit\", \"context\": \"deployments/org-staging\"}" \
```
backspace: I decided to use the same domain for this as the `org-production` deployments to avoid having to get another wildcard SSL certificate.

I see that I haven’t explained the actual point of all this, sorry. 😁

These scripts create test deployments for every in-team `travis-web` PR. So you can visit a live version of the application, backed by the true API. They add links to the “checks” at the bottom of the PR. You asked me on Slack about why your org-staging API changes weren’t getting PR deployments, which is because I only set up PR deployments for org-production and com-production. So this PR adds an org-staging deployment:

image

The PR deployments live at a host of this form: `branch-name.test-deployments.travis-ci.org` (and `.com`). For org-staging, I decided to put them at `branch-name-staging.test-deployments.travis-ci.org`, because they could then share the same Herkou application, Redis database, and SSL certificate. I didn’t have to change the Heroku application at all, which made this PR trivial to add org-staging support.

The bigger question I’m wondering about is whether this slows PR builds down too much. My superficial examination suggests it _doesn’t_, but that makes little sense to me as it has to rebuild the JS and CSS assets for every deployment. Ideally there would also be a com-staging deployment. So I need to look further into this.

BUT ultimately the point of this PR was so you could merge it into [your own](https://github.com/travis-ci/travis-web/pull/678) and have your own stable org-staging deployment without having to worry about it being trampled by anyone. You can still do that, and if it turns out we don’t want this in `master`, just revert it!

Let me know if this makes any sense haha… I also wrote this up [here](https://github.com/travis-pro/team-teal/wiki/travis-web-pull-request-deployments).

backspace:
This is [superseded](https://github.com/travis-ci/travis-web/pull/700).

PR #689, opened 2016-09-16
make subscription status a proper notice

lislis:
looks like this:
screen shot 2016-09-16 at 16 44 43
screen shot 2016-09-16 at 16 44 49
screen shot 2016-09-16 at 16 44 55

### DIFF
--- a/app/templates/account.hbs
+++ b/app/templates/account.hbs
@@ -3,19 +3,17 @@
 <header>
   {{#unless config.enterprise}}
     {{#if config.billingEndpoint}}
-      <div class="cta-btn">
-        {{#if subscribeButtonInfo.subscribed}}
-          <a class="button" href={{billingUrl}}>Subscription active!</a>
+      {{#if subscribeButtonInfo.subscribed}}
+        <p class="notice-banner--green">This account has an active subscription. <a  href={{billingUrl}} title="Go to Travis CI Billing" class="right">Manage your subscription!</a></p>
+      {{else}}
+        {{#if subscribeButtonInfo.education}}
+          <p class="notice-banner--green">This account's subscription is flagged as educational. <a  href={{billingUrl}} title="Go to Travis CI Billing" class="right">Take a  look!</a></p>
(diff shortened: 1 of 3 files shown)


app/templates/account.hbs
```diff
@@ -3,19 +3,17 @@
 <header>
   {{#unless config.enterprise}}
     {{#if config.billingEndpoint}}
-      <div class="cta-btn">
-        {{#if subscribeButtonInfo.subscribed}}
-          <a class="button" href={{billingUrl}}>Subscription active!</a>
+      {{#if subscribeButtonInfo.subscribed}}
+        <p class="notice-banner--green">This account has an active subscription. <a  href={{billingUrl}} title="Go to Travis CI Billing" class="right">Manage your subscription!</a></p>
```
clekstro: It might be nice to extract a component here. That would allow us to get rid of some of the repeated markup.

Something like:

```hbs
{{subscription-status-banner 
  color="green" 
  messageText="This account has an active subscription"
  billingLinkText="Manage your subscription!"
}}
```

app/templates/account.hbs
```diff
@@ -3,19 +3,17 @@
 <header>
   {{#unless config.enterprise}}
```
clekstro: Unrelated, but asking b/c I noticed it in this diff. Can we now remove the `{{unless config.enterprise}}` here?  It doesn't look like `config.billingEndpoint` is configured for enterprise (see [here](https://github.com/travis-ci/travis-web/blob/master/config/environment.js#L66)) , so we could remove it altogether.

joshk: I'm not sure, but we can try removing this and testing it out, but we would need to talk to @bnferguson about this too. Might be best to leave it for now.

joshk:
The design improvements look fantastic!

backspace:
🎉 merging!

joshk:
Thank you @clekstro and @backspace and @lislis for your amazing work on this

PR #694, opened 2016-09-21
Revert V3 API usage for debug builds due to possible regression

backspace:
This reverts commit 42a6b3187179507ae0d5ca39a77d5123739d8a8d.

Someone wrote in that some people at their organisation can’t restart jobs. This is a possible temporary fix.


backspace:
This fixed the problem! I’ll write this up elsewhere.

PR #696, opened 2016-09-26
Update settings to use atomic API

backspace:
Settings used to be a bulk endpoint that could change all settings at once; it’s now one endpoint per setting.

This is a preliminary implementation, I’m awaiting confirmation that I’ve got the right understanding of what changed.


backspace:
I’ve merged in the `bd-staging-pr-deployments` branch so this can be tested in org-staging with the unmerged API changes. (Tomorrow?)

I’m not sure whether org-staging deployments are desirable for every PR so I may revert that before merging this.

backspace:
Closing in favour of #707.

PR #697, opened 2016-09-27
Change restarts to use V3 API

backspace:
This is a combination of:
* 42a6b31 to change to using the V3 endpoints for job and build restarting
* and #687 to get an org-staging deployment

It also adds a com-staging deployment.


backspace:
I’m closing this but will be using parts and variations of it in other PRs.

PR #700, opened 2016-09-28
Optional staging PR deployments

backspace:
This is an idea for optionally creating org- and com-staging PR deployments if the branch name contains `staging`. Staging deployments aren’t useful most of the time so the extra time for the PR build to pass isn’t worth it, but when testing in concert with API changes, it can be helpful.

I’ve pulled this from various branches. Last time I tried it, the com-staging deployment actually used com-production, strangely. Hopefully I can figure out why.

### DIFF
--- a/config/deploy.js
+++ b/config/deploy.js
@@ -1,4 +1,5 @@
 var VALID_DEPLOY_TARGETS = [
+  'org-staging-pull-request',
   'org-production-pull-request',
   'com-production-pull-request'
 ];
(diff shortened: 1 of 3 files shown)


backspace:
So com-staging wasn’t working because of [this line](https://github.com/travis-ci/travis-web/blob/8e71629edc899d704c69b83d5c5972fff938d982/waiter/config.ru#L35) in `config.ru`. The Ruby server sets the API endpoint regardless of the `API_ENDPOINT` environment variable if `TRAVIS_PRO` is true. The com-staging deployment now uses the proper API!

I’ll check in about this tomorrow.

backspace:
I talked to Curtis and we agreed to merge this. My prime concern is “discoverability”, as most people won’t know this magical quality of having `staging` in your branch name will give you these two extra deployments. We’ll see how much of a problem this is.

PR #701, opened 2016-09-29
Upgrade to Ember 2.8.1

clekstro:
This also upgrades Ember Data to 2.8.0 and removes all relevant deprecation warnings.  I'm no longer seeing the issue that we saw when upgrading to 2.7.1 🎉

### DIFF
--- a/app/components/add-env-var.js
+++ b/app/components/add-env-var.js
@@ -7,6 +7,7 @@ export default Ember.Component.extend({
   classNames: ['form--envvar'],
   classNameBindings: ['nameIsBlank:form-error'],
   store: service(),
+  raven: service(),
   flashes: service(),
 
   isValid() {
@@ -40,10 +41,10 @@ export default Ember.Component.extend({
         this.reset();
       } catch (e) {
         // eslint-disable-next-line
-        console.log('Error saving environment variable!', e);
         this.get('flashes').error('There was an error saving this environment variable.');
-
-        throw e;
+        if (!this.get('raven').ignoreError(e)) {
+          this.get('raven').captureException(e);
+        }
       }
     }
   }).drop(),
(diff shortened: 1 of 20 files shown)


backspace:
Hey, @clekstro, I pushed some commits to this, hopefully that‘s okay with you!

9066c2e fixes all four failures; the failure in the environment variable acceptance test was cascading and causing failures in the other three. The problem was that the Mirage API wasn’t returning an ID as part of the save response, which Ember Data wasn’t okay with.

I then included b0afaf5 to at least log a save failure to the console, which would have probably helped us diagnose this more quickly. However, presumably a failure could even result in a flash message since otherwise it’ll just be a silent failure. What are your thoughts?

clekstro:
@backspace Of course :) In terms of catching an error, I really like this within our tests; it definitely would make tracking down the failure much easier. One question: do you think it'd impact error reporting in production? i.e. we'd be logging the exception to the console but not sentry?

backspace:
It does seem worth adding Sentry logging _and_ a flash message. Do you want to do this, or shall I?

clekstro:
@backspace Go for it!

backspace:
I’ve added another commit to show the flash message upon failure, but I’m actually confused about how to work with Sentry, I guess until now I’ve been totally ignorant of it. Does it just log all uncaught exceptions? Is, therefore, the idea to throw the exception after having caught it?

clekstro:
@backspace Yep. Sentry will automatically be sent the exception if we fail to catch it, so it has to be thrown manually.

backspace:
Working on a holiday! 😱

I’ve pushed another commit to throw the exception. If/when the tests pass, I’ll merge! 🎉

backspace:
hmm welllll when I throw the error, it [produces test failures](https://travis-ci.org/travis-ci/travis-web/jobs/164636090), presumably due to the uncaught exception… any ideas? I don’t understand… I tried passing it manually to the `raven` service but that didn’t seem to help. 😢

clekstro:
I've fixed that last pesky error. Going to review in a PR env and merge/deploy if things look good.

backspace:
Thanks for fixing that! If this kind of special handling is necessary elsewhere, hopefully we can add it to the service to avoid having to repeat it.

PR #703, opened 2016-09-29
Remove API-fetched deployment variables

backspace:
Thanks to @BanzaiMan’s travis-ci/travis-scheduler#38 and
travis-ci/travis-build#848, these are now present as
environment variables rather than needing to be fetched
via the GitHub API.


backspace:
I am perhaps more excited about this than it warrants! It’s just a pleasure to not have to consult the GitHub API for these useful pieces of information. Merging!

PR #708, opened 2016-10-03
Restore PR staging deployments

backspace:
I guess in my confusion of several open branches at once and cherry-picking, I missed the actual script that would create the deployments. So the _links_ were there but not the deployments! I’m confused but… here we are.

I don’t like the repetition of the conditional and probably will eventually DRY this up further, but not at the moment!


backspace:
Also I lost [this `tr`](https://github.com/travis-ci/travis-web/blob/1977bfe5becca78893197d3206d068e07567957b/config/deployment/deploy-pull-request.sh#L2) which will convert periods in branch names to `-`s, which is necessary to not have an infinite subdomain crisis 😆

backspace:
I’ve added a final (hopefully?) commit that inlines the sending of GitHub Status API updates, which means the deployment link shows up immediately after a deployment is created, rather than all as a group after all deployments are created. This will let us get access to deployments earlier, especially when staging deployments are happening, which increases the time for the job to complete.

backspace:
FINALLY 😁

PR #712, opened 2016-10-05
Add environment variable logging

backspace:
This is to help debug a problem with the storage of environment variables.


backspace:
This problem is fixed!

PR #713, opened 2016-10-05
refactor subscription status to component

lislis:


app/templates/components/subscription-status-banner.hbs
```diff
@@ -0,0 +1,3 @@
+{{message}} <a  href={{billingUrl}} title="Go to Travis CI Billing" class="right">{{billingLinkText}}</a>
+
+{{test}}
```
backspace: Is this maybe a leftover?

lislis: It is, ha 💃

PR #715, opened 2016-10-05
Add Ember beta and canary deployments

backspace:
This is in progress.


backspace:
Now that the site is almost working on the current canary version of Ember, I’m going to merge this. It hinges on `master`-only deployments, so I’m eager to see if it actually works!

PR #716, opened 2016-10-06
Optimize team photos by switching to .jpg

clekstro:
This uses jpeg.io to optimize our team photos. Great savings even compared to previously optimized `.png`s.


clekstro:
This appears to save us about 1.8M on disk:
screenshot 2016-10-06 13 54 19

lislis:
Wheee
Robert Redford says yes

PR #718, opened 2016-10-06
fix permission part of displaying the repo acitvate button

lislis:
Looks like this is a bit we missed when we refactored permissions into a service.

I added a notice for users that don't have admin rights on a repo. Looks like this
screen shot 2016-10-06 at 17 09 19

Nice to haves:
- [x] update/write unit/integration tests
- [x] the button should be `$oxid-blue` since we try to establish that as interaction element color


lislis:
Removing `blocked`-label as this should work from our side. 
Everything API related would result in the same situation anyway.

PR #720, opened 2016-10-07
Add guard against empty config

backspace:


backspace:
This would be better if I could add error reporting when a config is empty, but I can’t figure out how since the helper is only passed the config. Since we are unable to reproduce the error, I’m fine with leaving it like this.

PR #721, opened 2016-10-07
Improve static title

backspace:
This fixes the flash of `{{title}}` as the document title. Everything is already in place for dynamic titles otherwise.

### DIFF
--- a/app/index.html
+++ b/app/index.html
@@ -3,7 +3,7 @@
   <head>
     <meta charset="utf-8">
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
-    <title>{{title}}</title>
+    <title>Travis CI</title>
     <meta name="description" content="">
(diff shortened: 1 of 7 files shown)


clekstro:
I think I remember chatting with @drogus about this at some point previously. Do we see this behavior in production as well, or is it limited to development?

backspace:
The `<title>` element is [replaced by the `waiter` server](https://github.com/travis-ci/travis-web/blob/77fc8bf2c85ebb1c9f80549ec5a8d5eee0ed60e4/waiter/lib/travis/web/app.rb#L167-L169). 🤔

backspace:
I’ve included assertions against the document title in various tests, I think that generally covers it. I wonder if there’s some way we could _require_ an assertion on titles, but I’m not sure that’s desirable…

Since this is overridden by the current production server, I think this is harmless to merge.

PR #723, opened 2016-10-11
Remove deprecated baseURL setting

backspace:
This was removed from the blueprint in Ember CLI 2.7:
https://github.com/ember-cli/ember-cli/blob/ad5af6877c44653eaf121e6820f06d280007c93c/blueprints/app/files/config/environment.js#L34-L36

See the blog post:
http://emberjs.com/blog/2016/04/28/baseURL.html

This will remove this bit of log noise:
https://travis-ci.org/travis-ci/travis-web/jobs/166748870#L613

### DIFF
--- a/app/adapters/ssh-key.js
+++ b/app/adapters/ssh-key.js
@@ -1,7 +1,7 @@
 import ApplicationAdapter from 'travis/adapters/application';
 
 export default ApplicationAdapter.extend({
-  namespace: 'settings',
+  namespace: '/settings',
 
   findRecord(store, type, id) {
     const url = `${this.urlPrefix()}/ssh_key/${id}`;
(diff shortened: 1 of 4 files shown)


backspace:
As I suspected, 978560f fixes the tests but makes the API have a `//` when getting/setting SSH keys, which works but is unsightly. I’d like to figure out how to have a leading `/` for Mirage requests in the test environment without having the `//` in production.

clekstro:
@backspace Thanks for looking into this. I left it in because of the test failures, but it'd be good to be rid of that warning 👍

backspace:
👍🏻 @clekstro. I’m not too fond of my current solution, but I can’t think of anything better. I don’t know why the SSH key adapter doesn’t include a leading `/` when the others do! This `urlPrefix` override includes one when no host is present; it’s hackish, but it works. This would probably be fixed by V3ifying the SSH keys code.

backspace:
Despite my disdain for the adapter change for something specific to test failures, I’m going to merge this.

PR #725, opened 2016-10-11
Add yarn

backspace:
Let us see!


backspace:
It appears to have _increased_ the time to install dependencies, and `node-sass` is in crisis or something. I just wanted to try this out; probably these are things that can/will be fixed but I’m not in too much of a rush.

PR #726, opened 2016-10-12
Remove compare link in build header for api and cron eventTypes

lislis:
There is no new commit API and Cron builds can be compared to and our GH-Url helper often generates falsy urls because of that

### DIFF
--- a/app/components/build-header.js
+++ b/app/components/build-header.js
@@ -27,6 +27,15 @@ export default Ember.Component.extend({
     }
   }),
 
+  displayCompare: Ember.computed('item.eventType', function () {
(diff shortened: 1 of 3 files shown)


app/components/build-header.js
```diff
@@ -27,6 +27,15 @@ export default Ember.Component.extend({
     }
   }),
 
+  displayCompare: Ember.computed('item', function () {
```
backspace: The dependent key should be `item.eventType` I believe?

lislis:
@backspace I added a basic integration test for the component to test. wdyt?

backspace:
Yay, excellent, looks good to me!

PR #741, opened 2016-10-25
Move log code inside Ember application

backspace:
This is just a first pass. I’m creating this early because when I run locally the CSS is missing, so I want to see if this happens on a PR deployment.


backspace:
The CSS problem was [unrelated](https://github.com/travis-ci/travis-web/pull/749). I’ve tested this in the staging environments and all appears well.

PR #742, opened 2016-10-25
Add auto-cancellation switches when API response includes them

backspace:
If the `/repo/:repo/settings` response includes `auto_cancel_pr_builds` or `auto_cancel_branch_builds`, this will show switches for them. Labels TBD.


backspace:
It’s happening! 🎉

PR #746, opened 2016-10-28
Rename auto-cancel settings

backspace:
This future-proofs against a possible variant that will cancel
running jobs rather than waiting ones.


backspace:
The setting names are going to stay for now. This is a trivial change that can be easily recreated.

PR #748, opened 2016-10-28
Fix typo in component test

gusaiani:
Hi all again,

found and fixed another `langauges` → `languages`.

Did a project find and yup, that was the last one.

Thanks always


backspace:
hah thanks again. 💞

PR #752, opened 2016-10-31
Change ansiparse to be imported via Browserify

backspace:
The presence of ansiparse in the vendor directory predates
the possibility of importing it in a more NPM-y fashion
by a while.

I needed to fork the ansiparse repository to remove a
conditional around the module export; I’ll propose a
pull request and see what happens.


backspace:
I’m distressed by the lack of PR deployments which disrupts my usual workflow, but I deployed to org-staging and all appears to be in order.

PR #758, opened 2016-11-01
Improve offline testing experience

backspace:
Today, while riding a train with no internet access, I noticed some room for improvement in being able to still work on `travis-web` offline.

This currently includes some hackery around a Gravatar test that I think is Doing Too Much, will fix that.

This also deletes `app/utils/test-auth.js`, which appears to be unused. 🎉


app/utils/urls.js
```diff
@@ -69,7 +69,11 @@ email = function (email) {
 };
 
 gravatarImage = function (email, size) {
-  return 'https://www.gravatar.com/avatar/' + (md5(email)) + '?s=' + size + '&d=blank';
+  if (config.gravatarUrl) {
+    return `${config.gravatarUrl}?hash=${md5(email)}&s=${size}&d=blank`;
+  } else {
+    return 'https://www.gravatar.com/avatar/' + (md5(email)) + '?s=' + size + '&d=blank';
+  }
```
clekstro: Should the url params be identical here?

backspace:
I’m going to close this in favour of #762. I think it would be nice to remove the 96 (I think?) Gravatar 404s that happen in tests, but I can’t focus on it right now. My eventual desire is to remove all spurious console errors in the tests.

PR #766, opened 2016-11-03
MVD with starring

lislis:
this is `ce-lp-dashboard-behind-feature-flag` rebased onto master with the starring feature


app/controllers/dashboard/repositories.js
```diff
@@ -99,6 +102,20 @@ export default Ember.Controller.extend({
...
+            .success(`You successfully starred ${repo.get('slug')}`);
+        });
+    },
+    unstar(repo) {
+      this.get('ajax').ajax(`/v3/repo/${repo.get('id')}/unstar`, 'POST')
+        .then(() => {
+          this.get('flashes')
+            .success(`You successfully unstarred ${repo.get('slug')}`);
+        });
     }
```
lislis: I moved the starring action from the component to the controller. But I'm not sure if there is a better way of doing that with some api provided by Ember Data? 
Right now it doesn't trigger a refresh of the model, which makes sense - the model doesn't know about the POST request that way...
And everything else depends on that, but I guess there is a way to clean up the rest of the controller anyway

backspace: This is an unfortunate limitation with stock Ember Data, which is that “non-CRUD” actions aren’t really supported. Ember Data would like `starred` to be updated via the usual `PUT` (or `PATCH`?) to the endpoint (via `set`-ing the attribute and `save`ing), but the API wants a `POST` to `star` and `unstar`, which doesn’t match Ember Data conventions. (Hopefully I’m not just restating things you already know!)

I know of [`ember-data-actions`](https://github.com/davewasmer/ember-data-actions) for adding support for non-CRUD actions, or this [blog post showing a similar manual implementation](https://karolgalanciak.com/blog/2016/02/07/implementing-non-restful-actions-with-ember-data/). These may be worth looking more into to clean this up a little, but I’ve never used them myself…???

As a quick hack, you could use `repo.set('starred', true)` and `false` when starring and unstarring, which would cause the client to reflect the changes on the server. This would probably be best done within the `then` to guarantee that the change was successful. It makes me feel a bit 😳 but it really amounts to the same thing so I can push the 😳 away haha

Yet another option would be to call `repo.reload()` within the `then`, which would trigger a refresh from the API! This seems wasteful, having an entire API call refresh the model, but it’s more “clean” because it’s not manually mirroring the server- and client-side changes.

I like the first option the most because it would give us a nice `repo.star()` and `repo.unstar()` interface, which is the most Ember Data-y. It’s concise enough that I’d feel fine just having it within the component instead of the action needing to bubble up to the controller. However, this solution is also the most involved!

Well hopefully the above isn’t Too Much. Let me know if you want to discuss further!

joshk:
I took a sneak peak, and I love this!

Regarding the flash message, I think this can be removed. It feels clunky that we display flash messages during actions which could instead provide feedback that the action was successful. 

For example, if you click on the star, and the star is added visually, and there is no error, then I don't think we need to add a flash message. If there is an error, then displaying a message about the error would be useful.

Also, I wonder, should we keep the starred repo in the `Active Repositories` list after you click it so it doesn't seem like we just take it away straight away? Interaction wise, this would mean that a user would click a star to star a repo, that repo would be listed at the top in the `Starred Repositories` section, while also staying i the `Active Repositories` section, this way the repository doesn't disappear from in front of you, but if you scroll up it would be in the `Starred Repositories` section.

Does that all kinda make sense?

lislis:
Hey @josh! I agree on the flash messages. Once we get the proper star-refresh-workflow going I'll remove the success flash message.

I'm not sure about the a (newly) starred repo staying in active repos. Would we keep on listing it with a yellow star among the others? I think that could lead to confusion about what starring actually does besides turning the star yellow (which is listing at the top).

What I would like to have (but not sure how/if we can do that) is that:
- you click on the star
- for until the request promise returns show the loading indicator
- the promise returns, we fade out the repo, showing a ininline message 'you starred this repo, you can find it at the top' (or similar)
- fade out the inline message

wdyt? I think that would make a nice UX flow

joshk:
@lislis I really love what you have created!

I can agree in part that it seems a little counter productive to keep starred repos in the active list, although on the other hand I think the UX when starring a repo, and seeing what repos are in the starred list when scrolling through a long list is much better. All in all, this is a great cup cake.

I wonder if we might want to remove the Org filter and user sync for now too. The spacing feels a little off, and some of the UX around the org filter feels like it needs some more tweaking. We could remove those for now for the cupcake and aim for this release in Dec? (beta features permitting)

lislis:
@joshk I can remove the filter and user sync for now. I'll create a new issue for improvement suggestions

joshk:
This is awesome!

joshk:
@lislis regarding the switches on the feature page, can we turn off the 'loading'/'progress' animation/indicator?

backspace:
I added a commit to disable the dot reporter so we can hopefully get more information on what’s failing. This should be removed before merging 😳

backspace:
comic

lol

I neglected my support duties for a while and managed to get the build passing again, so you can try this out at the [PR deployment](https://lp-readd-stars.test-deployments.travis-ci.org/).

It doesn’t make much sense because the canary build was already passing, but I finally got it working by downloading PhantomJS 2.1.1, which is what we run on our own machines, rather than the 2.0.0 that is installed on the Trusty image by default.

The branch currently includes various experimental commits that should be removed before this is merged. Also, the 2.1.1 hackery may need some attention, but it’s beyond EOD for me now!

lislis:
closing in favour of https://github.com/travis-ci/travis-web/pull/937

PR #767, opened 2016-11-03
Don't display title if there is no build number

stewx:
There is a bug in the build tiles that makes the tooltip display "Build #undefined undefined". 
screen shot 2016-11-03 at 3 50 37 pm


backspace:
I’d normally like a test to accompany such a change, but I see we don’t have tests for this component. Thank you, merging!

backspace:
This is live in production now! 💞

PR #768, opened 2016-11-04
Using ember-inline-svg

lislis:
I've been wanting to change the way we deal with icons for quite some time. (moving from base 64 encoded in the stylessheet to inlining svg in the html)
This PR is the start


lislis:
I'd like to get this merged in first and then continue with the work

PR #771, opened 2016-11-07
Disable Sentry for beta/canary deployments

backspace:
Sentry swallows all errors, which makes debugging more difficult.


backspace:
meh apparently my workflow of continuing on a branch that’s already had a PR merged on it doesn’t work with squash-and-merge :cry:

PR #774, opened 2016-11-08
fix auto-run

stefanpenner:
screen shot 2016-11-07 at 5 03 13 pm

---

Note, didn't run tests. Just noticed via GH, and issues don't appear to be enabled. So this is my pseudo issue, they may fix the issue.


backspace:
Hey, thanks for this! Am I correct in believing this will help with performance?

It’s true that issues are disabled on this repository. Public bug reports can go [here](https://github.com/travis-ci/travis-ci/issues), but PR-as-bug-report-and-fix is excellent!

PR #775, opened 2016-11-08
Fix Pusher errors by pre-authorizing private job channels

clekstro:
Typically, the pusher.js client authorizes requests by hitting a user-defined endpoint,
which tells pusher whether to allow a user to connect. In our case, we've exposed an auth endpoint
at `/auth/push`. When the app boots, we make a single request with all the channels we receive from our `/auth` endpoint, authenticating them so that the client can connect to Pusher successfully.

Unfortunately, when we attempted to subscribe a to a private job channel, we would often (but not always) receive an auth error. 

This change hits our /auth/pusher endpoint with a single job channel to
authorize it before making the subscription call to Pusher.  Now that the subscribe call is
guaranteed to happen after the auth call has succeeded, we shouldn't see this error anymore.

I'm not super pleased about the implementation, but wasn't sure of a
better way to get at the socketId (normally simply yielded as an
argument to the `bulk_ajax` transport that Pusher calls internally) than
by setting it on the global `Travis.pusher`.

### DIFF
--- a/app/models/job.js
+++ b/app/models/job.js
@@ -1,4 +1,5 @@
-/* global moment, Travis */
+/* global moment, Travis, Pusher */
+
 import { compact } from 'travis/utils/helpers';
 import Ember from 'ember';
 import Model from 'ember-data/model';
@@ -140,8 +141,14 @@ export default Model.extend(DurationCalculations, {
       return;
     }
     this.set('subscribed', true);
-    if (Travis.pusher) {
-      return Travis.pusher.subscribe('job-' + (this.get('id')));
+
+    if (Travis.pusher && Travis.pusher.ajaxService) {
+      return Travis.pusher.ajaxService.post(Pusher.channel_auth_endpoint, {
+        socket_id: Travis.pusher.pusherSocketId,
+        channels: ['private-job' + this.get('id')]
+      }).then(() => {
+        return Travis.pusher.subscribe('job-' + (this.get('id')));
+      });
     }
   },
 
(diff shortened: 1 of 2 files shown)


app/models/job.js
```diff
@@ -140,8 +141,14 @@ export default Model.extend(DurationCalculations, {
       return;
     }
     this.set('subscribed', true);
-    if (Travis.pusher) {
-      return Travis.pusher.subscribe('job-' + (this.get('id')));
+
+    if (Travis.pusher && Travis.pusher.ajaxService) {
+      return Travis.pusher.ajaxService.post(Pusher.channel_auth_endpoint, {
+        socket_id: Travis.pusher.pusherSocketId,
+        channels: ['private-job' + this.get('id')]
```
backspace: Should this have a second dash?

app/models/job.js
```diff
@@ -140,8 +141,14 @@ export default Model.extend(DurationCalculations, {
...
+      }).then(() => {
+        return Travis.pusher.subscribe('job-' + (this.get('id')));
+      });
```
backspace: It would be nice if we had a Pusher service to encapsulate interactions with Pusher, as then this could be `this.get('pusher').subscribeToJob(this.get('id'))` or something like that, and all this Pusher-specific code wouldn’t need to pollute the job model. I guess `app/utils/pusher.js` is service-esque, what do you think of moving this code into there?

clekstro: Totally agree that this should be in its own service.  That said, I started going down the road of extracting this and started to run into issues and figured it would be better to get it out in its current state.

PR #778, opened 2016-11-08
Add job banner notice for jobs on `builds.ec2`

meatballhat:


lislis:
Updated styles. Looks like this:
screen shot 2016-11-09 at 12 22 03

PR #692, opened 2016-09-19
Continuous visual integration with Percy

fotinakis:
(chatted with @joshk and @backspace at GitHub Universe about this — basically, here's a fully-integrated visual testing solution for travis-web if you want it! :) )

After this PR, you would see a new commit status on future PRs:

screen shot 2016-09-12 at 10 13 10 pm
screen shot 2016-09-12 at 10 13 49 pm

^ notice that your tests pass, but there were visual changes that Percy caught which need review. Just look through the screenshots in Percy, click Approve if the changes are ok and it'll be green!

Here's what this looks like in Percy:

screen shot 2016-09-19 at 11 25 50 am
screen shot 2016-09-19 at 11 26 26 am

And here's what it looks like if we change `$turf-green: #39aa56` into `$turf-green: blue`:

screen shot 2016-09-19 at 11 26 58 am

Including responsive visual diffs!

screen shot 2016-09-19 at 11 28 46 am

Because I'm a nerd, here's a demo video: http://www.youtube.com/watch?v=MrikhPRczJQ

**Impact on test speed:** zero! Percy is totally architected around "DOM snapshots" and handles rendering entirely in our service outside of your test suite. This allows to do things like [freeze animations](https://blog.percy.io/freezing-animations-in-visual-regression-tests-e6db56a7b3a5) so they don't create noisy diffs, and heavily parallelize the rendering underneath (if you're interested, [here's a talk I gave at RailsConf](https://www.youtube.com/watch?v=5h-JJ2wqiIw) this year about why these are hard problems).

**Setup work needed:**
- You'd need to do the "Setup" steps for Percy — 1) login to Percy, 2) enable the **travis-ci/travis-web** repo, 3) set the `PERCY_TOKEN` and `PERCY_PROJECT=travis-ci/travis-web` env vars in your Travis repo settings. See https://percy.io/docs/setup/travis-ci for more info. 
- You'd need to sign up! :) I did some quick calculations (~100 travis-web builds over the last 15 days, 90 snapshots per build = ~18,000 snapshots/month) so I think our Medium ($149) or Large ($299) [plans](https://percy.io/pricing) would cover you for a long time.

**Future work:**
- Tuning your visual coverage. Removing snapshots where you might not care, adding more of them where you do. Adding snapshots when you catch a visual regression.
- Supporting builds from forks (chatted with @joshk about this, one day we'd love a way to re-run Travis builds with secure env vars present, or some other workaround for the fork security problem).
- Deeper travis integration! :)
[...]

PR #780, opened 2016-11-09
Continuous visual integration with Percy

backspace:
This is a continuation of #692. I’ve merged the latest `master` and will further tune as needed.


backspace:
Hey, @lislis and @clekstro, if you like, want to look over the [snapshots I kept](https://percy.io/travis-ci/travis-web/builds/128209), do they seem like a good overview of screens we’d like visual diffs for? We can always refine later, but I think this is a representative sample to start with.

(You’ll need to authorise through GitHub but I think the link should work after that.)

clekstro:
@backspace Seems like a great start!

fotinakis:
Awesome! Looking forward to your feedback, you can always use the Intercom button in Percy to chat if you have any questions.

PR #782, opened 2016-11-09
Remove Sauce Connect references

backspace:
This brings us closer to the Ember CLI blueprint for `testem.js`
and removes unused configuration and scripts.


clekstro:
Thanks for cleaning this up @backspace!

PR #783, opened 2016-11-09
Add step to save Redis URLs from Heroku

backspace:
The Heroku Redis URLs for the PR/beta/canary deployments change
periodically, which breaks things until I manually update the Redis URL
environment variables for the Travis environment. This obtains them
at build time from Heroku instead.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,10 +23,7 @@ cache:
     - bower_components
 
 before_install:
-  - mkdir travis-phantomjs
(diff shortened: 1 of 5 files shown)


backspace:
This also moves `travis-web` jobs onto the Trusty beta, because it includes the `heroku` command needed to fetch the Redis URLs. This has the added benefit of including PhantomJS 2.0.0, which comes pre-installed in that environment, so we no longer need the `.travis.yml` lines related to installing that.

PR #793, opened 2016-11-15
Add link to branch in build header

lislis:
screen shot 2016-11-15 at 17 16 47

### DIFF
--- a/app/components/build-header.js
+++ b/app/components/build-header.js
@@ -47,5 +47,13 @@ export default Ember.Component.extend({
 
   elapsedTime: Ember.computed('item.startedAt', 'item.finishedAt', 'item.duration', function () {
     return durationFrom(this.get('item.startedAt'), this.get('item.finishedAt'));
+  }),
+
+  displayBranch: Ember.computed('item', () => {
+    if (!Ember.isEmpty(this.get('item.build'))) {
+      Ember.computed.equal('item.build.branch.name', 'item.repo.defaultBranch.name');
+    } else {
+      Ember.computed.equal('item.branch.name', 'item.repo.defaultBranch.name');
+    }
   })
 });
--- a/app/components/build-header.js
+++ b/app/components/build-header.js
@@ -53,6 +53,6 @@ export default Ember.Component.extend({
 
   @computed('item.repo.slug', 'commit.branch')
   urlGitHubBranch(slug, branchName) {
-    return githubBranch(slug, branchName);
+    return this.get('externalLinks').githubBranch(slug, branchName);
   }
 });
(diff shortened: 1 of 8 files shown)


joshk:
👍 from me!

app/components/build-header.js
```diff
@@ -42,5 +42,9 @@ export default Ember.Component.extend({
 
   elapsedTime: Ember.computed('item.startedAt', 'item.finishedAt', 'item.duration', function () {
     return durationFrom(this.get('item.startedAt'), this.get('item.finishedAt'));
+  }),
+
+  urlGitHubBranch: Ember.computed('item.repo.slug', 'item.branch.name', function () {
+    return githubBranch(this.get('item.repo.slug'), this.get('item.branch.name'));
   })
```
backspace: This is perfect, but I thought you might be interested to see a peek of the future, available as the [`ember-computed-decorators`](https://github.com/rwjblue/ember-computed-decorators) addon, which would change this with the proposed ES2017 (maybe?) decorators feature. If you haven’t already heard:

```
@computed('item.repo.slug', 'item.branch.name')
urlGitHubBranch(slug, branchName) {
  return gitHubBranch(slug, branchName);
}
```

I think it’ll be nice to not have that duplication of specifying the dependent keys in the computed property header and then having to fetch them within the function.

app/templates/components/build-header.hbs
```diff
@@ -37,6 +37,12 @@
           {{/if}}
         </li>
       {{/if}}
+      <li>
+        <a class="commit-branch-url" href="{{urlGitHubBranch}}" title="Look at the branch on GitHub">
+          {{inline-svg 'images/stroke-icons/icon-github' class="icon"}}
+          <span class="label-align">Branch {{item.branch.name}}</span>
+        </a>
+      </li>
```
backspace: I seem to have found a bug:

image

While looking at this via PR deployment, it was working. Then I clicked the branch name to view it on GitHub, pressed back, and the branch name was missing. 😢

The link ends with `undefined` too, sadly.

(This isn’t really a comment on the template code, I just don’t think I can make a general comment without submitting the entire review.)

app/components/build-header.js
clekstro: Just wanted to mention I'm a huge fan of that addon and I'd love to see us move in that direction. Decorators are currently stage 2, so it'll still probably be awhile until they're implemented natively, but I think the benefits are high enough for us to justify the investment.

app/templates/components/build-header.hbs
lislis: hmm weird. I can't reproduce this neither locally nor on the PR deploy 😕

app/components/build-header.js
lislis: I haven't heard about this yet! thanks for sharing

app/templates/components/build-header.hbs
lislis: After some more back and forth clicking I was able to reproduce it once. I'll try and see if I can find something

PR #794, opened 2016-11-16
Declare with const and fix typo in utils test

gusaiani:
Hi @backspace,

doing my hour-a-day's worth of studying your app, saw more things like the ones we looked at in the past couple of days.

Took the liberty of removing one `assert.ok(!...` in favor of `assert.notOk(...` just like one in a recent update you guys made.

I was also wondering about a number of `let`s, for example:
```
let done = assert.async();
```
if that shouldn't be a `const`, or if it makes much of a difference at all :)

Anyway...will be glad to help out with those things if you think it's worthwhile.

Thanks always,
Gus

### DIFF
--- a/tests/unit/utils/favicon-manager-test.js
+++ b/tests/unit/utils/favicon-manager-test.js
@@ -35,15 +35,14 @@ test('set favicon if there is no link tag in head', function (assert) {
   }, 20);
 });
 
-test('replace exisiting link tag', function (assert) {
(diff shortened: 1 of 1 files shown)


backspace:
Thanks again!

For `let` vs `const`, which I personally always use `const` for one-time assignments, we don’t have an internal policy on that, so it doesn’t seem worth changing to me. But thanks for asking!

PR #795, opened 2016-11-16
Use chrome instead of phantomjs

clekstro:
This basically just merges the [config from ember-ci](https://github.com/mike-north/ember-ci/blob/master/blueprints/travis-chrome/files/.travis.yml) with our current setup.


clekstro:
This didn't really get any traction. Might revisit this in the future.

PR #799, opened 2016-11-18
Refactor towards routable controllers

kiwiupover:
By moving away from controllers that don't render template
then the app will in a place that can take advantages of ember engines.

Would you like me to add a test for the current-user service? There was not one for the current-user controller.

Cheers


clekstro:
Hi, thanks for opening this up!

I'll need to double-check, but it seems like we could use the `currentUser` that's already on the `auth` service, so I'm not sure the `currentUser` service would be necessary.

Could you offer some more details about the controller removal and how that's relevant to engines? We've been discussing their use internally, so it'd be great to find out more :)

kiwiupover:
I do think that the currentUser could be merged into the auth service. That is the right way to go. 

When it comes to engines there is a need to pass dependencies from the main app to the engine and those dependencies need to be services. With the travis-basic-route had a dependency on the current-user controller and controllers can't be passed as a dependency to the engine. Furthermore a service is a better abstraction for an object that holds long lived state. 

I have a pr here that turns the dashboard into an in-repo lazy loaded engine. https://github.com/kiwiupover/travis-web/pull/1 

This PR is a few of the changes that were needed to make the lazy-loaded engine work. 

I think the [profile](https://travis-ci.org/profile/kiwiupover) section of travis would be perfect for a lazy-loaded engine and make the initial js file size a bunch smaller for the user.

The only controllers I think that are needed these days in an ember app are the routable controllers they are the controllers that a route needs to render a route. If a controller doesn't have a template of the same name then it probably not a routable controller and needs to become an other object like a service.

clekstro:
Hey @kiwiupover, thanks for getting back to us!
> When it comes to engines there is a need to pass dependencies from the main app to the engine and those dependencies need to be services. With the travis-basic-route had a dependency on the current-user controller and controllers can't be passed as a dependency to the engine. Furthermore a service is a better abstraction for an object that holds long lived state.

I wasn't up to speed on how dependencies are passed to engines, so I appreciate the added details. We'd definitely like to move off of most of the controllers we currently have in the app and replace them with services, it's just a matter of priorities/bandwidth at this point.

> I have a pr here that turns the dashboard into an in-repo lazy loaded engine. kiwiupover#1

I had a look at your PR this weekend, and that's totally awesome! We'll probably want to wait to deploy this into production until the lazy loaded engines are considered stable, but there are a few other use cases besides `/profile` that we talked about internally last week, so we're really excited for this to become a reality.

kiwiupover:
@clekstro I updated this PR to consolidate the currentUser service into the auth service. 

Let me know if you want anything else changed.

clekstro:
Everything appears in order, thanks for contributing!

PR #800, opened 2016-11-18
Upgrade Ember-CLI, Ember and Ember-Data

clekstro:
Also use `ember-cli-sass` instead of `broccoli-sass` (styles were broken after the upgrade).

I should also note that I downgraded several packages to match the pinned versions in the Ember-CLI blueprint.

I still had issues with production builds failing because of directories being present in `dist/`:
screenshot 2016-11-18 09 08 46
However, the project appeared to build correctly and I was able to work around the error.

I'll do a PR -> Staging -> Production deploy assuming there aren't any unexpected issues.


backspace:
Excellent! 🎉

PR #802, opened 2016-11-18
Fix log error as reported by Sentry

clekstro:
This is the most common error we get in Sentry. Based on the stack
traces and source mapping, this is apparently where the issue originates.


clekstro:
This looks good on the PR env. Merging.

PR #809, opened 2016-11-21
Add path restriction for ember-inline-svg

backspace:
As suggested here:
https://github.com/minutebase/ember-inline-svg#atrociously-slow-build-times-

In my quite-informal testing, this brought a fresh build time
down from almost two minutes to less than 20 seconds.


clekstro:
Love how this shortens down all the image paths as well 💃

backspace:
Note to @lislis: if you end up `inline-svg`-ing any more SVGs, they may need to be added to the list of `paths` in `ember-cli-build.js`. But I think deliberately specifying paths to be included should keep things building more quickly. 😀

PR #811, opened 2016-11-21
Readme update to trigger a build/deploy

clekstro:


clekstro:
Just for clarification, only opened this to debug whether we were having deploy/environment issues yesterday or if the problem was code-related. I've isolated the issue in the code and will submit a new PR when that's ready to go.

PR #820, opened 2016-11-23
Turn off fingerprinting for team assets

backspace:
This should make them appear in PR deployments. But is it
a good idea? ¯\_(ツ)_/¯


backspace:
I was wrong 😆

PR #827, opened 2016-11-29
Wrap hook switch async behavior in task

clekstro:
We are seeing errors in production where `then` is undefined on `hook.toggle()`, which
is a product of the component having been destroyed before our promise can execute.

Wrapping this in a task should fix the error and make this behavior more robust.

### DIFF
--- a/app/components/hook-switch.js
+++ b/app/components/hook-switch.js
@@ -1,10 +1,15 @@
 import Ember from 'ember';
+import { task } from 'ember-concurrency';
 
 export default Ember.Component.extend({
   tagName: 'a',
   classNames: ['switch'],
   classNameBindings: ['hook.active:active', 'disabled:disabled', 'disabled:inline-block'],
   click() {
+    this.get('toggleHook').perform();
+  },
+
+  toggleHook: task(function* () {
     if (!this.get('disabled')) {
       this.sendAction('onToggle');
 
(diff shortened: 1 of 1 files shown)


app/components/hook-switch.js
```diff
@@ -13,12 +18,12 @@ export default Ember.Component.extend({
...
-      return hook.toggle().then(() => {
+      yield hook.toggle().then(() => {
         pusher.subscribe(`repo-${repoId}`);
       }, () => {
         this.toggleProperty('hook.active');
         return this.sendAction('onToggleError', hook);
       });
     }
-  }
+  }),
```
backspace: Excellent! Does there need to be anything [like `drop` or `restartable`](http://ember-concurrency.com/#/docs/task-concurrency) here?

clekstro: I don't think so. The task will fire whenever a user clicks to enable/disable any hook, and we want that behavior isolated. If we `drop` or `restart`, wouldn't that mean we'd potentially impact a previous toggle, e.g. if a user clicks two different hooks in quick succession?

backspace: I don’t think there would be any effect between hook switches, as each component is isolated. I was more thinking about what would happen if someone clicked a switch multiple times. But when I try clicking a switch when it’s already saving, it does nothing, so never mind! 😀

clekstro: Of course! I was thinking for some reason that this was happening in a parent component :facepalm:. Going to merge this in :)

PR #828, opened 2016-11-29
Add default warning category to broadcasts

backspace:


backspace:
HMM but I was apparently confused and a category-less broadcast _does_ show:

image

For some reason I thought it wouldn’t show at all. So I think this is unnecessary?

I do feel like the refactoring I performed in the second commit is worthy, so I’ll open another PR.

PR #830, opened 2016-11-29
Move link underline beneath text only

backspace:
The underline was extending to the whitespace between the
text and the icon, which I found quite irksome 😬

image

What do you think of this technique, @lislis? I am somewhat displeased by the `inner-underline` class, I originally had it as `label-align`, but it seemed weird to me to need to use that for the text elements within `build-status`, which don’t currently have any kind of wrapper.

uhhhh apart from the test failure, which I will address haha

### DIFF
--- a/app/styles/app/modules/build-header.sass
+++ b/app/styles/app/modules/build-header.sass
@@ -20,7 +20,8 @@
 
   a:hover,
   a:active
-    text-decoration: underline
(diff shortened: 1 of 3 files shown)


lislis:
I like it, especially since it keeps the icon clickable!

backspace:
excellentttttttt merging then 🎉

PR #837, opened 2016-11-29
Update OS X banners

MariadeAnton:
Update OS X banners to include:

- beta-xcode6.1 needs to show that it will be retired on Fri, Jan 20, 2017 at Noon PST
- The other Xcode 6.x images warn they are now routed to xcode6.4
- The Xcode 7.x images warn they are now routed to xcode7.3

### DIFF
--- a/app/components/job-infrastructure-notification.js
+++ b/app/components/job-infrastructure-notification.js
@@ -19,8 +19,13 @@ export default Ember.Component.extend({
 
   macOSImage: Ember.computed.alias('jobConfig.osx_image'),
 
+  isDeprecatedMacImageXcode6: Ember.computed('queue', 'macOSImage', function () {
+    const retiredImages = ['beta-xcode6.1'];
+    return this.get('isMacStadium6') && retiredImages.includes(this.get('macOSImage'));
+  }),
+
   isRetiredMacImageXcode6: Ember.computed('queue', 'macOSImage', function () {
-    const retiredImages = ['beta-xcode6.1', 'beta-xcode6.2', 'beta-xcode6.3'];
+    const retiredImages = ['beta-xcode6.2', 'beta-xcode6.3'];
     return this.get('isMacStadium6') && retiredImages.includes(this.get('macOSImage'));
   }),
 
--- a/app/templates/components/job-infrastructure-notification.hbs
+++ b/app/templates/components/job-infrastructure-notification.hbs
@@ -12,16 +12,20 @@
         Please read <a href="http://blog.travis-ci.com/2016-11-08-trusty-container-public-beta/?utm_source=trusty-container-based-beta-notice&utm_medium=banner&utm_campaign=trusty-container-based-beta" title="Trusty container-based beta">our blog post about the public beta</a>.
       {{/notice-banner}}
     {{/if}}
+    {{#if isDeprecatedMacImageXcode6}}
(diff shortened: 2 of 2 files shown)


backspace:
Hopefully this was slightly easier to update than before, after my refactoring. Merging!

backspace:
Hopefully this was slightly easier to update than before, after my refactoring. Merging!

backspace:
Hopefully this was slightly easier to update than before, after my refactoring. Merging!

solarce:
thank you for help in this @backspace. This is a really useful thing for us to be able to tweak and use, so I appreciate you making the overall code simpler to maintain and update as we go

app/components/job-infrastructure-notification.js
```diff
@@ -19,8 +19,13 @@ export default Ember.Component.extend({
 
   macOSImage: Ember.computed.alias('jobConfig.osx_image'),
 
+  isDeprecatedMacImageXcode6: Ember.computed('queue', 'macOSImage', function () {
+    const retiredImages = ['beta-xcode6.1'];
+    return this.get('isMacStadium6') && retiredImages.includes(this.get('macOSImage'));
```
MariadeAnton: But it's great to keep it in mind for future updates, thank you!

cotsog:
It seems the banner is shown even though a job didn't actually run on the Xcode 7.3 image e.g.

https://travis-ci.org/cotsog/travis_ci_prod_multi_os/jobs/178624218

This job actually ran on the `xcode7.1` image: https://travis-ci.org/cotsog/travis_ci_prod_multi_os/jobs/178624218#L4

Is it possible to add a date check i.e. if the job ran after November 28th then show the banner else don't show it?

@MariadeAnton, @backspace, @solarce: thoughts?

backspace:
hmm, good point, I didn’t think through the display implications before deprecation and after rerouting. It seems like a parallel structure of notifications is needed.

Such as, based on the date of the job:

> This job ran on an OS X image that was retired on Monday …

vs

> This job ran on our Xcode ZZZ image. Please update your .travis.yml accordingly.

It also seems unfortunate to me that the code needs to be updated on the date of retirement to change from “will be retired” to “was retired”. I can make a PR that attempts to incorporate all these variations.

MariadeAnton:
@backspace that would be great, thank you! Let me know if there's anything I can help with. In the meantime, we could workaround this issue with something similar to: 

> This Xcode image has been retired <a href="https://blog.travis-ci.com/2016-11-17-retiring-some-osx-images/">on Monday, November 28th</a>.

>This configuration now routes to our <strong><a href="https://docs.travis-ci.com/user/osx-ci-environment/#Xcode-6.4">Xcode 6.4 image</a></strong> instead.

backspace:
Thanks for the wording, @MariadeAnton. I might as well just solve this immediately rather than making an interim fix. Here’s my proposed sentence that could incorporate all possibilities:

> This job [ran|is running] on an OS X image that [will be retired|was retired] on [date]. [After that, t|T]his configuration [will route|now routes] to our Xcode [version] infrastructure instead.

I think this has reached a level of combinatoric explosion that it no longer makes sense to use a set of conditionals in the template to manage slightly different wording for the different Xcode versions and past/future-ness, so I’ll move to an in-code solution, but I am to make it still intelligible for updating.

PR #841, opened 2016-11-30
Add alternate determination of commit ID

backspace:
Gatekeeper includes the commit_id in the build:started event,
but not in the build:created one. This produces Ember Data
errors about the missing ID.


backspace:
I opened this for posterity, but I have instead fixed Gatekeeper! 🎉

PR #845, opened 2016-12-02
Add Pusher tests

backspace:
This is very much in progress. I first planned to convert Pusher to a service,
but I’ve decided to add tests first, because it’s too unwieldy to convert without
some kind of reassurance that I haven’t broken everything.


backspace:
This is an **exciting moment** where I have an acceptance test for `/` where Pusher events cause the foregrounded repository to change when a new build is created. It’s a total mess at the moment but refinement should be less laborious. After that I’ll probably put this aside for a bit because it was quite the struggle.

backspace:
This seems good for now! I don’t think all Pusher events are covered and I didn’t end up converting it into a service, but now we at least have some tests that will make that easier.

PR #847, opened 2016-12-02
Remove .buildpacks because we will no longer use deprecated heroku-bu…

aakritigupta:
…ildpack-multi

~The test deployments are expected to be broken~


backspace:
I don’t think this will affect the test deployments, as they don’t use that buildpack.

Thanks for doing this!

PR #848, opened 2016-12-02
Add crisp edges for SVG in testing

backspace:
This is Firefox-only, it appears, but Percy uses Firefox
for diffs, so that seems fine.


backspace:
This is an experiment, I won’t be able to see whether it worked until this is merged into `master` and I run a new diff.

PR #851, opened 2016-12-02
Add fonts to test environment to mirror main index.html.

fotinakis:
@backspace 

I believe this is why your fonts are not working in Percy correctly — `tests/index.html` doesn't load the same fonts!

### DIFF
--- a/app/index.html
+++ b/app/index.html
@@ -10,7 +10,7 @@
 
     {{content-for "head"}}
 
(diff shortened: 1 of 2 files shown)


fotinakis:
Yuuuuuup:
screen shot 2016-12-02 at 2 33 08 pm

backspace:
🤦🤦🤦

I have the font installed locally so I never noticed this discrepancy 😱

I opened #852 to get a Percy build for this. Thank youuuuuu

PR #857, opened 2016-12-06
Hide cron form if there are no branches

lislis:
Looks like this
screen shot 2016-12-06 at 11 53 19

### DIFF
--- a/app/templates/settings.hbs
+++ b/app/templates/settings.hbs
@@ -55,7 +55,7 @@
       {{#if sortedBranchesWithoutCron.length}}
           {{add-cron-job branches=sortedBranchesWithoutCron}}
       {{else}}
-          <p class="text-medium">There are no branches you can currently add Crons to.</p>
+          <p class="text-medium">There are no branches you can currently add crons to.</p>
       {{/if}}
 
     </section>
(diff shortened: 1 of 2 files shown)


app/templates/settings.hbs
```diff
@@ -43,15 +43,20 @@
...
         {{#each model.cronJobs.jobs as |cron|}}
           {{cron-job cron=cron}}
         {{/each}}
       </ul>
-      {{add-cron-job branches=sortedBranchesWithoutCron}}
+
+      {{#if sortedBranchesWithoutCron.length}}
+          {{add-cron-job branches=sortedBranchesWithoutCron}}
+      {{else}}
+          <p class="text-medium">There are no branches you can currently add Crons to.</p>
```
aakritigupta: IMO we should keep `crons` in small case here

app/templates/settings.hbs
```diff
@@ -43,15 +43,20 @@
...
+        <p class="text-medium display-inline"><a href="https://github.com/travis-ci/beta-features/issues/1" title="Give your feedback on Crons"  target="_blank">We'd love your feedback</a></p>
 
       <ul class="settings-list--crons">
         {{#each model.cronJobs.jobs as |cron|}}
           {{cron-job cron=cron}}
         {{/each}}
       </ul>
-      {{add-cron-job branches=sortedBranchesWithoutCron}}
+
+      {{#if sortedBranchesWithoutCron.length}}
```
backspace: This is no big NBD but note that you can drop `.length` here with the same result, as in handlebars, `if` on an empty array will be skipped.

PR #861, opened 2016-12-07
update signin route and template

lislis:
The template can now be used with signin and auth route.
Instead of a blank screen we'll have Mr T and a message that we're signing the user in.


backspace:
The code looks good to me, can you think of a way I can see it in action? But merge at your leisure 😀

lislis:
@backspace you can sign out and then go to `/signin` or `/auth` and click the button

PR #873, opened 2016-12-20
Replace `renderTemplate` hook with simpler `templateName` property

locks:


backspace:
Thanks! I tried this locally and it appears to cause the Pull Requests tab to not load. It’s not immediately clear to me why that would happen. I’m in the middle of a couple of things right now but I’ll come back to this.

locks:
@backspace I can also do some local testing later tonight, if necessary! Fixed the trailing whitespace (_shakes fist at GitHub editor_)

locks:
@backspace `templateName` only changes the template and not the controller it seems, which was necessary in this case. Thanks @ef4 for the tip!

backspace:
oh nice, thanks!

PR #874, opened 2016-12-20
Removes `renderTemplate` that only calls super

locks:
Given that the hook was only calling super, the route should continue working properly if removed.


backspace:
Thanks again!

PR #876, opened 2016-12-20
Remove unused `renderDefaultTemplate` method

locks:
GitHub's code search isn't all that, but this method doesn't seem to be called from anywhere, and it's checking if itself exists before calling itself?


backspace:
Hmm, yeah, I imagine this is leftover from some old implementation. I’ve been finding other such artefacts scattered about. Thanks again!

PR #877, opened 2016-12-20
Add Ember Data beta/canary deployments

backspace:
As inspired by @runspired’s [appearance on Ember Weekend](http://emberweekend.com/episodes/run-ember-data-canary).

My current plan is for these to appear at:
* `ember-data-beta.travis-ci.*`
* `ember-data-canary.travis-ci.*`

I don’t know if the combinatoric explosion of having Ember Data beta/canary combined with Ember beta/canary is worth it. Maybe?


runspired:
:D 👍 

You could configure the tests to run ember stable against ED beta/canary, but obviously that comes with the trade off of more test runs.

backspace:
Yeah, I’m thinking about doing that on the nightly cron build that happens on `master`, at least. I think there’s some disagreement within the team about how big our build matrix should be 😬

PR #884, opened 2016-12-28
Remove empty div with empty conditional

gusaiani:
Hi all, 

saw this `<div>` containing an empty conditional, seemed safe to delete it.

Cheers!


backspace:
hah, so much dead code, thank you!

PR #885, opened 2016-12-28
Upgrade x-select to 3.x

Robdel12:
This is currently working off of x-select master because I want to
make sure there aren't any bugs before cutting 3.0. All the tests
pass! But I can't seem to poke around running the server to verify.

### DIFF
--- a/app/components/cron-job.js
+++ b/app/components/cron-job.js
@@ -19,19 +19,5 @@ export default Ember.Component.extend({
 
   delete: task(function* () {
     yield this.get('cron').destroyRecord();
-  }).drop(),
-
-  actions: {
-    setSelectedBranch(value) {
-      this.set('selectedBranch', value);
-    },
-
-    setSelectedInterval(value) {
-      this.set('selectedInterval', value);
-    },
-
-    setSelectedOption(value) {
-      this.set('selectedOption', value);
-    }
-  }
+  }).drop()
 });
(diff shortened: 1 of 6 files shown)


backspace:
oooh, thanks for getting started! I was planning to get to this soonish, I didn’t actually realise the v3 change was happening until I was looking into that Edge problem.

What’s happening with running the server? Does it not let you log in properly?

Robdel12:
Heh, I quickly started the server and got a blank white screen. Then hopped right into tests. 

Just gave it another shot and everything works beautifully :D I took some notes while upgrading so I'm going to incorporate those into the [migration guide](https://emberx-select.pagefrontapp.com/migration-guide) and then probably cut 3.x

backspace:
Excellent!

Do you at The Frontside prefer to avoid the `mut` technique and use controller actions instead?

Robdel12:
We actually prefer the template mut helper but I fell back to old habits this AM :p

backspace:
hah no problem. I didn’t really even “get” `mut` until recently but it seems like it’s nice in this case. Thanks for all your attention on this!

Robdel12:
It took me a little while for it to click with me too. Lots of watching @flexyford & @cowboyd using awesome template helpers.

And no problem! I've been enjoying my 2 weeks of OS time :)

backspace:
ah that’s a great end-of-year activity!

I’ve run this locally to confirm that all is well. Merging!

PR #893, opened 2017-01-05
Fix handling of carriage returns

backspace:
This is a potential fix for [the issue reported here](https://github.com/travis-ci/travis-ci/issues/7106). Since I’m barely familiar with what’s going on here, I’ll be asking for input.


backspace:
I‘ve been given thumbs-ups from people with more knowledge in this area, so merging and deploying!

PR #910, opened 2017-01-17
Refactor avatar

lislis:
Refactoring around avatars
- [x] Step 1: make sure user-avatar component is used everywhere we use avatars
- [x]  Step 2: make avatar look smooth on high-dpi devices

### DIFF
--- a/app/models/commit.js
+++ b/app/models/commit.js
@@ -43,29 +43,14 @@ export default Model.extend({
     'committerName',
     'committerEmail',
     function () {
-      let namesMatch = this.get('authorName') === this.get('committerName');
-      let emailsMatch = this.get('authorEmail') === this.get('committerEmail');
-      return namesMatch && emailsMatch;
+      if (this.get('committerEmail') !== 'noreply@github.com') {
+        let namesMatch = this. get ('authorName') === this. get ('committerName');
+        let emailsMatch = this. get ('authorEmail') === this. get ('committerEmail');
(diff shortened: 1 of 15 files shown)


app/models/commit.js
```diff
@@ -43,29 +43,14 @@ export default Model.extend({
     'committerName',
     'committerEmail',
     function () {
-      let namesMatch = this.get('authorName') === this.get('committerName');
-      let emailsMatch = this.get('authorEmail') === this.get('committerEmail');
-      return namesMatch && emailsMatch;
+      if (this.get('committerEmail') !== 'noreply@github.com') {
```
drogus: I'm not sure if this is needed. If `committerEmail` is noreply@github.com `authorEmail` could still be different and we should show both.

app/models/user.js
```diff
@@ -151,9 +152,5 @@ export default Model.extend({
     user = JSON.parse(this.get('sessionStorage').getItem('travis.user'));
     user[name.underscore()] = this.get(name);
     return this.get('sessionStorage').setItem('travis.user', JSON.stringify(user));
-  },
-
-  avatarUrl: Ember.computed('email', function () {
-    return this.get('externalLinks').gravatarImage(this.get('email'), 36);
```
drogus: I think that with this removal we could also remove `externalLinks: service()` from the beginning of the class.

lislis:
@drogus I made the changes you requested! wydt?

PR #912, opened 2017-01-19
add beacon instruction

lislis:
The HS beacon documentation only offers the 'instruction' property to be set. This is would be the easiest way to inform users to include a link in their message.
screen shot 2017-01-19 at 16 38 55

### DIFF
--- a/app/utils/init-hs-beacon.js
+++ b/app/utils/init-hs-beacon.js
@@ -4,6 +4,7 @@ export default function initHsBeacon() {
 
   HS.beacon.config({
     modal: true,
-    attachment: true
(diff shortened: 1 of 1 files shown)


BanzaiMan:
@lislis I think this is a good start.

lislis:
Sweet. I'm merging and shipping and then we can iterate

drogus:
Is there a way to send arbitrary data with HS payload? Maybe we could restore the old behaviour of attaching the URL ourselves?

lislis:
@drogus it's not mentioned in the Beacon docs directly but there might be a way to hack it in?

PR #931, opened 2017-01-31
Fix sentry issue 187862253

clekstro:
While we generally need to overhaul this code, this should quiet down our logging in the interim.


clekstro:
Tested on PR env, works fine. Merging.

PR #932, opened 2017-01-31
Prefer use of fat arrow syntax

clekstro:


app/components/remove-log-popup.js
```diff
@@ -17,7 +17,7 @@ export default Ember.Component.extend({
 
       return job.removeLog().then(() => {
         this.get('flashes').success('Log has been successfully removed.');
-      }).fail((xhr) => {
+      }, (xhr) => {
```
backspace: `catch` perhaps?

clekstro: I just reverted to what it was before, though I think it might be clearer if we used `catch`/`error`/`fail` in the future instead of the implicit failure. We seem to use this pattern in multiple places atm.

backspace: yeah, I think explicit is better than “second argument represents failure”, I just thought [`catch`](http://emberjs.com/api/classes/Ember.PromiseProxyMixin.html) was the only option (or `finally`).

PR #935, opened 2017-01-31
Bump Sentry version

clekstro:
Getting warnings when viewing errors on Sentry, as well as bug fixes.


clekstro:
Tested on PR env, looks good.

PR #938, opened 2017-02-02
adjust web persmission check to match api permisson check

lislis:


lislis:
Closing this. Turns out the behaviour in web is correct and that it was an api bug

PR #940, opened 2017-02-03
Remove unused request route

backspace:
This appears not to be linked from anywhere and when I manually
construct a path, it doesn’t render properly. I’ll check in
with others just in case.

Here’s an example:

image

It looks like @drogus is the only in-team person to have touched `request.hbs`, what do you think about this removal?


backspace:
I’m just going to remove it.

drogus:
:+1: this was prepared for a detailed view, which we experimented with in the past, but we never ended up using it and I forgot to remove it.

PR #946, opened 2017-02-07
7291: Fix incorrect date in title attribute

stewx:
Fix issue with showing date of Dec 31, 1969 described here: https://github.com/travis-ci/travis-ci/issues/7291

### DIFF
--- a/app/templates/components/branch-row.hbs
+++ b/app/templates/components/branch-row.hbs
@@ -33,7 +33,7 @@
     </div>
   </div>
   <div class="row-calendar">
-    <div class="row-item" title="{{pretty-date branch.last_build.finished_at}}">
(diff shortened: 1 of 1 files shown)


BanzaiMan:
Thanks for opening this PR. Please fix the failing spec.

drogus:
Thanks for the PR. I think that the problem with failed test is a result of an incorrect syntax. When using `if` statement inside of an attribute you shouldn't put braces around if ([more info here](https://guides.emberjs.com/v2.2.0/templates/conditionals/)). In the same time we use a helper there and I'm not sure if you can use a helper with an if inside an attribute.

In general the best way to fix it would be to fix the [`pretty-date` helper itself](https://github.com/travis-ci/travis-web/blob/b2dd006de5b1cff77d6cf3a256204e9f98941b1c/app/helpers/pretty-date.js) to make it return an empty string if the argument passed to it is `null`. A nice side effect is that it will fix all of the places where we use it and where the date might be `null`

PR #947, opened 2017-02-08
Set pro/enterprise feature flags on runtime

drogus:
from the commit message:

```
Historically ENV vars were not enabled when building an app on Heroku,
so we couldn't rely on any checks made in environment.js file (like for
example checking for process.env.TRAVIS_PRO). At some point it must have
changed, because on travis-pro the config is properly compiled with a
proper value for `pro-version` feature flag, but the Travis CI
Enterprise setup relies on the old method of updating the config on
runtime.

This commit makes the ruby server update the pro-version and
enterprise-version feature flags at the time of making request and thus
it remove reliance on availability of ENV vars during build time.

Additionally we don't know if accessing ENV vars when deploying to
heroku is something supported, so this way of doing things is more
future proof. That said, once we confirm that we can rely on availability
of ENV vars during Heroku deploys we could think about simplifying the
config code.
```

### DIFF
--- a/waiter/lib/travis/web/app.rb
+++ b/waiter/lib/travis/web/app.rb
@@ -185,7 +185,17 @@ class Travis::Web::App
       # TODO: clean up
       config = {}
 
-      config['enterprise'] = options[:enterprise] if options[:enterprise]
+      config['featureFlags'] ||= {}
+
+      if options[:pro]
+        config['pro'] = true
+        config['featureFlags']['pro-version'] = true
+      end
+      if options[:enterprise]
+        config['enterprise'] = true
+        config['featureFlags']['enterprise-version'] = true
+      end
+
       if config['enterprise']
         config['pagesEndpoint'] = false
         config['billingEndpoint'] = false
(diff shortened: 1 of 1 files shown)


clekstro:
@drogus Just had a look and this appears to be officially supported (unless I've misunderstood something): https://devcenter.heroku.com/articles/nodejs-support#environment-variables

drogus:
Nice, I'll have to think about possibility of removing *a lot* of the config manipulation that we do (or, quire possibly, all of it).

bnferguson:
:clap: :clap: :clap: Thanks so much! I don't think I would have ever caught this. 

> This commit makes the ruby server update the pro-version and
> enterprise-version feature flags at the time of making request and thus
> it remove reliance on availability of ENV vars during build time.

Oh interesting! I wasn't aware that env vars were pulled in during the  build, if that's the case that should be something we can do during the enterprise build phases. Though if it's customer config, it's a bit trickier. Have to think what we could do there, especially when we break `travis-web` into its own Docker container.

drogus:
> I wasn't aware that env vars were pulled in during the build, if that's the case that should be something we can do during the enterprise build phases.

It really depends if you ever change the config without rebuilding travis-web. If it sometimes happens then it would be safer to stay with runtime values injected in there, otherwise we could think about making enterprise to behave more like Heroku now does in that regard.

bnferguson:
> It really depends if you ever change the config without rebuilding travis-web.

In the current model, yes. The customer may change hostnames, turn on and off services (librato etc).  And only thing that happens is a server restart. travis-web is built before it ever gets to the customer.

PR #949, opened 2017-02-09
Add error when environment variable deletion fails

backspace:


drogus:
If this returns 404 maybe we could be a little bit descriptive, saying that we failed to find the env var and asking to reload the page?

drogus:
To be more precise, I'm thinking about checking a response code and displaying a generic flash message for everything except 404 and displayin a specific flash message for 404

PR #953, opened 2017-02-17
Extract contents of BasicRoute into BaseRouteMixin

clekstro:
This is the first step in a series of refactorings to hopefully make the
transition to engines a bit easier.  Because we inherited from
`BasicRoute` in the Application route (and that route referenced another
controller), it was impossible to add a routable engine, as you can't
currently share controllers across the app/engine boundary, only
services/components.

This will also ensure that we're not doing more work than we need to. We
currently opt-in to a lot of behavior and overwrite framework hooks when
it's not entirely clear if that has to happen or not. We'll need to look
into this further to see if it's possible to break down the extracted
`BaseRouteMixin` further into even smaller chunks.

This doesn't actually fix the issue I highlighted with the controller
dependency, it's just the first step in that direction. Because I
converted the previous inheritance implementation, this should behave
identically.

### DIFF
--- a/app/mixins/base-route.js
+++ b/app/mixins/base-route.js
@@ -2,7 +2,7 @@ import Ember from 'ember';
 
 const { service } = Ember.inject;
 
-export default Ember.Route.extend({
+export default Ember.Mixin.create({
   auth: service(),
 
   activate() {
--- a/app/routes/abstract-builds.js
+++ b/app/routes/abstract-builds.js
@@ -1,7 +1,7 @@
 import Ember from 'ember';
-import TravisRoute from 'travis/routes/basic';
+import BaseRouteMixin from 'travis/mixins/base-route';
 
-export default TravisRoute.extend({
+export default Ember.Route.extend(BaseRouteMixin, {
   templateName: 'builds',
   controllerName: 'builds',
 
--- a/app/routes/account.js
+++ b/app/routes/account.js
@@ -1,6 +1,7 @@
-import TravisRoute from 'travis/routes/basic';
+import Ember from 'ember';
+import BaseRouteMixin from 'travis/mixins/base-route';
 
-export default TravisRoute.extend({
+export default Ember.Route.extend(BaseRouteMixin, {
   titleToken(model) {
     if (model) {
       return model.get('name') || model.get('login');
--- a/app/routes/accounts.js
+++ b/app/routes/accounts.js
@@ -1,6 +1,7 @@
-import TravisRoute from 'travis/routes/basic';
+import Ember from 'ember';
+import BaseRouteMixin from 'travis/mixins/base-route';
 
-export default TravisRoute.extend({
+export default Ember.Route.extend(BaseRouteMixin, {
   model() {
     return this.store.query('account', {
       all: true
--- a/app/routes/accounts/index.js
+++ b/app/routes/accounts/index.js
@@ -1,9 +1,9 @@
 import Ember from 'ember';
-import TravisRoute from 'travis/routes/basic';
+import BaseRouteMixin from 'travis/mixins/base-route';
 
 const { service } = Ember.inject;
 
-export default TravisRoute.extend({
+export default Ember.Route.extend(BaseRouteMixin, {
   auth: service(),
 
   redirect: function () {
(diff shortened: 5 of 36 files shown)


drogus:
Another way to tackle it could be to change the `error` controller to a service, but I'm fine with both. The only worry that I have with changing inheritance to mixin is that it potentially might change the way routes work, because there's a chance that we relied on the order of calling route methods. That said, if that's the case, I'd rather fix any places where we do it than do a defensive programming.

clekstro:
@drogus I like the idea of moving the error controller over to a service 👍 . I just didn't tackle it here because I wanted to get these small changes merged more often than not to avoid things getting behind and having to maintain a long-running branch.

I definitely tried to keep the order of the current implementation intact (because I'm not 100% what the consequences could be were that to change). My assumption was that inheriting from the base route would be equivalent to ensuring that the refactored mixin was the first to be mixed in. Assuming that's true, the method order should be intact, right?

clekstro:
Added the label so we don't merge before the MVD release. I'd rather deal with the potential conflicts than delay that work.

drogus:
> My assumption was that inheriting from the base route would be equivalent to ensuring that the refactored mixin was the first to be mixed in. Assuming that's true, the method order should also be the same, right?

Yes, you're right.

That said, I thought about it a little more, because I felt that this change might result in a different behaviour and I realised it's not about the difference between inheritance and mixin, but the existence of a basic route. If you have `app/routes/basic.js` defined, it will be used as a default parent class for all of the routes, no matter if you defined them or not. Here is an example twiddle: https://ember-twiddle.com/3f1b5dec01105861fe1fb46b7cb689a1 - a foo route is just defined in a router and when you visit it, it will run the `console.log` statement from the basic route.

I'm not sure if we already have any routes that are defined in `router.js` and don't have an explicit definition, but regardless I think that it's safe to assume that someone can forget to add a mixin if they don't define the class (I know that you could argue that in the same way one can forget to inherit from a basic route when you indeed define the class explicitly, but I'd argue that's easier to spot).

clekstro:
@drogus TIL. That's super interesting, as I've never seen this convention of a `basic` route before. I wasn't able to find any public documentation on this either but stumbled across a few issues where this was mentioned:

- https://github.com/emberjs/ember.js/issues/4474
- https://github.com/emberjs/ember.js/pull/4483

I'm not sure which mistake would be more common/easier to make and given the implicit `basic` behavior you make a great point. I prefer using mixins over inheritance almost on principle (due to the limitations/"hidden" behavior that inheritance imposes), but I wouldn't want to impose that on us to the detriment of stability. But given that our tests all pass and there's no obvious breakage, if we ship and something does break, I more see that as an opportunity to improve our test suite than a reason to avoid making the switch. It also sounded like that `basic` behavior will go away in the future.

drogus:
TL;DR: In spite of some of the objections I have, I think it's OK to merge it. I'd personally prefer to still use a basic route for now, but I'm fine with both solutions.

> I prefer using mixins over inheritance almost on principle (due to the limitations/"hidden" behavior that inheritance imposes)

I'd argue that in this particular case using a mixin isn't really that much different than inheritance, but that would be an entirely different and long discussion into OOP principles ;)

> But given that our tests all pass and there's no obvious breakage, if we ship and something does break

The problem is that the kind of code that you moved to the mixin is impossible to test in a way we catch all the possible problems. The biggest problem is the part, where we [do an auto sign in](https://github.com/travis-ci/travis-web/blob/9ddac89c274f736d6063ff2cc68a5755186cb624/app/mixins/base-route.js#L16-L18). This is needed only when you enter any given route as a first route and you're not signed in yet. In order to test it, you would have to have an acceptance test that checks if you're signed in from `localStorage` after first visit. There are also other things that would be hard to test.

All that said, almost all of our routes have an explicit class defined, so I think it's still fine to merge it, I just want to point out that the kind of code that is expected to run every time you enter a route is hard to test, because there's too much permutations to test (which, I guess, is a problem of acceptance/integration tests in general).

clekstro:
@drogus Sounds good. This definitely isn't meant as the final destination, just the first step.

backspace:
Is this okay to close? Or do you want to recover it, @clekstro?

PR #970, opened 2017-03-06
V3 builds

clekstro:
There are still some open items to address, namely:

- [x] BUG: Fix broken build history (unrelated builds are showing as well as PR's)
Builds controller is used for both builds and PR's, which is part of the reason this isn't working.

- [ ] Consider refactoring app/services/store.js with new v3 conversion of build
payload coming from pusher

- [x] Check if certain attrs for app/models/build.js can be removed now that we're using v3

- [x] Remove duplication in mirage for builds (we request v3 sometimes with header
and sometimes via url). Ensure we always request v3 builds the same way (preferably via the v3 header).

- [x] Look into filtering/improving mirage implementation in general

- [x] Fix current implementation to avoid complexity around circular dependencies
(copying API behavior to keep track of which records we've already serialized).

- [x] Serializer unit tests (potentially restore deleted tests).

- [x] There's also potentially just some leftover cleanup/commented out code that should be deleted that I haven't cleaned up yet.


clekstro:
The percy diffs seem quite innocuous. We don't assert everything that the visual diffs catch, so I'm fine with either fixing the tests (probably by changing the way we build up the mirage associations in the tests) or simply approving them as-is. Some of the changes are different, but expected. For instance, I changed a few assertions to simplify/split out the tests, which caused some of the failures.

app/components/branch-row.js
```diff
@@ -41,7 +41,8 @@ export default Ember.Component.extend({
         };
       }
       let path = `${apiEndpoint}/v3/repo/${repoId}/builds`;
-      let params = `?branch.name=${branchName}&limit=5&build.event_type=push,api,cron`;
+      // eslint-disable-next-line
+      let params = `?branch.name=${branchName}&limit=5&build.event_type=push,api,cron&sort_by=finished_at:desc`;
```
drogus: This doesn't seem like something that belongs to this PR, meaning that this changes the current functionality of how the branch page looks like

backspace:
I haven’t looked at this thoroughly, but there‘s something weird happening with the first Percy diff. Here’s the `master` rendering:

image

Here‘s what’s showing on this branch:

image

The [changes I made](https://github.com/travis-ci/travis-web/commit/f318ca4e1517f0ad62b45a6797fd8b5dd961499e#diff-b4db497c9196c4e1ce5f53f61c6c2069) to the “complex log” acceptance test are seemingly missing, though when I actually look at the code in this branch, that’s not the case.

Maybe this will just disappear if another commit gets made to this PR? Confusing.

tests/acceptance/repo/build-list-routes-test.js
```diff
@@ -127,32 +92,60 @@ test('build history shows, more can be loaded, and a created build gets added an
     assert.ok(page.builds(2).errored, 'expected the third build to have errored');
 
     assert.ok(page.showMoreButton.exists, 'expected the Show More button to exist');
-
-    assert.equal(page.builds(3).name, 'rarely-used', 'expected the old default branch to show');
```
backspace: I wonder why this has been removed? I made this assertion specifically to test for some weirdness, but perhaps the weirdness is now gone.

image

drogus: I'm not sure, maybe Curtis removed it to make the test pass when working on it and then he forgot to put it back? I'll revert this and see how it goes.

drogus:
@backspace thanks for the info! this seems to not have been caught by the tests.

I still need to do some manual testing, so hopefully I spot any other issues if there's anything more.

mirage/config.js
```diff
@@ -295,17 +248,42 @@ export default function () {
...
-     * The modelName causes Mirage to know how to serialise it.
-     */
-    return this.serialize({
-      models: builds.models.reverse(),
-      modelName: 'build'
-    }, 'v3');
+      * TODO remove this once the seializers/build is removed.
+      * The modelName causes Mirage to know how to serialise it.
+      */
+    return this.serialize(builds, 'build');
```
clekstro: Is this still necessary?

mirage/serializers/build.js
```diff
@@ -1,13 +1,3 @@
...
-    const response = object.attrs;
-
-    if (object.commit) {
-      response.commit = this.serializerFor('commit').serialize(object.commit, request);
-    }
-
-    return response;
-  }
-});
+export default V3Serializer.extend({});
```
clekstro: I think Mirage will fall back to using the `Application` serializer if we delete this, right? Given that inherits from `v3`, I think this could be deleted.

package.json
```diff
@@ -69,6 +69,7 @@
     "glob": "^4.5.3",
     "loader.js": "^4.2.3",
     "morgan": "^1.7.0",
+    "testem": "1.15.0",
```
clekstro: I think this is left over from one of my attempts to fix the constant rebuild issues I was having. I'm fine if we leave it in, but it can probably be done in a separate PR. If we have to revert, it'd be nice not to have the churn related to the testem version.

tests/acceptance/builds/pull-requests-test.js
```diff
@@ -0,0 +1,81 @@
+import { test } from 'qunit';
```
clekstro: 👏

tests/acceptance/builds/view-pull-request-test.js
```diff
@@ -0,0 +1,47 @@
+import { test } from 'qunit';
```
clekstro: 👏

tests/acceptance/builds/pull-requests-test.js
clekstro: I guess this comes out weird in the review diff. Really just 👍 on adding another test.

PR #978, opened 2017-03-10
Remove beta feature link from cron settings

backspace:
This is ready for when crons exit beta 😄


aakritigupta:
Wow that was too quick! I barely mentioned this 15minutes ago!

backspace:
haha well the PR is teeny and deletion-only 😆

PR #983, opened 2017-03-13
Use Babel 6 for transpilation.

rwjblue:
Figured I'd share how to get the decorators working (and I likely need to update documentation in ember-computed-decorators now).


ember-cli-build.js
```diff
@@ -33,8 +33,10 @@ module.exports = function () {
     'ember-cli-babel': {
       includePolyfill: true,
     },
-    babel: {
-      optional: ['es7.decorators']
+    // need to use babel6 key here until
+    // we can upgrade to ember-cli@2.13
+    babel6: {
```
backspace: ahhhhh `babel6` was the key, thanks!

rwjblue: Yeah, sadly using babel with older ember-cli makes it try to use these babel 6 plugins for ember-cli's own internal transpilation (which is still using babel@5).

backspace:
I’ve finally returned to this. In my v informal local testing, this nearly halved initial build time. Excellent!

PR #985, opened 2017-03-14
Remove usage of /v3 prefix

drogus:
There's a mix of /v3 prefix and Travis-API-Version header usage in the
app, which makes it hard to maintain the mirage config. This commit
unifies usage of V3 calls to always use a header version.

### DIFF
--- a/app/components/status-images.js
+++ b/app/components/status-images.js
@@ -30,11 +30,12 @@ export default Ember.Component.extend({
 
       if (this.get('auth.signedIn')) {
         options.headers = {
-          Authorization: `token ${this.auth.token()}`
+          Authorization: `token ${this.auth.token()}`,
+          'Travis-API-Version': '3'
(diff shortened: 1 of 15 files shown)


app/components/status-images.js
```diff
@@ -30,11 +30,12 @@ export default Ember.Component.extend({
 
       if (this.get('auth.signedIn')) {
         options.headers = {
-          Authorization: `token ${this.auth.token()}`
+          Authorization: `token ${this.auth.token()}`,
+          'Travis-API-Version': '3'
```
backspace: This seems like it shouldn’t be in the conditional, right?

backspace:
Excellent!

app/components/status-images.js
drogus: Good catch and thanks for fixing this.

PR #987, opened 2017-03-14
Use ember-exam to split the tests

drogus:
This commit splits the tests and runs 2 instances of test runners, so
the tests finish faster

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,7 +39,7 @@ before_script:
   - echo $PERCY_ENABLE
 
 script:
-  - ember try:one $EMBER_VERSION --skip-cleanup=true --- ember test --reporter dot
+  - ember try:one $EMBER_VERSION --skip-cleanup=true --- ember exam --reporter dot --split=2
 
 before_deploy:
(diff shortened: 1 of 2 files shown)


clekstro:
I'm generally in favor of this (I had previous concerns about test flakiness that I saw with the last PR). One thing I'm curious about is the current build only appears to be running half of the tests. 2 cores makes sense given the [build environment](https://docs.travis-ci.com/user/ci-environment/#Virtualization-environments) though 👍 .

Should we also modify this so this runs locally as well?

clekstro:
Looks like we accidentally added `ember-exam` in #992.

drogus:
@clekstro yes, I think that I didn't remove it after I was splitting my work into different branches. This is just in npm, though, so it still doesn't change anything.

Regarding the test splitting I'm not really sure it helps  a lot. When I was testing it initially I saw improvement, but tests for this PR ran slower than the previous build. I think it's due to a fact that VMs run slower during busier times. I'd need to do more testing to figure out what the gain really is.

PR #992, opened 2017-03-15
Unify mirage serializers

drogus:
At the moment mirage serializers are not organised in any specific way. This PR is meant to refactor some of the code and clear serializers implementation.


drogus:
I pretty much finished work on this, here is a message from the commit that does the actual work here:

```
This commit unifies V3 Mirage serializers by making all of the V3
serializers inherit from a base V3 serializer. Base V3 serializer
creates a payload based on API spec fetched from API's index
(https://api.travis-ci.org/v3). This ensures that Mirage's responses
will be much closer to reality than what we had right now.

On top of that, with API's index as a base for creating a response, it
was extremely easy to implement `include` query parameter support. So
now if a request uses `include` parameter, we will properly return a
standard representation of included records.

V3 Mirage serializer also has a mechanism similar to API's mechanism for
avoiding circular dependencies. There might be a situation when a record
embeds another record, which in turn embeds the original record. For
example a repository may embed a build (for example as `current_build`
attribute) and the build will embed the repository then. This would
normally result in a circular dependency problem (a repository embeds a
build that embeds a repository that embeds a build that embeds... etc).
In such situation API doesn't embed a full record, but only record's
@href. So the response could look like:

    {
      "@type": "repository",
      "@href": "/repository/1",
      "id": 1,
      "current_build": {
        "@type": "build",
        "@href": "/build/1",
        "id": 1,
        "repository": {
          "@href": "/repository/1"
        }
      }
    }

We mimic this mechanism in V3 serializers now.

This commit also needed some changes in V2 Mirage serializers and some
of the tests, because of the change in V3 payloads.
```

The only problem is that there're some weird test failures that I don't experience locally with a failure message `Assertion after the final `assert.async` was resolved`. I thought that maybe it's a problem with PhantomJS, but it seems that using Chrome doesn't help.

drogus:
There were some differnces in Percy outputs, but I approved them, because all of them are a result of changes in the test setup. In some places we show more info (like slug when there was no slug present) and in some places we show less (for example when branch is not loaded). Ideally I would like to have the UI always have everything that should be there on Percy, but at the moment it would mean a lot of work that would need to be reworked once we switch to V3.

PR #1003, opened 2017-03-21
Try fetching job's config multiple times if it doesn't succeed

drogus:
From commit message:

```
Not all of the endpoints return job's config. Because of that we try to
fetch job's config if it doesn't exist. The problem is that till now we
were trying to fetch it only once and we were giving up if the record
was loading. That means that if a config property is accessed while
loading, we will just return it as `undefined` and not attempt a second
try. I haven't noticed this problem in production, but it happened
during tests, so I guess it could happen on production as well.

This commit fixes the situation by retrying until we can do an actual
request.
```

I haven't added tests here, because I'm not sure what would be the best way to reliably test it. I'll think about it.

### DIFF
--- a/app/models/job.js
+++ b/app/models/job.js
@@ -64,14 +64,26 @@ export default Model.extend(DurationCalculations, {
     let config = this.get('_config');
     if (config) {
       return _object.pickBy(config);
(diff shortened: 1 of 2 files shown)


drogus:
I've added a test, so it's good for a review.

PR #1004, opened 2017-03-21
Update Ember and family to 2.12

backspace:
This is significantly broken, but I’m creating this to save my progress while I work on other things.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -4,10 +4,9 @@ module.exports = {
     ecmaVersion: 6,
     sourceType: 'module'
   },
-  parser: 'babel-eslint',
   extends: 'eslint:recommended',
   env: {
-    'browser': true
+    browser: true
   },
   rules: {
     // TODO: Remove this to ensure we handle errors properly in UI
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
+# See https://help.github.com/ignore-files/ for more about ignoring files.
+
 # compiled output
 /dist
 /tmp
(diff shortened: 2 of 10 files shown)


clekstro:
I also started having a look at this this morning. It looks like upgrading Ember-CLI itself is fine (just bumping `ember-cli` to `2.12.1`), but after following the standard upgrade guide the main issue is that the tests run and run until the build times out.

I'll dig into this a bit more throughout the day and see if we need to open up an issue with the Ember-CLI folks.

clekstro:
Strangely enough, this appears tied to upgrading `ember-data`. @drogus had mentioned some async stuff related to testing at EmberConf, but not sure if that's relevant here without looking more into the CHANGELOG and seeing what's actually different in the most recent version.

drogus:
> had mentioned some async stuff related to testing at EmberConf

I don't think that it's relevant, unfortunately

One of the problems that I mentioned is that we don't wrap some log operations (in `app/models/log.js`) in a run loop and also log rendering code uses setTimeout, which is also not wrapped, so tests will not wait for log rendering (that's why we have to use `waitForElement` in some places). 

Another thing was that some tests were passing locally and failing on Travis CI, because locally some of the `andThen`s were not run at all. I believe that in all of the cases that I ran into it was a legitimate error in the app or in tests, but it was just swallowed by test runner.

clekstro:
This appears to have been "introduced" within `ember-data` within the first beta cycle. No idea yet if this is on their end or ours though, as I'm honestly not at all sure how to go about debugging this. I've got the rest of the upgrade working as expected in #1032.

clekstro:
Closing in favor of #1032.

PR #1006, opened 2017-03-22
Use host without path when dealing with hrefs from API V3

drogus:
We use API's returned @href in cron's adapter. As API returns the full
path, it will also return any prefix that server uses, for example if
the API is ran at https://example.com/api, the returned @href might look
like `/api/v3/repo/1`. That's why when we use hrefs from the API V3 we
need to attach them to a bare host, without a path prefix.


bnferguson:
This fixed it! :tada:

PR #1026, opened 2017-03-29
rename user_settings to the new settings for the API V3 endpoint change

renee-travisci:


renee-travisci:
I've verified this is working on the PR branch that was deployed and checked with @drogus.

PR #1027, opened 2017-03-30
Demonstrate placeholder UI

backspace:
This is but an experiment! I’ve been thinking about it for so long I just wanted to see how it was.


backspace:
GIF for your delectation:

placeholder

joshk:
😍

lislis:
Awesome!
I think I'd like it even better without the icons. just grey bars that then fill up with icons and content

backspace:
yeah, @lislis, that’s a question I had, of what exactly to make look placeholder-y and what should be rendered. I’ll make an issue to get design feedback.

joshk:
This is super awesome

@backspace are we able to apply the same idea to the Repo/Build/Job main content section?

backspace:
heh of course, but things are more of an entangled mess there so I picked something easy to start with 😬

I do have partial progress on the main panel showing placeholders and it feels quite zippy, as the slug is already known so there‘s no reason to not render it right away.

PR #1032, opened 2017-04-03
Upgrade Ember things to 2.12 (except Ember Data)

clekstro:
There were some issues that we encountered within #1004. This fixes those except for the very strange behavior seen with more recent versions of Ember Data (within the test suite).


clekstro:
I was able to run a `git bisect` to determine the particular Ember Data [commit](https://github.com/emberjs/data/commit/a5a8cd7714555655ae9fb7209b5b87c11aabd1cc) that first introduced the strange test issue we're seeing, though I'm not sure what the underlying issue is yet.

clekstro:
Merging this for now. We can fix the Ember Data issues in a subsequent PR.

PR #1037, opened 2017-04-06
Fix serializers for nested relationships

drogus:
This is yet another problem with serializers that I ran into when working on #970. When API returns nested relationships, we should put all of the records with representation set to `standard` into the `included` array. The problem is that with code currently in production it will only do that for relationships in the first level. As with other issues that I found, it doesn't affect us on production, because in all cases that we use V3 API, we don't actually deep nested records to be pushed in the store, but it would be nice to get it working consistently. And I want to fix it before merging #970 to isolate any problems that it might cause.

### DIFF
--- a/app/serializers/build.js
+++ b/app/serializers/build.js
@@ -63,6 +63,20 @@ var Serializer = V2FallbackSerializer.extend({
       // value. This line makes V3 payloads also populate pull_request property
       resourceHash['pull_request'] = true;
     }
+    // when we get the commit from V3 we need to put a branch on a commit as the
+    // current commit model expects it
+    if (type && commit && resourceHash.branch && !commit.branch) {
+      let name = resourceHash.branch.name;
+      if (!name) {
+        let match = resourceHash.branch['@href'].match(/\/repo\/.*?\/branch\/(.*)/);
+        if (match) {
+          name = match[1];
+        }
+      }
+      if (name) {
+        commit.branch = name;
+      }
+    }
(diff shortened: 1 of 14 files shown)


clekstro:
@drogus Thank you for your saintly patience in working out these edge cases 🙌 .

drogus:
I added more info, here's a commit message from one of the commits:

```
Fix serializers to work with deep nested relationships
There's a problem in a way we normalize V3 records into JSON API format.
When we extract relationships from a top level resource, we don't
normalize any of the related records, which means that their
relationships are lost. Given a payload like this for a repository:

    {
      @type: 'repository',
      @href: '/repo/1'
      id: 1,
      current_build: {
        @type: 'build',
        @href: '/build/1'
        id: 1,
        jobs: [
          {
            @type: 'job',
            @href: '/job/1
            id: 1
          }
        ]
      }
    }

We would run normalize on the payload, we would loop through the
relationships (`current_build` only in this case) and add relationships
to a `relationships` property. Because we don't run `normalize` on the
`current_build` relationship, we don't get any info about related jobs.

This commit changes the V3 serializer to run `normalize` on related
records, so we essentially end up with a tree of records. After we do
this, we also add all of the records that have a `standard`
representation to the `included` array in order to siedload them into
the store (`relationships` hash is just a set of references to related
records, it isn't supposed to carry extended record info).
```

I added also [a bit of documentation to v3 serializer](https://github.com/travis-ci/travis-web/blob/4d85becb4c39bd42dfc4d8c60313fb4c2c9d017e/app/serializers/v3.js#L6-L93)

PR #1047, opened 2017-04-11
Remove live network requests from test suite

clekstro:
The test suite currently hits our production API in a few instances. This addresses most of them, though there's still an outstanding issue with a `DELETE` of an SSH key that I haven't solved yet.


clekstro:
Ok, the `DELETE` mentioned before has been removed. Merging.

PR #1048, opened 2017-04-11
Trigger custom build

lislis:


app/components/requests-item.js
```diff
@@ -3,27 +3,32 @@ import { computed } from 'ember-decorators/object';
...
   isGHPages(message) {
     return message === 'github pages branch';
   },
 
-  @computed('request.isAccepted')
-  requestClass(isAccepted) {
-    return isAccepted ? 'accepted' : 'rejected';
+  @computed('request.result')
+  requestClass(result) {
+    return result;
```
backspace: This change means this be expressed more simply as an alias: `@alias requestClass: null`

app/templates/components/trigger-custom-build.hbs
```diff
@@ -0,0 +1,32 @@
...
+<div class="form-elem {{if jsonError 'form-error'}}">
+  <label for="trigger-build-config" class="form-label display-inline">Custom config</label><p class="form-helptext display-inline m-l-s">Enter config as YAML</p>
+  {{textarea value=triggerBuildConfig on-change=(action (mut triggerBuildConfig)) placeholder="script: echo 'Hello, World'" id="trigger-build-config"}}
+</div>
+
+<div class="form-elem">
+  {{#if triggering}}
+    {{loading-indicator}}
+  {{else}}
+    <button {{action 'triggerCustomBuild'}} type="button" class="trigger-build-submit {{if jsonError 'button inactive' 'button--blue'}}" title="Trigger custom build with your settings">Trigger custom build</button>
```
backspace: Since this action does nothing but call `perform` on the task, you can bypass the need for an action using the `perform` template helper here directly. There’s an example [here](https://ember-concurrency.com/#/docs/tutorial/refactor), though I wish I could link directly to it! In this case, you could replace the `action` with `onclick={{perform sendTriggerRequest}}`.

app/controllers/index.js
```diff
@@ -10,7 +10,6 @@ export default Ember.Controller.extend({
   @service tabStates: null,
   @service('updateTimes') updateTimesService: null,
   @service statusImages: null,
-  @service popup: null,
```
backspace: hah why was this even here 😆

mirage/config.js
```diff
@@ -325,6 +325,33 @@ export default function () {
     return this.serialize(builds, 'build');
   });
 
+  this.post('/repo/:repo_id/requests', function (schema, request) {
+    const requestBody = JSON.parse(request.requestBody);
+    const fakeRequestId = 5678;
+    let repository = schema.find('repository', { id: request.params.repo_id });
```
backspace: Since finding by id is the most common case, `schema.find('repository', request.params.repo_id)` would do this too. But this way is fine too 😀

package.json
```diff
@@ -60,6 +60,8 @@
     "ember-inflector": "^2.0.0",
     "ember-keyboard-shortcuts": "0.3.0",
     "ember-load-initializers": "1.0.0",
+    "ember-modal-dialog": "^2.2.0",
+    "ember-percy": "1.2.12",
```
backspace: Did this maybe sneak in by mistake during a merge?

package.json
```diff
@@ -75,10 +77,12 @@
...
-    "visibilityjs": "^1.2.4"
+    "visibilityjs": "^1.2.4",
+    "yamljs": "^0.3.0"
   },
   "engines": {
     "node": "6.9.5"
   },
-  "private": true
+  "private": true,
+  "dependencies": {}
```
backspace: I suspect this is unnecessary, but maybe you found it was needed by something?

tests/acceptance/repo/trigger-build-test.js
```diff
@@ -0,0 +1,84 @@
+import { test } from 'qunit';
+import moduleForAcceptance from 'travis/tests/helpers/module-for-acceptance';
+import triggerBuildPage from 'travis/tests/pages/trigger-build';
+
+moduleForAcceptance('Acceptance | repo/trigger build', {
+  beforeEach() {
+
+  }
+});
```
backspace: If there’s nothing in `beforeEach`, the entire second argument to `moduleForAcceptance` is safely left out.

tests/acceptance/repo/trigger-build-test.js
```diff
@@ -0,0 +1,84 @@
...
+
+test('trigger link is not visible to not logged in users', function (assert) {
+  const repo = server.create('repository', {
+    name: 'difference-engine',
+    slug: 'adal/difference-engine'
```
backspace: Ada Lovelace GIF 💞

tests/acceptance/repo/trigger-build-test.js
```diff
@@ -0,0 +1,84 @@
...
+    sha: 'c0ffee'
+  });
+
+  repo.currentBuild = latestBuild;
+  repo.save();
+
+  triggerBuildPage.visit({ slug: repo.slug });
+
+  andThen(() => {
+    assert.equal(currentURL(), 'adal/difference-engine', 'we are on the repo page');
```
backspace: This seems unneeded to me…??? But I’m open to suggestion hehe

tests/acceptance/repo/trigger-build-test.js
```diff
@@ -0,0 +1,84 @@
...
+  });
+
+  assert.ok(triggerBuildPage.popupIsHidden, 'modal is hidden');
```
backspace: This probably needs to be in the `andThen` to be a reliable assertion.

tests/acceptance/repo/trigger-build-test.js
```diff
@@ -0,0 +1,84 @@
...
+    assert.ok(triggerBuildPage.popupIsVisible, 'modal is visible after click');
+  });
+
+  triggerBuildPage.selectBranch('master');
+  triggerBuildPage.writeMessage('This is a demo build');
+  triggerBuildPage.writeConfig('script: echo "Hello World"');
+  percySnapshot(assert);
+  triggerBuildPage.clickSubmit();
+
+  andThen(() => {});
```
backspace: ugh is this One Of Those Times 😢

tests/unit/services/trigger-build-popup-test.js
```diff
@@ -0,0 +1,12 @@
...
+moduleFor('service:trigger-build-popup', 'Unit | Service | trigger build popup', {
+  needs: ['service:auth']
+});
+
+test('it toggles the isShowingFlag', function (assert) {
+  let service = this.subject();
+  assert.equal(service.get('isShowingTriggerBuildModal'), false, 'initally the modal is hidden');
+  service.toggleTriggerBuildModal();
+  assert.equal(service.get('isShowingTriggerBuildModal'), true, 'after toggling the modal is visible');
+});
```
backspace: IMO this is already covered by the acceptance test. But there’s no One True Way about which tests to rely on, so if you feel like this should be here, that’s fine too.

backspace:
I pushed that empty commit just to see if it would fix the PR deployments, but it didn’t help. I don’t see anything in the logs that explains why the deployments are broken. 😢

lislis:
Yeah, strange. I get a 403 on the assets. Maybe that's a place to start?

tests/acceptance/repo/trigger-build-test.js
lislis: I wanted to have this as starting url to show that the transition later is happening and important. Does that make sense?

tests/acceptance/repo/trigger-build-test.js
lislis: yeah, without it the tests were super flaky

tests/unit/services/trigger-build-popup-test.js
lislis: I'd rather have the base functionality of the service documented in the test as well. Just to be sure :D

lislis:
Thanks for the review @backspace that was super helpful! I think I addressed all the things and left comments on the others

tests/acceptance/repo/trigger-build-test.js
backspace: Yeah, I just thought that `triggerBuildPage.visit` would do exactly this. But it’s not a problem to leave. 😄

app/components/trigger-custom-build.js
```diff
@@ -0,0 +1,89 @@
...
+        config: config
+      }
+    };
+
+    if (! Ember.isEmpty(this.get('triggerBuildMessage'))) {
+      body.request.message = this.get('triggerBuildMessage');
+    }
+
+    try {
+      yield this.get('ajax').postV3(`/repo/${this.get('repo.id')}/requests`, body)
```
lislis: @backspace I just remembered why I had the extra `isTriggering` flag. The `sendTriggerRequest` task yields here but there is the nested api call that is then not considered. So `sendTriggerRequest.isRunning` returns false while we still wait for the other call.

I couldn't get it so that it would also wait for the 2nd call. Do you have any ideas?

backspace: hmmmm maybe you need a `return` inside the `Ember.run.later`? Then it should get added to the promise chain and be considered part of the task.

But now that I think about it, maybe there should be more `yield`s here, like in [this example](http://ember-concurrency.com/#/docs/examples/autocomplete)…??? I think there could be some flattening out of this getting-to-pyramid-esque-ness

app/components/trigger-custom-build.js
```diff
@@ -0,0 +1,98 @@
...
+    if (! Ember.isEmpty(this.get('triggerBuildMessage'))) {
+      body.request.message = this.get('triggerBuildMessage');
+    }
+
+    try {
+      this.get('ajax').postV3(`/repo/${this.get('repo.id')}/requests`, body)
+        .then((data) => {
+          let reqId = data.request.id;
+
+          Ember.run.later(this, function () {
```
lislis: @backspace I tried splitting up the task into two but I still have trouble getting the yield and return right so that the task only returns when the 2nd task finishes. I think the Ember.run might be a problem?

backspace: hmm yeah… what about something the “works” example [here](https://github.com/machty/ember-concurrency/issues/26#issuecomment-192498503)?

Maybe it doesn’t exactly align though… perhaps it’s not worth getting too deep into trying to get it to be more ember-concurrency-y, the way it was did look to me to have an unnecessary flag but I don’t have The Answer about how to get rid of it! 😢

lislis: No worries. I'll bring back the extra flag for now

mirage/config.js
```diff
@@ -325,6 +325,33 @@ export default function () {
     return this.serialize(builds, 'build');
   });
 
+  this.post('/repo/:repo_id/requests', function (schema, request) {
```
clekstro: This doesn't seem like a very typical use case, so probably fine to make it a one-off response with raw JSON. That said, it might be good to add a mirage model/serializer if we're going to work with these endpoints more in the future.

package-lock.json
```diff
@@ -0,0 +1,14020 @@
+{
```
clekstro: Should this file be included? We haven't been able to figure out a way to get npm5/node 8 deployed yet, but not sure if this will cause issues or not 🤔

package.json
```diff
@@ -60,7 +60,9 @@
     "ember-inflector": "^2.0.0",
     "ember-keyboard-shortcuts": "0.3.0",
     "ember-load-initializers": "1.0.0",
+    "ember-modal-dialog": "^2.2.0",
     "ember-percy": "1.2.13",
+    "ember-modal-dialog": "^2.2.0",
```
clekstro: Duplicate entry. Npm removed it when I ran `npm i` locally.

app/components/repository-layout.js
```diff
@@ -17,10 +22,30 @@ export default Ember.Component.extend({
     return this.get('externalLinks').githubRepo(slug);
   },
 
+  @computed('repo.id', 'repo.defaultBranch.name')
```
lislis: good catch! I think I can actually get if from the repo model as well and remove this completely!

mirage/config.js
lislis: I'm not sure if/where/when we'd use it again atm. I'd leave it like this?

clekstro: 👍

PR #1050, opened 2017-04-12
Allow unused args

drogus:
I think that not allowing unused vars is not giving us any value and it
usually makes life harder. Despite the fact Javascript has dynamic
signatures, I usually like to still define them even if I don't use all
the args, because the signature hints on how to call the function. This
is especially important when you intend to override a function in
subclasses - you may want to define an empty (or almost empty) default
and start using it by passing all of the arguments and only use some of
the arguments in different subclasses.

The only advantage of not allowing unused arguments is that we will know
when an argument is misspelled, but that should be also caught by other
rules (like: don't allow using undefined variables).

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -10,6 +10,9 @@ module.exports = {
     browser: true
   },
   rules: {
+    // Don't allow unused vars, but allow unused arguments
+    "no-unused-vars": ["error", { "vars": "all", "args": "none", "ignoreRestSiblings": false }],
+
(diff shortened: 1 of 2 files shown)


clekstro:
@drogus Sounds totally reasonable. As you mentioned the `undefined variables` rule, do you happen to know if that will be triggered more often as a result of this change, i.e. we leave all arguments in the signature but then have to remove them based on the `unused` rule?

backspace:
I have no strong feelings about it so if you prefer this change, fine with me!

PR #1051, opened 2017-04-12
Fix broken GitHub issue link in build title

clekstro:
The GitHub issue link (see https://travis-ci.org/travis-ci/travis-web/builds/221251257)
was being incorrectly generated because the `owner` attribute
we were using was assumed to be a string, but was actually an object.

This meant we were generating
`https://github.com/[object%20Object]/travis-web/issues/1047` in
production.

I'm not sure what caused this, but applying this fix corrects the links
without breaking other tests.


clekstro:
@drogus Thanks for helping to confirm this shouldn't cause other regressions 🥇

PR #1053, opened 2017-04-18
Directly transition to first_sync instead of invoking action

clekstro:
Based on our previous spelunking (as well as reported errors in Sentry), we currently appear to have a race condition where the first route
transition has not always completed when we attempt to call an action on the `ApplicationRoute`.

Given the `renderFirstSync` action simply redirects, we should be able to use the (currently private) `routing` service as an interim replacement.

If the exception is what is keeping users from proceeding through the typical auth flow, this should hopefully address it.

### DIFF
--- a/app/routes/application.js
+++ b/app/routes/application.js
@@ -105,10 +105,6 @@ export default TravisRoute.extend(BuildFaviconMixin, KeyboardShortcuts, {
         return true;
       }
     },
-
-    renderFirstSync() {
-      return this.transitionTo('first_sync');
-    },
   },
 
   afterSignIn() {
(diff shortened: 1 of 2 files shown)


testclekstro:
This doesn't appear to be the cause of the issue. Digging in further.

clekstro:
Given this wasn't the source of the authentication issues, I think we should wait until the router API is public. Going to close this for now.

clekstro:
Given this is still causing issues in production, I'm going to move forward on this. We're already using the private service in other parts of the code, so we can simply replace with the public API when it lands.

PR #1056, opened 2017-04-18
Build stages again

backspace:
These are build stages-related changes reapplied to the current version of the build V3ification branch. I haven’t managed to merge the Mirage build serialiser yet so the stages test is failing, but this should restore the build history tab.


backspace:
Closing in favour of #1070.

PR #1068, opened 2017-04-25
V3 jobs

drogus:


drogus:
It mostly works now. It needs to wait for travis-ci/travis-api#563 in order to not break the running jobs tab, but all of the other places that use jobs work.

The only caveat is that, since the job payload from V3 doesn't include the config, I still need to hit V2 endpoint to get the config. But this is the only place where it happens, so fixing it later will be very simple.

drogus:
Tests pass and it works, I'll do more testing and probably deploy it early next week.

drogus:
For some reason PR builds fail and pushes pass here (even though I updated to have everything from the `master` branch). I don't want to waste time on debugging that, so I think I'll just merge it when it's ready and tested and see if there're problems on master.

PR #1069, opened 2017-04-25
Use Chrome --headless for testing

clekstro:
Given that phantom.js' maintainer is stepping down and this would enable us to test directly on our most popular browser, this seems like the most sensible path going forward.

I'm able to run tests like this both on macOS and Linux, but not sure of Windows support.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,8 +33,6 @@ cache:
     - bower_components
 
 before_install:
-  - "export DISPLAY=:99.0"
-  - "sh -e /etc/init.d/xvfb start"
   - "npm config set spin false"
   - npm install -g bower
   - bower --version
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,7 +33,7 @@ cache:
     - bower_components
 
 before_install:
-  - "npm config set spin false"
+  - npm config set spin false
   - npm install -g bower
   - bower --version
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,13 +20,22 @@ matrix:
 sudo: false
 dist: trusty
 
+addons:
+  apt:
+    sources:
+    - google-chrome
+    packages:
+    - google-chrome-stable
+
 cache:
   directories:
     - node_modules
     - bower_components
 
 before_install:
-  - npm config set spin false
+  - "export DISPLAY=:99.0"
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -20,13 +20,22 @@ matrix:
...
+
 cache:
   directories:
     - node_modules
     - bower_components
 
 before_install:
-  - npm config set spin false
+  - "export DISPLAY=:99.0"
+  - "sh -e /etc/init.d/xvfb start"
```
simi: Is xvfb really needed with `--headless` flag?

clekstro: @simi I'm honestly not sure. We shall soon find out!

clekstro: @simi Nope! Thanks for pointing this out 👍

backspace:
I’m unable to run the tests locally via `ember test` at the moment, it fails with this error:

```
not ok 1 Chrome - error
    ---
        message: >
            Error: Browser exited unexpectedly
            Non-zero exit code: null
            Stderr:
             [0425/154445.394309:ERROR:resource_bundle.cc(767)] Failed to load /Applications/Google Chrome.app/Contents/Versions/58.0.3029.81/Google Chrome Framework.framework/Versions/A/resources.pak
            Some features may not be available.
            [0425/154445.484346:WARNING:dns_config_service_posix.cc(154)] dns_config has unhandled options!


        Log: |
            { type: 'error', text: 'Error: Browser exited unexpectedly' }
            { type: 'error', text: 'Non-zero exit code: null' }
            { type: 'error',
              text: '[0425/154445.394309:ERROR:resource_bundle.cc(767)] Failed to load /Applications/Google Chrome.app/Contents/Versions/58.0.3029.81/Google Chrome Framework.framework/Versions/A/resources.pak\nSome features may not be available.\n[0425/154445.484346:WARNING:dns_config_service_posix.cc(154)] dns_config has unhandled options!\n' }
    ...

1..1
# tests 1
# pass  0
# skip  0
# fail  1
```

I’m hoping this will just cure itself; I can still run them by visiting `http://localhost:4200/tests`

clekstro:
@backspace Ugh, that's no good. I was apparently using Chrome 57 when I tested this on my Mac, which appears to have just ignored the `--headless` option silently. #1071 should fix this.

PR #1070, opened 2017-04-25
Add build stages rendering

backspace:
This is yet another continuation, finally based on `master`. Should be the last one! 🤞


clekstro:
I wanted to make sure that my refactoring in #1072 didn't break this/cause pain in getting it merged in. The rebase I performed was clean and the tests are passing, so trusting that it doesn't break things.

backspace:
It appears to have caused no problems! 🎉

backspace:
itshappening.gif

PR #1071, opened 2017-04-26
Detect OS and set testem browser args accordingly

clekstro:
The use of Chrome --headless breaks on Chrome versions under 59 on macOS/Windows. This detects the OS version and ensures we only run headless on Linux for the time being.


backspace:
Thanks for finding this. I’m looking forward to when this works locally!

PR #1073, opened 2017-04-27
Fix a typo in the markup

eiriksm:
Just randomly came across what either is a typo, or an interesting new type of attribute called `clas`. If it is in fact the latter, then I apologise for the clutter :)

Just wanted to take this opportunity to thank travis for the awesome service which I almost literally use every day.

Have a nice day!


clekstro:
@eiriksm Thanks!

PR #1079, opened 2017-04-28
Add missing tests concerning missing data/fix small bug

clekstro:
When visiting a build tab, we're currently missing some assertions based on the tab we're currently on having no data.  This effort also resulted in surfacing a bug, as we currently show `No builds for this repository` messages on the `Branches` tab, which based on the expected template logic is incorrect.

### DIFF
--- a/app/controllers/branches.js
+++ b/app/controllers/branches.js
@@ -1,6 +1,12 @@
 import Ember from 'ember';
 
+const { controller } = Ember.inject;
+
 export default Ember.Controller.extend({
+  repoController: controller('repo'),
+
+  tab: Ember.computed.alias('repoController.tab'),
+
   defaultBranch: Ember.computed('model', function () {
     return this.get('model').filterBy('default_branch')[0];
   }),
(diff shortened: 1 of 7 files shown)


backspace:
I’m happy to see more tests! This isn’t worth extra work but I think with page objects, something like `showsNoBuildsMessaging` might be better called `noBuildsText` or whatever, as the assertion is against the text, not the showing of it. But this isn’t a critical thing, just stylistic/wording. 😀

There’s a newish feature in the page objects library where what they call a `component` comes with a bunch of [default attributes](http://ember-cli-page-object.js.org/docs/v1.8.x/components#default-attributes), so it could even be written like this:

```
noBuildsMessage: { scope: '.missing-notice h2.page-title' }
```

Then you can make assertions against the attributes:

```
assert.ok(page.noBuildsMessage.isVisible, 'expected the no builds message to be visible');
assert.equal(page.noBuildsMessage.text, '… whatever text
…
```

But for a single assertion like you have here, that seems like overkill.

PR #1092, opened 2017-05-04
update PreciseEOL banner to announce the image update

MariadeAnton:


backspace:
Maybe the time is unnecessary? Since this will be showing forever into the future, I assume?

PR #1093, opened 2017-05-04
Replace Uglify with Babili

backspace:
My current understanding is that the choice to exclude IE 9 from
the supported browsers has resulted in ES2015 code making it past
the transpliation step, since the supported browsers now all
support some parts of ES2015. Uglify can’t handle ES2015, so
all deployments have failed since that was merged in. This is the
recommended solution I found in this thread:
https://github.com/ember-cli/ember-cli/issues/7006


backspace:
What would happen if you viewed the site with IE 9? I suspect nothing would render, which seems less than ideal.

backspace:
After having looked at browser usage statistics and the way things currently render in IE 9 (not at all), this seems fine. On Monday, I’ll remove IE 9 from the targets and merge and deploy this.

backspace:
I’m going to hold off on this because it also breaks other browsers (Safari 8 and 9, for instance) because of the ES2015 that makes it through. At the least I think we should be able to handle such a failure and display a message that people need a newer browser.

I have to prioritise build stages but can look into that afterward.

PR #1106, opened 2017-05-15
Remove beta flag on auto-cancellation

backspace:


joshk:
@backspace I'm ready when you are. Then I'll do the beta-features update dance.

backspace:
Okay, @joshk, I’ll wait for the `master` tests to pass, then deploy, and message you.

PR #1107, opened 2017-05-17
Upgrade to Ember 2.13

clekstro:
This completes the upgrade to 2.13, but does so through some significant refactoring.
Here are some of the changes:
- custom location code (URL rewrites) no longer exists. Instead, we vary the behavior of the index page by your auth status (signed in, you see repositories at `/`, otherwise you see the landing page)
- the nesting of multiple routes within the `main` route/controller has been removed
- Remove `sendToApp` as part of sign-in/sign-out. Those actions are now centralized within the application route.
- Remove legacy `!#` hash rewriting rules
- Increase test coverage based on bugs I saw while testing this locally (though our suite was passing).
- Added ember-route-action-helper. This was what enabled us to get away from using `sendToApp` from the auth service. I talked more about my thinking in https://github.com/travis-ci/travis-web/pull/1107/commits/a84ddad5fa216e6da72e33b96537e501ef759826.


drogus:
In general it looks good, I really like getting rid of the custom location code and I really like removing auth events in favour of simple function calls - this is something that I wanted to fix for a long time now.

One caveat about this is that it changes URL behaviour - the purpose of a custom location was to always be on `/` regardless you're signed in or not. Maybe it's ok, but I'd like to think about it more and discuss it.

clekstro:
@drogus If that's true, it's definitely not intended. Do you happen to have an example?

app/routes/basic.js
```diff
@@ -31,9 +31,6 @@ export default Ember.Route.extend({
     return this.get('auth.currentUser');
   },
 
-  // on pro, we need to auth on every route
-  needsAuth: Ember.computed.alias('features.proVersion'),
```
drogus: Why did you remove it?

app/routes/index.js
```diff
@@ -0,0 +1,61 @@
...
+  activate(...args) {
+    this._super(args);
+    if (this.get('auth.signedIn')) {
+      this.get('tabStates').set('sidebarTab', 'owned');
+      this.set('tabStates.mainTab', 'current');
+      this.controllerFor('repos').activate('owned');
+    }
+  },
+
+  deactivate() {
```
drogus: I'm not sure if this is needed here, could you explain the reasoning behind this change?

app/routes/main/index.js
```diff
@@ -6,10 +6,8 @@ export default TravisRoute.extend({
       if (this.get('features.dashboard')) {
```
drogus: I think that this file could be removed as you removed the `main` route.

drogus:
>  If that's true, it's definitely not intended. Do you happen to have an example?

Sorry, that's my mistake, I thought that on `master` we don't redirect when you enable the dashboard and you go to `/`.

app/routes/basic.js
clekstro: Good catch! iirc I was having some test issues that I was trying to work around. I had made a note to revisit it, but simply forgot.

app/routes/index.js
clekstro: I just copied over the implementation from `app/routes/repo/index.js`, as the behavior is identical in terms of showing the current build on the index route now when a user is logged in.

app/routes/main/index.js
clekstro: Yep. Done.

clekstro:
It looks like the job clocks aren't being updated for some reason (though the builds update fine). Looking into it.

app/templates/main/error.hbs
```diff
@@ -1,7 +0,0 @@
-{{#if slug}}
```
clekstro: @drogus This seems like a reasonable change given the removal of the `main` route, but I don't want to have made our UX worse with this PR. We didn't have test coverage around this, so not sure if we really care about this or not. Thoughts?

app/routes/index.js
drogus: Ah, right, I forgot that now you need to display repo page here.

app/templates/main/error.hbs
drogus: This should be rendered when a repository is not found, for example: https://travis-ci.org/foobarbaz/barbazfoo. The way it worked was that we set the `slug` here: https://github.com/travis-ci/travis-web/blob/master/app/routes/repo.js#L55-L63 and then `main/error` would be rendered by default when an error is thrown in the routes. So I think that the proper change here would be to move this code to `app/templates/error.hbs` (if `slug` is present render "repo not found", otherwise render whatever we're rendering there now)

clekstro: @drogus I just pushed up a new set of commits which I think should address this. The UI is slightly different, but I personally think it's an improvement. Instead of showing the loading indicator in the top left corner next to the sidebar, we use the same loading content as when the application boots. This makes transitioning between the routes feel seamless, and simply shows the error message in place of the loading indicator.

You can see this live in action on https://ce-ember-2-13.test-deployments.travis-ci.org/foobarbaz/barboofar.

clekstro: screenshot from 2017-05-30 14-58-10

clekstro: Also, I should mention that I wanted to use the "tears" version of the logo, but we didn't have an `.svg` so it didn't look as crisp. @lislis do you happen to know if we have that?

lislis: I'm afraid we don't have it as SVG. Or at least not somewhere I have access to it. Maybe @ctrlaltjustine knows more about it?

clekstro: For reference, this is what current prod looks like:
screen shot 2017-05-31 at 10 48 30 am

drogus: I like the new look, but the text looks hard to notice below the image, could you try making it bigger and adding a more space between them?

clekstro:
One last thing I need to fix. I laughed when I saw it, because we literally just talked about it yesterday: I used the wrong header. We show the unauthenticated version regardless of auth status, which isn't right.

app/templates/main/error.hbs
clekstro: loadinganimation

drogus: Sure, it's fine for me to deal with it in another PR.

ctrlaltjustine:
> I'm afraid we don't have it as SVG. Or at least not somewhere I have access to it. Maybe @ctrlaltjustine knows more about it?

There's no SVG for that. It was made many moons ago and I don't seem to have a live file for that art. 

Perhaps you could use the no builds found artwork instead — https://travis-ci.org/travis-repos/loomio/pull_requests?

clekstro:
@ctrlaltjustine Sure. Apologies for not looping you in sooner, I just really wanted to get the meat of this out sooner rather than later. The error messaging was a minor piece of this work and cropped up rather late in the process. I'll try the new image out in the am and submit a PR for you to review.

PR #1114, opened 2017-06-02
Load `/about` via lazy engine

clekstro:
The assets should still be moved out of the root `public` folder an into the addon, but it's cool to see this being possible.

Still need to do some measuring around how much of an impact this makes (positive or negative) before merging.


backspace:
Since `about` has been extracted entirely to its own application, I’m closing this.

PR #1117, opened 2017-06-06
Use cone image instead of logo for missing repo message

clekstro:
@ctrlaltjustine As promised in #1107. Could you take a look if you get a chance?


ctrlaltjustine:
@clekstro I'm having difficulty getting this to run locally — is there a staging URL I can view 👀?

clekstro:
@ctrlaltjustine Sure. You can see the new image [here](https://ce-change-repo-error-image.test-deployments.travis-ci.org/foobarbaz/asdfasdfasdf).

ctrlaltjustine:
@clekstro thanks for hooking me up with that URL!

I took a look at it and I think it'd be a good idea to try and use more of the existing styles we have for cases like [no builds (as seen here)](https://travis-ci.org/travis-ci/heaven). Perhaps since @lislis did the work on those styles, she can help you out with re-using them?

Additionally I'd like us to have a more friendly tone for the copy. Something a bit more personal/approachable than a traditional `does not exist` and the ability to redirect someone to the docs to get started. Here's a quick comp of what I had in mind:

unknown-repository-01

lislis:
The issue here is that this page is only shown if the repo really doesn't exist, not on Travis not on GitHub. So setting up would mean actually going to GH and creating a repo.
How about:
```We couldn't find that repository
Maybe there is a typo in owner/repo?
```
or something along those lines?

lislis:
Also, the styles for the messages can be found there https://travis-ci.org/lislis/swaj just copying the classes should work ™️

ctrlaltjustine:
@lislis thanks for that explanation. That makes sense. Let's tweak the copy as you suggested to something that might indicate a misspelling or bad link and remove the docs button.

screen shot 2017-06-07 at 2 03 56 pm

clekstro:
A different approach was taken in #1206. Closing.

lislis:
@clekstro ah, sorry. Totally forgot this still existted :/

PR #1119, opened 2017-06-07
Pagination protoype for dashboard

lislis:
looks like this:
screen shot 2017-06-07 at 17 25 55

### DIFF
--- a/app/components/pagination-navigation.js
+++ b/app/components/pagination-navigation.js
@@ -13,16 +13,8 @@ export default Ember.Component.extend({
     return this.get('inner') || 2;
   }),
 
-  currentPage: Ember.computed('pagination.offset', 'pagination.perPage', function () {
-    return (this.get('pagination.offset') / this.get('pagination.perPage') + 1);
-  }),
-
-  numberOfPages: Ember.computed('pagination.total', 'pagination.perPage', function () {
-    return Math.ceil(this.get('pagination.total') / this.get('pagination.perPage'));
-  }),
-
-  pages: Ember.computed('numberOfPages', function () {
-    let numberOfPages = this.get('numberOfPages');
+  pages: Ember.computed('pagination.numberOfPages', function () {
+    let numberOfPages = this.get('pagination.numberOfPages');
     let thresholdDisplayAll = ((this.get('outerWindow') + 1) * 2) + (this.get('innerWindow') + 1);
     let pageArray = [];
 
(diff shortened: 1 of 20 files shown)


app/components/pagination-navigation.js
```diff
@@ -0,0 +1,122 @@
...
+  @alias('collection.pagination') pagination: null,
+
+  outerWindow: Ember.computed('outer', function () {
+    return this.get('outer') || 1;
+  }),
+  innerWindow: Ember.computed('inner', function () {
+    return this.get('inner') || 2;
+  }),
+
+  currentPage: Ember.computed('pagination.offset', 'pagination.perPage', function () {
```
drogus: I think that we could move this to the `PaginatedCollection` class, so here in the component you could just say `pagination.currentPage`

app/components/pagination-navigation.js
```diff
@@ -0,0 +1,122 @@
...
+    return (this.get('pagination.offset') / this.get('pagination.perPage') + 1);
+  }),
+
+  numberOfPages: Ember.computed('pagination.total', 'pagination.perPage', function () {
```
drogus: Same as with `currentPage`

lislis:
@drogus I tried to set up the acceptance test here https://github.com/travis-ci/travis-web/pull/1119/commits/babc67a47102e611b2ab981cdb7b6e332c7ff84e#diff-f1cdb42a58626276a4d0f95a5ec7f3d8R108

But I think there is a problem with the `withFeature` test helper. It seems it doesn't flip the feature switch in the test so we can't actually test things on the Dashboard. All other tests are set to `skip` for probably exactly that reason.

Maybe we can look at that tomorrow?

lislis:
@drogus we're almost there! This test is now failing https://github.com/travis-ci/travis-web/blob/ps-lp-pagination/tests/acceptance/repo/build-list-routes-test.js#L176
It's testing loading more builds on the build history tab, so it is almost certainly related to the changes we made in mirage

drogus:
@lislis I think that it would work if you remove [these lines](https://github.com/travis-ci/travis-web/blob/fd5f50c730b24e2d98096f79cacdfc6638ed0c54/mirage/config.js#L298-L301). Now that we apply limit and offset for all of the collection endpoints in mirage, the offset there would duplicate the global offset.

lislis:
@drogus I removed the lines and tests are still failing locally

drogus:
@lislis That's curious. I'm checking it now

PR #1120, opened 2017-06-07
pusher: add new pusher debug setting to environment

igorwwwwwwwwwwwwwwwwwwww:
Allows logging pusher events to console as per https://pusher.com/docs/debugging#pusher_logging

### DIFF
--- a/app/utils/pusher.js
+++ b/app/utils/pusher.js
@@ -28,6 +28,15 @@ TravisPusher.prototype.init = function (config, ajaxService) {
   if (config.host) {
     Pusher.host = config.host;
   }
+
+  if (ENV.pusher.debug) {
+    Pusher.log = function (message) {
+      if (window.console && window.console.log) {
+        window.console.log(message);
+      }
+    };
+  }
+
   return this.pusher = new Pusher(config.key, {
     encrypted: config.encrypted,
     disableStats: true
--- a/app/utils/pusher.js
+++ b/app/utils/pusher.js
@@ -29,7 +29,7 @@ TravisPusher.prototype.init = function (config, ajaxService) {
     Pusher.host = config.host;
   }
 
-  if (ENV.pusher.debug) {
+  if (config.debug) {
     Pusher.log = function (message) {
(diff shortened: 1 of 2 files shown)


app/utils/pusher.js
```diff
@@ -28,6 +28,15 @@ TravisPusher.prototype.init = function (config, ajaxService) {
   if (config.host) {
     Pusher.host = config.host;
   }
+
+  if (ENV.pusher.debug) {
+    Pusher.log = function (message) {
+      if (window.console && window.console.log) {
```
lislis: I think the guard is a bit over the top here. We can expect all browsers these days to just work with 
`console.log();`

igorwwwwwwwwwwwwwwwwwwww: AFAIK `console.log` is sometimes not present when the console is not open? I might be mistaken though. I copy-pated it from the [pusher docs](https://pusher.com/docs/debugging#pusher_logging).

lislis: Hmm, would be new to me. But if they have it in their docs ¯\_(ツ)_/¯

igorwwwwwwwwwwwwwwwwwwww: Appears to be the case with some IE versions at least, so probs safer this way: https://stackoverflow.com/a/5473193

drogus:
@igorwwwwwwwwwwwwwwwwwwww do you plan to use it in production? One problem that I see here is that enabling it will enable it for everyone. I'd prefer to add an option to enable it on page load. There's also a way to do it directly from the console, you can run `Pusher.log = (message) => { console.log(message) }` after the page loads.

igorwwwwwwwwwwwwwwwwwwww:
@drogus it's for debugging locally, I found it extremely helpful when debugging things, and not having to paste things into the console on every page load.

If you have a different way of handling code not intended for prod, I'd be happy to change it.

drogus:
@igorwwwwwwwwwwwwwwwwwwww if it's for debugging locally, it's totally fine. I was just worried that it's for production

PR #1121, opened 2017-06-07
Add guard against null settings responses

backspace:
The API is returning `null` settings objects for `emxsys/WorldWindServerKit`, so I plan to add a guard against those. First, the tests should fail.


backspace:
The person wrote back to confirm that this made the settings page load as expected at the test deployment, so I’ll merge and deploy.

PR #1128, opened 2017-06-09
Create a Docker image on each commit and push to Quay

joshk:
This adds a build stage which builds and pushes a private Docker image to Quay.io

You can find it here: https://quay.io/repository/travisci/travis-web?tab=tags

All commits are built, all pushes are tagged with the commit ref, the branch name, and with latest if the branch is master.

In the Travis settings UI, a QUAY_ROBOT_HANDLE and QUAY_ROBOT_TOKEN is configured per quay repo for pushing the Docker image.

@bnferguson and I have been talking about this setup for all Travis apps so that Enterprise can use them.

Extra bits to note:

- The mobile redirect code is removed as it isn't used, and hasn't been for a long time
- I've added a way to disable the SSL redirect based on a env var config, which makes it easier for testing


clekstro:
This behavior should now be implemented via a combination of master (Docker config) and [this branch](https://github.com/travis-ci/travis-web/tree/nightly-quay-releases). Waiting for confirmation that it works as expected, after which we'll create a daily cron job via the web UI.

bnferguson:
> This behavior should now be implemented via a combination of master (Docker config) and this branch. Waiting for confirmation that it works as expected, after which we'll create a daily cron job via the web UI.

This is a clever way to do it, but in my read of it unfortunately doesn't help us for one of our most important base cases where we build the docker images for `enterprise-X` on any merge of a pull request into those branches. You can see how we do it over in: https://github.com/travis-ci/travis-logs/blob/master/script/docker-build-and-push#L6-L8 and how it covers a few different cases 

Additionally, I'm hesitant to do it completely differently in `travis-web` from every other project as it can make automation of things trickier. We already see this in a lot of places with `travis-web` in the Enterprise project as it's a very different project (which is understandable as Ember is not Ruby).   I'm hoping the Docker efforts lessen these things and I'd like to be able to rely on each project behaving the same when it comes to these Docker builds. Makes life a lot easier on anyone supporting them.

drogus:
@bnferguson so would it be fine to copy the `if` that you linked to? (ie. build only on`enterprise` branch, on cron and with an ENV var?)

If that's true, I'm fine with that - we don't need any special config then and we won't have slower builds in general.

bnferguson:
> @bnferguson so would it be fine to copy the if that you linked to? (ie. build only onenterprise branch, on cron and with an ENV var?)
> 
> If that's true, I'm fine with that - we don't need any special config then and we won't have slower builds in general.

Yup! That'd work perfectly well. As I said in the meta thread in dev. That's pretty much what needs to happen with all of these but I just haven't had time due to pressing support concerns and then 2.2 and trusty schedules getting to be a bit much. I may go through and do this on other projects right quick as a way of getting back into things now that I'm back from vacation however.

PR #1129, opened 2017-06-09
pusher: process branch:created and branch:deleted events

igorwwwwwwwwwwwwwwwwwwww:
refs travis-pro/team-teal#1663
refs travis-ci/travis-gatekeeper#188


igorwwwwwwwwwwwwwwwwwwww:
the goal of this patch is to get the branches dropdown on the repo settings page to update when a new branch is pushed.

i'm still having some trouble getting the UI to update.

igorwwwwwwwwwwwwwwwwwwww:
managed to get UI propagation to work with the help of @clekstro! 🎉 

next up: writing a test for this.

igorwwwwwwwwwwwwwwwwwwww:
wrote a test with @clekstro!

tests/pages/settings.js
```diff
@@ -104,6 +104,15 @@ export default PageObject.create({
     }
   }),
 
+  cronBranches: collection({
+    scope: '.form--cron form select:nth(0)',
+    itemScope: 'option',
+
+    item: {
+      // name: text()
```
clekstro: We can get rid of this now

igorwwwwwwwwwwwwwwwwwwww:
Yeah I considered that, but didn't really want to put that much extra serialization logic into gatekeeper. This seemed like a solid trade-off.

backspace:
👍🏻 yeah I hope we can eventually replace the pseudo-V2-API output that Gatekeeper produces.

app/services/store.js
```diff
@@ -68,6 +69,20 @@ export default DS.Store.extend({
...
     }
 
+    if (name === 'branch') {
+      // force reload of repo branches
+      // delay to resolve race between github-sync and live
+      Ember.run.later(() => {
+        const repo = this.recordForId('repo', data.repository_id);
+        repo.notifyPropertyChange('branches');
+        repo.get('branches').then(branches => {
+          this.set('branches.amount', branches.content.length);
```
drogus: Services are singletons, so this would set amounts of branches regardless of a repository that the branch was created for, I'm not sure if that's what you intended to do here.

clekstro: Not really. I think this is a valid edge case we should handle. We could reset the state to 0 again when we destroy the `add-cron-job` component. /cc @igorwwwwwwwwwwwwwwwwwwww

drogus: If this is used only to subscribe to in the component isn't it enough to subscribe to the repo's `branches` property (as you do `notifyPropertyChange`)?

igorwwwwwwwwwwwwwwwwwwww: Unfortunately not, for some reason the change did not propagate.

drogus: That's curious, I'll take a look into it locally

drogus: So, the problem here is that `model` here is not a repository, it's just a hash with a `branches` collection included there.

drogus:
I'd like to propose a different approach here. In a `Repo` model (we fetch branches using `store.filter` function)[https://github.com/travis-ci/travis-web/blob/04132ee3abfd6dffed372cbe4639fa8d5edac83b/app/models/repo.js#L104-L107], which means that it will get updated whenever a new branch is added to a store. Since we get the branch's name and the repository id in pusher payload, the easiest way to use it is to make a request to `/repo/{repository.id}/branch/{branch.name}` and load the result into the store.

Something like that in the store should work (this is pseudo code as `fetchBranch` doesn't exist, but I hope it shows what I mean):

```js
if (name === 'branch') {
  fetchBranch(data.branch.name, data.branch.repository_id).then(responseJSON) => {
    this.push(this.normalize('branch', responseJSON));
  });
```

igorwwwwwwwwwwwwwwwwwwww:
@drogus thanks for your feedback, that looks like a way better approach. going to give that a try!

igorwwwwwwwwwwwwwwwwwwww:
So I paired on this with @lislis and we managed to use findRecord to get the branch, which pushes it into the store. Unfortunately for some reason that does still not trigger a re-render at the UI level, so I am keeping the service for now.

@drogus If you would like to take another look into why the store update is not propagating that'd be great. But I would like to ship this patch as-is, if that's ok.

drogus:
@igorwwwwwwwwwwwwwwwwwwww the problem is with [the definition of the `branchesWithoutCron` property](https://github.com/travis-ci/travis-web/blob/1d3610b2e81bcd818bd997befbd1dd50ef06f446/app/controllers/settings.js#L14-L15):

```js
@computed('cronJobs', 'model.branches')
branchesWithoutCron(cronJobs, branches) {
```

`model.branches` dependency means that it will be updated when `model.branches` reference changes (so if the reference to `model.branches` would change), but it doesn't depend on any changes within the collection. Changing it to `@computed('cronJobs', 'model.branches.[]')` will fix the problem and the collection will be updated.

igorwwwwwwwwwwwwwwwwwwww:
Going to try that, thank you @drogus!

PR #1155, opened 2017-06-28
Deduplicate repository layout

clekstro:
Part of moving the `index` page to rendering components forced me to duplicate the logic between the `repo` template and `index`. That was meant to be temporary, and this should fix it.

@drogus @lislis @backspace One thing I'm not really sure of here is naming. I'd like for to start having stronger naming conventions/better organization between components (given we have a number of them all under `app/components/`, but I'm not sure that `repository-layout` is in line with that or if that even conveys what I intended.

I was trying to capture the non-sidebar/main section of the page, and given this component wraps all build/job information at the repository level, I think this makes sense. But I'd be happy to hear other ideas if you have them 😄

### DIFF
--- a/app/components/repository-layout.js
+++ b/app/components/repository-layout.js
@@ -1,0 +1,32 @@
+import Ember from 'ember';
+import computed from 'ember-computed-decorators';
+
+const { service } = Ember.inject;
+
+export default Ember.Component.extend({
+  statusImages: service(),
+
+  @computed('repo.slug', 'repo.defaultBranch.name')
+  statusImageUrl(slug, branchName) {
+    return this.get('statusImages').imageUrl(slug, branchName);
+  },
+
+  @computed('build.jobs.@each.{config}')
+  jobsLoaded(jobs) {
+    if (jobs) {
+      return jobs.isEvery('config');
+    }
+  },
+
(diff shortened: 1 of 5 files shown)


lislis:
I think that makes sense 👍

drogus:
I'm not sure about the naming, but I think that `repository-layout` sounds fine.

:+1: for this PR

backspace:
I don’t have a better suggestion either. I’m happy this is happening, though, I was actually just thinking it should happen last night while unable to sleep!

clekstro:
Thanks for the feedback. I'll be merging this then.

@backspace One thing that I noticed yesterday related to this was that we actually have a lot of "divergent duplication" going on across the app. When I moved the `index` page over to components, I used the content of the `repo` template to show the builds. I actually think this is the real cause behind https://github.com/travis-pro/team-teal/issues/2062, as we're simply not even trying to render build stages in https://github.com/travis-ci/travis-web/blob/master/app/templates/index.hbs#L43-L53. 

Unless I'm mistaken, we expect the same "main/layout" content across the following routes: `index`, `search`, `build`, then showing just the job data on the `job` route.  I think that extracting similar `build-layout` and `job-layout` components that include build stages would fix that bug and prevent others like it.

PR #1156, opened 2017-06-28
Remove dead initializers

clekstro:
@drogus Can you confirm that I haven't overlooked something here?

I think the `pusher` initializer was actually replaced by the pusher _instance-initializer_, but I wasn't able to find the `stylesheetManager` being used anywhere else.


drogus:
It's fine to remove it. `stylesheetManager` was basically a hack to switch between dashboard styles and the rest of the site long time ago when we were just starting the implementation and didn't want to clutter main site with extra CSS.

PR #1157, opened 2017-06-29
Additional component extractions across templates showing builds

clekstro:
We current share a good amount of boilerplate between templates (as mentioned in #1155), which this seeks to further reduce by extracting a `build-layout` component and leveraging the previously added `repository-layout` component where possible.


clekstro:
There's actually a decent amount more of duplicated computed properties as a result of this. ✂️ ✂️

clekstro:
Given the component portion of this is done and I've started creeping into controller improvements, going to merge this and proceed in another PR...

PR #1162, opened 2017-07-04
Change to using individual lodash functions

backspace:
This saves over 130kB from vendor.js!

### DIFF
--- a/app/models/build.js
+++ b/app/models/build.js
@@ -1,6 +1,6 @@
 /* global moment */
 import safelistedConfigKeys from 'travis/utils/safelisted-config-keys';
-import _object from 'lodash/object';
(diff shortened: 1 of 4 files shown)


backspace:
Thanks to @simonihmig’s [tweet](https://twitter.com/simonihmig/status/881625402430631937), I wanted to check out the visualisation of a production build of `travis-web`.

image

It looked pretty good, but `lodash` was taking up a lot of space, especially considering we only use two of its functions. Hence this PR, which removes almost all of that `lodash` space!

image

simonihmig:
@backspace Awesome, good to see this has already been helpful! And thanks for pulling me in. Had a chat with @stefanpenner, and we will try make this even easier to use and to integrate it better into ember-cli...

PR #1166, opened 2017-07-05
Add hack to subscribe to channels in background

backspace:
This is another possibility for #1159.


backspace:
While this does mean people could miss events as the backlog of channels is subscribed to, it seems worth it.

PR #1168, opened 2017-07-06
Bump Ember to 2.14.0

clekstro:


clekstro:
Seems fine in the PR env. Merging.

PR #1170, opened 2017-07-07
Glimmer3

iezer:
@mixonic Here's a branch where I've run the migrator this morning. Best to review commit by commit to see what I had to manually change. Here are some notes on things that need to be addressed. I'll look at them today.

 - No initializers are being run.

 - Two-word services are not getting looked up correctly. For example `sessionStorage: service()` has to change to `sessionStorage: service('session-storage')`. I think this is because the old resolver is converting to from camelCase to dasherized but not the new one.

 - The travis app uses a custom `Location` type and these aren't getting resolved. I think these can be fixed in the resolver and/or resolver-config. I had a fix for this on an old resolver branch.

 - There are a few places where I had to change the import path for `environment/config`, the Resolver, and the App. These should be done by the migrator. Some of these are handled by https://github.com/rwjblue/ember-module-migrator/pull/49


clekstro:
@iezer We're thrilled to be your guinea pigs!

> The travis app uses a custom Location type and these aren't getting resolved. I think these can be fixed in the resolver and/or resolver-config. I had a fix for this on an old resolver branch.

Just wanted to mention that this was removed in an earlier commit (see e71862c7). It's probably a valid use case to support, but not something that affects us anymore.

iezer:
@clekstro Good to know. Thanks for the info. I'm going to close this PR. We'll have another branch to show more progress here in a few days.

PR #1171, opened 2017-07-08
Avoid / in component names

mixonic:
This makes the repo much happier when converted to the module unification filesystem layout.


backspace:
I merged this in #1172. For reasons mysterious to me, the PR builds were never running, but when I made a PR myself, the push build did run and passed. [This PR build is still not running](https://travis-ci.org/travis-ci/travis-web/builds/251363717) as I write this but GitHub is showing it as passed because it was the same commit in my PR. I’ve somehow lost the attribution of the commit to you though! I’m sorry.

clekstro:
@mixonic Thanks. I was totally unaware of this.

PR #1172, opened 2017-07-08
Avoid / in component names

backspace:
This makes the repo much happier when converted to the module
unification filesystem layout.

(This is the same as #1171, just wondering why the jobs haven’t started yet)


backspace:
I don’t understand why the PR build for #1172 and this PR are still not running, but the push build has completed. I’m going to merge anyway… 😳

PR #1173, opened 2017-07-10
Bump Ember-Data to 2.14

clekstro:
Includes @drogus' workaround for unexpected `unloadRecord` behavior.

### DIFF
--- a/app/components/add-ssh-key.js
+++ b/app/components/add-ssh-key.js
@@ -12,10 +12,19 @@ export default Ember.Component.extend({
   didInsertElement() {
     let id = this.get('repo.id');
     let store = this.get('store');
-    const model = store.peekRecord('ssh_key', id)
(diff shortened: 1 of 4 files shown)


clekstro:
It looks like we are stuck on 2.13 for the time being. The `unloadRecord` issue is fixed, but upgrading has brought in several other `re-render` errors that aren't present in 2.13.2. It's possible that we'd need to change something on our end, but I started digging into it and don't see exactly where the issue would be coming from. Came across [this issue](https://github.com/emberjs/data/issues/4942) and it seems like a different Ember-Data bug.

clekstro:
Unfortunately, https://github.com/emberjs/data/pull/5119 doesn't appear to have fixed this. But oddly enough, I'm only seeing the issue in our test suite.

clekstro:
We've since upgraded to 2.15 in #1307, closing.

PR #1179, opened 2017-07-11
Fix build state display

clekstro:
The referenced property in the `buildState` computed property was incorrect.

This should fix https://github.com/travis-pro/team-teal/issues/2043.


clekstro:
Have tested this out for both .org and .com in the PR envs, looks good. Merging.

PR #1180, opened 2017-07-11
Randomize loading logo

clekstro:
This shows a random logo (from all the SVG variants I noticed)
when the user first boots up the app.

Implementation-wise, it might be confusing at first as to why the
`random-logo` service was added instead of making this a part of the
`loading-screen` component. The reason is that we utilize several layers of loading
templates which -- were the randomization done at the component level --
would mean that the logo might jump between available options as we transition between loading screens. This would be jarring, so I opted for making the selection when the service
is first initialized and then referencing that state whenever the
component is used.

This should likely be behind some sort of feature flag before it goes
public.

See https://github.com/travis-pro/team-teal/issues/1729 for more details.

### DIFF
--- a/app/services/random-logo.js
+++ b/app/services/random-logo.js
@@ -1,0 +1,28 @@
+import Ember from 'ember';
+
+export default Ember.Service.extend({
+  logoVariants: [
+    'Tessa-1',
+    'Tessa-2',
+    'Tessa-3',
+    'Tessa-4',
+    'Tessa-pride-4',
+    'Tessa-pride',
+    'TravisCI-Mascot-1',
+    'TravisCI-Mascot-2',
+    'TravisCI-Mascot-3',
+    'TravisCI-Mascot-4',
+    'TravisCI-Mascot-4',
+    'TravisCI-Mascot-pride',
+  ],
+
+  init(...args) {
+    this._super(args);
+    this.set('logo', this.randomLogo());
+  },
+
+  randomLogo() {
+    const logos = this.get('logoVariants');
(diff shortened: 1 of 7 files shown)


clekstro:
screenflow

backspace:
Exciting!

This won‘t work in the test deployments for the same reason the team page doesn’t work: because the asset paths don’t get fingerprinted when generated dynamically. It’s my hope to use `ember-cli-ifa` to get this kind of thing working, but [this bug](https://github.com/RuslanZavacky/ember-cli-ifa/issues/21) stopped my attempt. Eventually…

backspace:
By feature flag, do you mean the user-controlled ones, or the ones in admin? This seems good enough for everyone to see, to me 😁

clekstro:
@backspace I meant a user-facing one, though I don't think we have a mechanism to persist that behavior currently without a dedicated endpoint (which is currently tied to admin iirc). I'm ok with merging as-is as well.

app/services/random-logo.js
```diff
@@ -0,0 +1,28 @@
...
+    'Tessa-2',
+    'Tessa-3',
+    'Tessa-4',
+    'Tessa-pride-4',
+    'Tessa-pride',
+    'TravisCI-Mascot-1',
+    'TravisCI-Mascot-2',
+    'TravisCI-Mascot-3',
+    'TravisCI-Mascot-4',
+    'TravisCI-Mascot-4',
```
clekstro: Thanks! Addressed in #1187.

PR #1186, opened 2017-07-13
Add background subscription to Pusher channels

backspace:
This somewhat works, but I get subscription errors when testing
locally for 16 of the 167 channels it tries to subscribe to.
It seems to fail for the first channel of every chunk.

I’m putting this aside for now, too frustrated!


backspace:
That was with a chunk size of 10, which made it easier for local testing; thanks to @clekstro for the idea. Since the chunk size I originally chose was 1000, the problem would be less pronounced.

backspace:
This has been better solved by changing messages to be sent through the user channel instead.

PR #1190, opened 2017-07-18
Update flash messages

backspace:
These are getting a much-needed overhaul.


backspace:
From the [current `flashes` service](https://github.com/travis-ci/travis-web/blob/02d08de8fe9bb603dda729eee070188c9fa7a578/app/services/flashes.js#L36-L41):

```
  // TODO: when we rewrite all of the place where we use `loadFlashes` we could
  // rewrite this class and make the implementation better, because right now
  // it's really complicated for just displaying a flash message (especially
  // that we show only one at the moment anyway). We still get some error
  // messages from API responses in V2 that way, so I think that cleaning this
  // up once we're using V3 would be a good point.
```

I couldn’t find any calls to this function that didn’t follow the style of everything else. Do you think this is something I can clean up now, @drogus?

drogus:
@backspace the problem is not really with any calls from the code. API V2 design was wrong in some places and instead of returning meaningful status codes or at least machine readable errors we were returning a message that was supposed to be displayed in the UI, for example something like [this](https://github.com/travis-ci/travis-api/blob/c4ae7cd2d7e403d4bf1649c3c7d1d5a68d871095/lib/travis/api/app/endpoint/jobs.rb#L33-L35). As long as we use API V2 I think that the function will need to stay to handle such cases. Fortunately we're left with only a few places where we use V2.

backspace:
ah, I understand now, thanks for the link to where V2 produces these errors. 😀

PR #1192, opened 2017-07-21
Update Trusty banner

MariadeAnton:
For the incremental update of our default Linux environment from Precise to Trusty

https://blog.travis-ci.com/2017-07-11-trusty-as-default-linux-is-coming

### DIFF
--- a/app/components/job-infrastructure-notification.js
+++ b/app/components/job-infrastructure-notification.js
@@ -23,7 +23,7 @@ export default Ember.Component.extend({
   isTrustySudoFalse: Ember.computed.equal('queue', 'builds.ec2'),
 
   @computed('job.startedAt', 'job.config')
-  isTrustyStable(startedAt, config) {
+  isTrustyStable(startedAt, config = {}) {
     if (config.dist === 'trusty' && config.group === 'stable') {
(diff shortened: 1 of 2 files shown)


MariadeAnton:
Great catch, @backspace - thanks much :)

backspace:
haha I have a particular eye for links that go on slightly too long 😆 

There used to be a banner somewhat like this on `travis-web` jobs but there isn’t anymore (like [this one](https://travis-ci.org/travis-ci/travis-web/jobs/256041768))… this PR doesn’t change anything about that, but do you know if that’s deliberate? I guess it’s because it doesn’t have `sudo`?

@meatballhat asked recently about how we could possibly extract the logic for displaying these messages; I think we can still come up with a way to have flexibility here so these messages can be customised without requiring a push to production. 🤔 Something to open in [dev](https://github.com/travis-pro/dev)?

MariadeAnton:
oh, an even better catch - this update applies to all Trusty environments, so I've updated the sudo condition. What do you think?

As for adding flexibility on how we generate these messages - that'd be fantastic!

MariadeAnton:
re-requested a review @backspace so that you can check the new updates in Percy or here as needed, thanks!

backspace:
It’s back!

image

PR #1196, opened 2017-07-25
Don't always fetch stages info

drogus:
We need stages info only on a build page and then we can fetch it
lazily, so we don't need to always include it.

### DIFF
--- a/app/adapters/build.js
+++ b/app/adapters/build.js
@@ -1,7 +1,7 @@
 import V3Adapter from 'travis/adapters/v3';
 
 export default V3Adapter.extend({
-  includes: 'build.commit,build.branch,build.stages',
+  includes: 'build.commit,build.branch',
 
   pathPrefix(modelName, id, snapshot, type, query) {
     if (type === 'query' && query.repository_id) {
--- a/app/adapters/repo.js
+++ b/app/adapters/repo.js
@@ -4,5 +4,5 @@ export default V3Adapter.extend({
   defaultSerializer: '-repo',
 
   includes: 'build.branch,repository.default_branch'
-    + ',repository.current_build,build.commit,build.stages'
+    + ',repository.current_build,build.commit'
 });
--- a/app/adapters/stage.js
+++ b/app/adapters/stage.js
@@ -1,0 +1,26 @@
+import V3Adapter from 'travis/adapters/v3';
+
+export default V3Adapter.extend({
+  coalesceFindRequests: true,
+
+  groupRecordsForFindMany(store, snapshots) {
+    let stagesByBuildId = {};
+
+    snapshots.forEach((snapshot) => {
+      let buildId = snapshot.belongsTo('build').id;
+
+      if (!stagesByBuildId[buildId]) {
+        stagesByBuildId[buildId] = [];
+      }
+
+      stagesByBuildId[buildId].push(snapshot);
+    });
+
+    return Object.values(stagesByBuildId);
+  },
+
+  findMany(store, modelClass, ids, snapshots) {
+    let buildId = snapshots[0].belongsTo('build').id;
+    return this.ajax(`${this.buildURL()}/build/${buildId}/stages`);
+  }
+});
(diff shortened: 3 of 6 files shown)


carlad:
Hey @drogus  I had a look at this on the test org-production deployment. The errant repo now loads the build history without timing out 

https://ps-dont-always-download-stages.test-deployments.travis-ci.org/emc-mongoose/mongoose/builds

This is great :)

But now the Build Page isn't showing the Stage Names: https://ps-dont-always-download-stages.test-deployments.travis-ci.org/emc-mongoose/mongoose/builds/257265765

It's also a little slow to load.

I think both these things can be solved by using the new v3 `build/build.id/stages` endpoint...

drogus:
@backspace this works when you go to builds page and then navigate to the individual build, because we still get stages info even without `includes`, but with a `minimal` representation. Take a look at [the API call](https://api.travis-ci.org/v3/repo/9106952/builds?event_type=push%2Capi%2Ccron&repository_id=9106952&include=build.commit%2Cbuild.branch) that we make from the builds page that @carlad linked to.

We also get the same info when getting an individual build, so I'm not sure why it fails when you go to the build's page directly. I'll look into it.

drogus:
Actually, I'm slightly wrong in this. This works on the current build only, because we include `build.stages` in a repo request, so we get the full stages info for the current build of the repo. For the other stages we get only minimal representation and then what usually happens is that Ember Data fetches missing data, but for stages it doesn't happen. So if we fix that, it should work properly.

drogus:
I've investigated a bit more on how relationships are handled internally and unfortunately it doesn't play well with the API design. Ember Data has two primary ways of loading the records for an async relationship:

1. If we have associated `id`s then Ember Data will do the query to fetch the associated records by `id`s
2. If there's no associated records info, Ember Data will try to use a link to a relationship to fetch the full info

We're in the first situation here, because we load `id`s from a minimal representation of stages. When using API V2 we were solving this by implementing endpoints with ability to fetch records by `ids` (for example `jobs?ids=1,2,3,4`), so we could load all of the associated records at once.

So now, our options are:

1. Implement ability to fetch records by `ids` in API
2. Change our adapters to not load `stages` info if it's in a `minimal` representation and then load the entire relationship when we need it
3. Make sure that we always load the `stages` relationship manually
4. Implement partial record loading, so we could use `stages` in minimal representation and then load more info only if we need it (though this could mean that we still need to do 1.)
5. Try to change our adapters to somehow figure out how to fetch stages from the `builds/{build.id}/stages` endpoint based on stage ids (or some other info, maybe saved in the `meta`?)

I'm not sure which solution is the best. The first option is the easiest from my point of view, given how Ember Data works, but then we would have to always support it in the API (not a problem in my opinion, but we would have to get a green light from the API team). If that's not an option I would probably lean towards 5. if it can be implemented easily and if not then probably 2.

drogus:
I looked into it with fresh mind today and it seems that option no. 5 may be the easiest to implement at the moment. In the adapter in the `findMany` function we get records snapshots, so we can group them by build and then query each build for stages. I'll look into implementing it.

drogus:
Unfortunately it won't be that simple. `findMany` expects that only one request will be made (or in other words, it acts on a promise with the payload as a return value and promises can't be resolved with multiple different values).

drogus:
OK, I didn't know about `groupRecordsForFindMany`, it'll do the job.

drogus:
OK, it seems that it works, but I don't really understand the changes in percy, for some reason the message about open source disappeared from some of the pages and appeared on others. It works locally and on the test deployment, so I'm going to approve it, but we will need to look into it if it happens again.

carlad:
Just checked the repo that was timing out before and it now loads :-)
https://travis-ci.org/emc-mongoose/mongoose/builds
Thanks @drogus !

backspace:
@drogus I’ve mostly fixed the extraneous diffs in Percy but the problem with the intermittently-appearing open source header is lingering. I have [this issue](https://github.com/travis-pro/team-teal/issues/2047) for addressing it but I haven’t been able to return to it for a while. Hopefully soon!

PR #1202, opened 2017-07-31
Render github icon

lislis:


ctrlaltjustine:
I just took a look at this @lislis and visually it looks like exactly what we want/need. GH's black logo with the opacity turned down. 

👍 from me

PR #1204, opened 2017-08-01
Adding Katherine and optimizing/reducing some other assets

ctrlaltjustine:
This pull request includes:
- Adding Katherine D. (new blue member)
- Size reduction of some assets that exceeded our max 226x226
- Optimized images that had some spare bits


lislis:
@ctrlaltjustine the percy tests are a bit flaky with the `support open source` banner. We don't yet know exactly why this is, but you can approve the diffs (top right green button in percy) and then this should be mergeable

ctrlaltjustine:
@lislis cool — looks like that was all that Percy was upset about. So I've approved those and this PR is good to go.

lislis:
online now https://travis-ci.org/about

PR #1205, opened 2017-08-02
Live updates for pagination

drogus:
This is still a work in progress, but I wanted to have it on GitHub and have builds for a PR.

The implementation is somewhat complicated, but I'll try to refactor it when I'm done with all of the things that I want to implement and I'll add much more comments and documentation. One interesting thing here is that I needed to implement a better `store.filter` primitive in order to make live data updates for pagination easier and this could allow us to simplify code in some places (for example using the new `store.filter` version we don't need starred repos service and we can just return whatever filter returns).


app/utils/fetch-live-paginated-collection.js
```diff
@@ -0,0 +1,31 @@
+import LivePaginatedCollection from 'travis/utils/live-paginated-collection';
+
+// Fetches a LivePaginatedCollection instance
+//
+// It will use store.filter function in order to get a base for pagination and
+// then it will create a paginated collection with a content set to the filtered
+// collection.
+let fetchLivePaginatedCollection = function (store, modelName, queryParams, options) {
+  let dependencies = options.dependencies || [],
+    filter = filter || (() => true),
```
backspace: Did you mean this to be `filter = options.filter || (() => true),`?

drogus: Yup, I fixed it [later](https://github.com/travis-ci/travis-web/commit/8e155969afd78ce2362c72c7c3ca50ad65e72f28). Tests didn't catch it, because I tested the filtering capabilities only in `store.filter` tests, but not on `store.paginated` tests.

backspace: Oops, I didn’t see that followup PR, thanks 😄

PR #1207, opened 2017-08-03
Revive branches page project

lislis:


lislis:
rebased

lislis:
I'm getting `The result of a server query is immutable` when trying to fetch more deleted branches. Which is weird because I'm basically doing the same for active branches where it works 🤔

PR #1213, opened 2017-08-07
Fix download link for raw logs (with token)

clekstro:
The computed property signature is very odd and could probably be
simplified, but this fixes the issue.


clekstro:
This also needs a unit test :\

PR #1221, opened 2017-08-08
Add customisable flash message preamble

backspace:
This lets the default preamble be overwritten. It also adds acceptance tests around flash messages.


lislis:
yaaay

PR #1222, opened 2017-08-08
Add running tab on org

backspace:


app/components/repository-sidebar.js
```diff
@@ -68,20 +70,18 @@ export default Ember.Component.extend({
...
-    if (!proVersion) { return []; }
-
+  @computed('currentUser')
+  queuedJobs(user) {
     const queuedStates = ['created'];
     const result = this.get('store').filter(
       'job',
-      job => queuedStates.includes(job.get('state'))
+      job =>
+        queuedStates.includes(job.get('state') && user && user.hasAccessToRepo(job.get('repo')))
```
backspace: Hey, @drogus, since you mentioned `store.filter` [here](https://github.com/travis-ci/travis-web/pull/1205#issue-247375781), do you think it would be worth waiting for your changes there so I can make this actually work? 

The current problem is that `job.get('repo')` is asynchronous, of course, so `hasAccessToRepo` will always be false. I need to do some promise-resolving before I can add another filter, but I’m not really sure how to proceed… any ideas?

drogus: @backspace yes, I think that it's the only easy way to do it.

One interesting thing here, though, is that I realised my diagnosis of a problem is wrong. `store.filter` will actually re-evaluate records when their property change, but it will only happen for defined properties. So for example when you use filter on jobs, this will trigger a re-evaluation of a record: `job.set('state', 'created')`, but this won't: `job.set('foo', 'bar')`. This won't work here, because what you need here is to depend on a property from another object, but I just wanted to clarify what I found.

I plan to merge it soonish, probably today or tomorrow, so I hope that you won't have to wait long.

drogus: I merged the PR. I believe that you could achieve what you're after with:

```
store.filter('job', {}, 
  (job) => job.get('state') && job.get('repo.isCurrentUserACollaborator'),
  ['state', 'repo.isCurrentUserACollaborator']
)
```

drogus: `isCurrentUserACollaborator` is sort of a [hack that I recently added](https://github.com/travis-ci/travis-web/blob/d285d2f5a5de897f017e000054648473d52db75f/app/models/repo.js#L36-L43) just to make permissions check a computed property. In the future we will get it from the API as `@is_collaborator` property.

drogus: Actually, I think that this won't be enough in some cases, because I think that we don't update `permissions` array when a new repository is added. So it will be OK when you have access to a repo when the page loads, because the repo is already included in permissions. If you, however, keep the page opened and a new repo is added in the meantime, it won't be in permissions, so `isCurrentUserACollaborator` will be false. This will be fixed by `@is_collaborator` property, but we will also need to add this to Pusher.

backspace: But won’t `job.get('repo.isCurrentUserACollaborator')` not work because it’s an asynchronous relationship? That’s how it seems to be behaving for me…

drogus: It will work (or at least it should!), because the property [depends on permissions array](https://github.com/travis-ci/travis-web/blob/master/app/models/repo.js#L39) and `store.filter` will refresh the collection once any of the dependencies change. But please let me know if it doesn't work that way, I'm sure that it's something we can fix (because, in theory it should work).

backspace: hmm it seems you’re right! I don’t know why it wasn’t working yesterday 😳 

image

Thanks for this, I should be able to make progress on it this week.

drogus:
I don't see any changes to a query that fetches running jobs, what endpoint do you plan using? I think that unfortunately you may need to wait for running jobs endpoint on V3.

backspace:
hmm I was just using the same endpoint that is used on .com with the added permissions filter, but maybe I’m missing something?

app/services/job-state.js
```diff
@@ -18,7 +23,8 @@ export default Ember.Service.extend({
     const result = yield this.get('store').filter(
```
backspace: FUNNY that I can comment on this line but not the one above because it’s in the fold, but this one isn’t part of the PR either!

Another question for you, @drogus; why is it that `queued` is considered `running`? The labels of the sections are running and queued, seems confusing…???

drogus:
> hmm I was just using the same endpoint that is used on .com with the added permissions filter, but maybe I’m missing something?

Currently the entire implementation of the `/jobs` endpoint logic for .org is [here](https://github.com/travis-ci/travis-api/blob/master/lib/travis/services/find_jobs.rb) and as you can see it would just get all of the running jobs, not scoped by the current user. On .com we override most of the services to scope them with the authenticated user. So in essence, if there're less then 250 jobs running at any moment on .org, you will get proper results with your implementation, but if there're more jobs running, your jobs might not get on the list.

app/services/job-state.js
drogus: Good question! I would say that `queued` state shouldn't be there. I tracked it down to [this commit](https://github.com/travis-ci/travis-web/commit/c0917d2), but we will have to confirm if this was done by mistake or is it a deliberate change for some reason that we're not aware of.

drogus:
I created a [PR on API to bring `/jobs` endpoint there](https://github.com/travis-ci/travis-api/pull/563). I plan to deploy it on Monday, so you should be able to switch to V3 for running jobs. I also moved forward on [bringing jobs to V3](https://github.com/travis-ci/travis-web/pull/1068) and I hope to finish it early next week, so I think it would be best to apply your work on top of my changes, so you can just use V3 everywhere (otherwise it might complicate things too much).

app/services/job-state.js
drogus: I couldn't find any rationale behind that change, so I think that we should change `queued` to be only for queued jobs list.

backspace:
Sure, once your V3 work is merged, I’ll update this 😀

drogus:
I just deployed `/jobs` endpoint to both `.org` and `.com`, so API part is done :D

drogus:
I merged and shipped my work on moving jobs to API V3. It caused some conflicts here; let me know if you need any help with resolving them.

backspace:
Will do, thanks for the notice. The work on this PR hasn’t been that in-depth really, so I suspect it’ll be pretty easy. If you don’t have anything high-priority and wanted to continue with this branch, I’d be fine with handing it over, as I have a lot of other things happening at the moment 😀

clekstro:
@backspace I think this might be even easier after some of the merge work that's happened as well as #1547. Is this still something we're wanting to do?

backspace:
I think so, I haven’t heard otherwise… if you want to take over, @clekstro, go ahead, or I can continue on it at some point. 😀

joshk:
With .org on a limited life span, I think we can close this.

PR #1223, opened 2017-08-09
Speed up test suite

clekstro:
I've removed some of our timeouts completely and replaced another one that was causing an acceptance test to hang with a config value that we can set to `0` in tests.

Based on my local testing, this seems to speed things up by a little over 10%.

Before:
screen shot 2017-08-08 at 3 15 31 pm

After:
screen shot 2017-08-08 at 3 15 29 pm


backspace:
wow, nice!

PR #1227, opened 2017-08-09
Bump PR deploy packages

clekstro:


clekstro:
PR deploy was successful, so merging.

PR #1229, opened 2017-08-09
Improve switch accessibility

backspace:
Switches are currently invisible to keyboard navigation, as @webczat has kindly pointed out to us. I’m not sure why the switches were created as `a` but this converts them to `button`.

This includes a larger-scope change in that I removed `outline: none` on all `button` elements, as discussed [here](http://a11yproject.com/posts/never-remove-css-outlines/).

Still to come: Aria attributes.


backspace:
I’m confused by some of the Percy diffs because when I look at them locally in Firefox, they don’t look as they do on Percy! 😢

backspace:
ugh it’s because Percy uses Firefox 45; confirmed with Browserstack:

image

I’ve installed 45 locally and I can’t figure out where the extra space is coming from 😭

backspace:
Thanks to @lislis’s now-merged PR on this PR, the biggest diff on the hooks page is reduced; the hooks are now all one line. While it galls me to merge this knowing that the switch-buttons are unnecessarily indented on Firefox 45 (and possibly later versions, not sure when it was fixed), it’s comparatively little-used and I have been completely unable to find any CSS that makes the indent go away. I think it‘s better to get this merged despite that. (Which I will do after everything’s done running.)

app/styles/app/modules/switch.sass
```diff
@@ -43,7 +45,9 @@
 
   .on
     display: none
-    margin-left: 0
+    margin-left: -49%
```
BanzaiMan: @backspace This appears to have caused https://github.com/travis-ci/travis-ci/issues/8260. If I remove this, I get the correct display. Could you take a look?

backspace: hmm sadly removing that rule causes the switches to look wrong on the hooks page:

image

But I’ll find a workaround, the C of CSS is a perpetual problem 😆

BanzaiMan: CASCADE ALL THE THINGS!!! 🌊

PR #1231, opened 2017-08-10
Fix logging truncation

clekstro:
As part of changes related to the linting rules, the logging behavior was somehow impacted. This resulted in logs being cut off prematurely, but no error was being thrown (consistently cut off after about 470 lines).

### DIFF
--- a/app/utils/log.js
+++ b/app/utils/log.js
@@ -2,7 +2,7 @@
 
 import ansiparse from 'npm:ansiparse';
 
-var Log = function () {
+let Log = function () {
   this.autoCloseFold = true;
   this.listeners = [];
   this.renderer = new Log.Renderer;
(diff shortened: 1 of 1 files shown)


app/utils/log.js
```diff
@@ -240,7 +238,7 @@ Log.Part.prototype = Log.extend(new Log.Node, {
       }
     }
     if (!(slice >= this.slices.length - 1)) {
-      return setTimeout((() => { _this.process(slice + 1, num); }, Log.TIMEOUT));
+      return setTimeout(() => _this.process(slice + 1, num), Log.TIMEOUT);
```
clekstro: This was the culprit.

backspace:
I did a little more-granular investigation and found that keeping only this line from this PR fixed the error:

````
return setTimeout(() => this.process(slice + 1, num), Log.TIMEOUT);
```

So if you wanted to preserve the other changes, you could, but maybe it’s no big NBD

clekstro:
@backspace yep, see my review from before.

backspace:
oh oops, I missed that, sorry!

PR #1232, opened 2017-08-11
Remove unused route and page object

backspace:
I’ve come across these while working on #1229.


lislis:
Oh wow, I think that's a remnant from when each setting had its own route

PR #1238, opened 2017-08-11
Change test randomisation to not run with Percy

backspace:
I did manage to fix a lot of the spurious visual diffs by [eliminating some test bleed-through](https://github.com/travis-ci/travis-web/pull/1140/files) (and we got some [welcome help for others](https://github.com/travis-ci/travis-web/pull/1147)), but we still have one or two on each build since then. I don’t have time at the moment to look deeply enough to remove the rest of the bleed-though, so instead I’m suggesting that we stop randomising tests when we’ll be producing diffs with Percy.

I don’t love it, but the noise from the false diffs is grating and I don’t know if/when I’ll be able to address the cause.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,7 +46,7 @@ install:
 before_script:
   - if [ "$EMBER_VERSION" = 'default' ] && [ $TRAVIS_PULL_REQUEST = 'false' ]; then echo "Enabling Percy on push with default Ember" && export PERCY_ENABLE=1; fi
   - echo $PERCY_ENABLE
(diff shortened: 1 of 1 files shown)


backspace:
I’m going to merge this for now, since it’s easily undone. Since it was you who turned on randomisation, @clekstro, what do you think of this as a workaround until we have time to eliminate the bleed-through?

clekstro:
@backspace I definitely feel the pain of spurious test failures, so I'm fine with this being an interim solution. These things have a way of getting lost in the shuffle though, so it'd probably be good to create a new issue to address this and tackle it within the next several weeks.

PR #1240, opened 2017-08-14
Remove status link in test environment

backspace:
This should reduce false negatives in Percy.


backspace:
Here’s a diff from Percy showing that the green status dot is something that changed between builds:

image

There’s nothing tested about this code and probably doesn’t need to be, so this PR should just hide the status check entirely in the test environment.

backspace:
I was wrong that nothing was tested, but IMO, the assertions weren’t of significant value, since the only thing that determines whether the status link shows in the footer is the presence of the API URL, which the tests aren’t exercising. There’s a component test that does it, which seems sufficient.

PR #1241, opened 2017-08-14
Add Coveralls for coverage reporting

backspace:


backspace:
meh I don’t know what the problem with this is, probably some configuration I need re Babel, but I’m giving up for now.

PR #1242, opened 2017-08-14
Restore live favicon-as-status

backspace:
This broke with some refactoring. As of this commit, the job state is indicated for a favicon and backed up with a test, but I haven’t gotten it working for builds yet.

### DIFF
--- a/app/components/build-layout.js
+++ b/app/components/build-layout.js
@@ -19,12 +19,6 @@ export default Ember.Component.extend({
     }
   },
 
-  buildStateDidChange: Ember.observer('build.state', function () {
-    if (this.get('sendFaviconStateChanges')) {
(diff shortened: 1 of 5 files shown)


backspace:
This may be sufficient to restore the desired functionality but I’m thinking about changing it to use `ember-cli-head` and removing the use of observers. It may be possible to set the job-or-build-to-watch similarly to the way `sendFaviconStateChanges` is used now.

backspace:
I’m giving up on the refactoring for now due to insufficient time. I’ve created an issue elsewhere to track this and perhaps I can return to it.

clekstro:
@backspace Thanks for tracking this down! That was...surprising.

backspace:
hey yeah the codebase is full of surprises! 😬

PR #1243, opened 2017-08-15
Address svgjar warnings

lislis:
ids, and viewBox


backspace:
This makes me wonder whether we should have the team page as a Percy screenshot? 🤔

backspace:
BUT thanks for addressing this, I kept meaning to and then failing hehe

lislis:
That makes sense, I can add that here as well

lislis:
I fixed the one regression and added snapshots of the about, plans and logo page

backspace:
Excellent!

PR #1244, opened 2017-08-16
remove a thing

lislis:


backspace:
Ah excellent, thanks so much.

PR #1246, opened 2017-08-17
Fix switch alignment

backspace:
I’m also going to add a screenshot and proper tests for feature flags.

I currently can’t run tests locally because all sorts of things that should be globally present like `signInUser` and `server` aren’t available 😢 

image


backspace:
For reference, this is a fix for the bug reported [here](https://github.com/travis-ci/travis-ci/issues/8260).

selection_002

I introduced the visual regression in #1229.

PR #1248, opened 2017-08-21
Sort starred repos in the same order other repos are sorted

drogus:
Before this PR starred repos were not explicitly sorted, which could
be confusing, because the list of all repositories is sorted by current
build - latest builds are shown first.


clekstro:
@drogus 👍  Also, thanks for extracting the sort logic. Makes reading through this code path much easier.

PR #1256, opened 2017-08-24
Refactor trigger build logic to use derived states

clekstro:
This extracts out the ember-concurrency tasks into separate parent/child tasks and uses
`timeout` instead of Ember.run.later.


clekstro:
Need to double-check that this doesn't affect the UI transitions from modal to requests page. If it behaves differently, I'll wrap the closing of the modal within the parent task.

clekstro:
This looks good on the PR envs. Merging.

PR #1257, opened 2017-08-24
Show error message when viewing invalid jobs/builds

clekstro:
This should address https://github.com/travis-pro/team-teal/issues/2044. 

I basically piggy-backed on what we currently do in the case of a non-existent repo, which shows a message to the user whenever a user requests a build or job that doesn't belong to the owner/repo as the URL would claim.

There are some tests currently failing because I'm throwing an error because some of our test data appears to fail the validation (url === job.repo.slug/build.repo.slug), but I'll tackle these soon.


clekstro:
This is basically ready, assuming we're all on the same page regarding UX. Current implementation is somewhat bare-bones, but we can add additional help text to maybe make it friendlier.

clekstro:
I've added some helpful text which we can tweak at a later date if we'd like. Going to merge this.

PR #1260, opened 2017-08-25
use repo.defaultBranch instead of hardcoded master

lislis:


backspace:
I appreciate this, as a non-`master` type on my own repositories, so thanks!

I notice that you removed the loading indicator, so when I tried it out on the PR deployment, there was a little delay before the branch name popped in. Did it not work when the loading indicator showed? Not a big deal, though.

Nice cleanup too!

lislis:
@backspace ah true, I'll add the loading indicator back in

PR #1261, opened 2017-08-25
handle account not found on profile page

lislis:


app/components/no-account.js
```diff
@@ -0,0 +1,4 @@
+import Ember from 'ember';
+
+export default Ember.Component.extend({
+});
```
backspace: I think you could safely delete this file; like with a lot of things in Ember (routes, controllers, for instance), a basic implementation will be created if one isn’t given. I think the default would be this exact file!

lislis:
@backspace I haven't tried it out but I wasn't sure if throwing an error  in the model would result in the `-error` template being displayed in this location. The rendering/ outlet logic on the profile page is not as straight forward as in other places

backspace:
Can I not reply to your reply? CONFUSING

hmm yes I forgot about the nesting situation, maybe it would indeed render in the wrong place. So never mind then, thanks for the reminder!

PR #1263, opened 2017-08-28
Add docker configurations for nightly quay releases

clekstro:
This isn't needed for master/production builds, but referencing a non-master file when on a different branch is tedious. As crons are currently limited to a single branch and we're wanting to avoid having a conditional-littered build configuration, committing them here seems like a good compromise. Also handy in case you'd like to just work with Docker right out of the gate.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,77 @@
+FROM ruby:2.3
+
+LABEL maintainer Travis CI GmbH <support+travis-app-docker-images@travis-ci.com>
+
+RUN groupadd --gid 1000 node \
+  && useradd --uid 1000 --gid node --shell /bin/bash --create-home node
+
+# gpg keys listed at https://github.com/nodejs/node#release-team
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -0,0 +1,77 @@
...
+RUN npm install -g bower
+
+RUN npm install --quiet
+RUN bower install --allow-root
+
+COPY . /usr/src/app
+
+RUN ./node_modules/.bin/ember build --environment=production
+
+CMD bundle exec je puma -I lib -p ${PORT:-4000} -t ${PUMA_MIN_THREADS:-8}:${PUMA_MAX_THREADS:-12} -w ${PUMA_WORKERS:-2} --preload waiter/config.ru
```
mohitmun: @clekstro 
When I try running docker image, it gives this error
```
bundler: command not found: je
```
Am I missing out anything here?

PR #1267, opened 2017-08-28
Don't prefix pusher channels

drogus:
We don't need to prefix pusher channels manually in travis-web, because
we either get pusher channels from the API (which are already prefixed
where needed) or we subscribe to public repositories (and then we don't
need to prefix). This commit removes prefixing logic and instead
subscribes to whatever we get from the API.


drogus:
This should wait for: https://github.com/travis-ci/travis-live/pull/16

PR #1269, opened 2017-08-29
Make travis-web work with public repositories on pro version

drogus:
We're preparing for the merge and one of the requirements is to be able
to view public repositories and owners profiles. This commit enables
such a possibility.


drogus:
I requested a review here, but I realised I still want to change a few things. I'll request again when I'm done (probably today).

app/mixins/scroll-reset.js
```diff
@@ -2,7 +2,7 @@ import Ember from 'ember';
 
 export default Ember.Mixin.create({
   beforeModel: function () {
-    this._super(...arguments);
     window.scrollTo(0, 0);
+    return this._super(...arguments);
```
lislis: Does this make a difference perception-wise?

drogus: You mean for users? This is the bug that I was talking about the other day - if we don't return the result of the `_super` call, we may miss something that one of the routes did. For example if the `_super` call returns a rejected promise, if we don't return it here, it will get ignored. I'm not sure which places could be affected by this, but not returning here could lead to a confusing behaviour (in fact it took me quite some debugging time to figure out why auth doesn't work as it should for some routes)

app/components/repository-layout.js
```diff
@@ -12,12 +12,12 @@ export default Ember.Component.extend({
...
   @computed('repo.slug', 'repo.defaultBranch.name')
   statusImageUrl(slug, branchName) {
-    return this.get('statusImages').imageUrl(slug, branchName);
+    return this.get('statusImages').imageUrl(this.get('repo'), branchName);
   },
 
   @computed('repo.slug')
   urlGithub(slug) {
-    return this.get('externalLinks').githubRepo(slug);
+    return this.get('externalLinks').githubRepo(this.get('repo'));
```
backspace: These could both have `repo` as their first dependent keys now, instead of `slug`, right?

app/components/status-images.js
```diff
@@ -17,31 +17,35 @@ export default Ember.Component.extend({
...
   },
 
-  @computed('format', 'repo.slug', 'branch', 'repo.defaultBranch')
-  statusString(format, slug, branch, defaultBranch) {
-    const imageFormat = format || this.get('formats.firstObject');
-    const gitBranch = branch || defaultBranch.name;
-    return this.formatStatusImage(imageFormat, slug, gitBranch);
+  @computed('format', 'repo.slug', 'branch', 'repo.defaultBranch.name')
+  statusString(format, slug, branch, defaultBranchName) {
+    const repo = this.get('repo');
```
backspace: samesies here, might as well just switch out `slug` for `repo`. 😀

tests/acceptance/owner/not-found-test.js
```diff
@@ -0,0 +1,51 @@
...
+    adapterException = Ember.Test.adapter.exception;
+    loggerError = Ember.Logger.error;
+    Ember.Test.adapter.exception = () => null;
+    Ember.Logger.error = () => null;
+  },
+
+  afterEach() {
+    Ember.Test.adapter.exception = adapterException;
+    Ember.Logger.error = loggerError;
+  }
```
backspace: wow, this is quite the technique! I’ll try to remember this, seems useful.

tests/acceptance/owner/not-found-test.js
```diff
@@ -0,0 +1,51 @@
...
+test('visiting /non-existent-owner shows error message when unauthenticated', function (assert) {
+  withFeature('proVersion');
+  nonExistentOwnerPage.visit();
+
+  andThen(function () {
+    percySnapshot(assert);
+    assert.equal(currentURL(), '/non-existent-owner');
+    assert.ok(nonExistentOwnerPage.showsBarricadeIllustration, 'Shows image for aesthetics');
+    assert.equal(nonExistentOwnerPage.errorMessage, 'We couldn\'t find the owner non-existent-owner', 'Shows message that repository was not found');
+    assert.ok(nonExistentOwnerPage.errorMessageProUnauthenticated, 'shows .com authenticated message');
```
backspace: Hopefully this isn’t too nitpicky, but I think the [page object component concept](http://ember-cli-page-object.js.org/docs/v1.8.x/components#default-attributes) could help these be a bit more streamlined. Then these assertions could look like this:

```
assert.ok(nonExistentOwnerPage.barricade.isVisible, …
assert.equal(nonExistentOwnerPage.errorMessage, …
assert.ok(nonExistentOwnerPage.proErrorMessage.isVisible, …
```

That last assertion in the previous test could then be:

```
assert.ok(nonExistentOwnerPage.comErrorMessage.isHidden, …
```

What they call components come with a lot of built-in properties that work in typical situations.

tests/pages/owner/non-existent.js
```diff
@@ -0,0 +1,15 @@
...
+  text,
+  isHidden
+} from 'ember-cli-page-object';
+
+export default create({
+  visit: visitable('/non-existent-owner'),
+  showsBarricadeIllustration: contains('svg', { scope: '.page-graphic' }),
+  errorMessage: text('.missing-notice .page-title'),
+  errorMessageProisHidden: isHidden('.missing-notice p'),
+  errorMessageProUnauthenticated: contains('.missing-notice p')
```
backspace: ahhhh now that I look at the actual page object, my previous commentary on the acceptance test isn’t quite adequate, because I didn’t understand the structure. This could look differently:

```
errorMessage: {
  scope: '.missing-notice',
  text: text('.page-title'),
  comMessage: {
    scope: 'p'
  }
}
```

Then you get `errorMessage.text`, `errorMessage.comMessage.isHidden` and `….isVisible`. What do you think?

app/components/repository-layout.js
drogus: I actually want to observe `repo.slug`, because this computed property should recompute when slug changes (otherwise it would recompute only when repo changes to another repo). But that got me thinking that this code is not really readable. I pass the entire repo, because I need `slug` and `private` properties (and then it means I also need to depend on `private`. So to make this less confusing I'll change it to:

```js
@computed('repo.slug', 'repo.private')
   urlGithub(slug, private) {
    return this.get('externalLinks').githubRepo(slug, private);
  }
```

(this is why I hate the fact that Ember.js requires listing dependencies, btw ;) ).

tests/acceptance/owner/not-found-test.js
drogus: I actually just copied this from `repo/not-found-test`, but although it is useful I keep wondering if that's the proper way to do it. Maybe I'll investigate it later.

app/components/repository-layout.js
backspace: oh yes, I forgot about that reason about what to observe. Well, nice change, thanks.

tests/pages/owner/non-existent.js
drogus: That's a good point. I want to get it out today, though, so I'll address this (and `repo-not-found` test where I copied it from) in a separate PR later this week.

bnferguson:
Working on building Travis Enterprise 2.2 (currently building off of the `master` branch), and this PR introduced a bug with log contents. Around here: https://github.com/travis-ci/travis-web/blob/master/app/models/job.js#L173-L177 

The issue being that we configure `Travis.config.pusher.secure` as `true` globally for Enterprise, so `travis-logs` tries to broadcast updates on the `private-job-#{id}` channel for everything (https://github.com/travis-ci/travis-logs/blob/master/lib/travis/logs/pusher.rb#L20-L22) while `travis-web` is listening either on `job-#{id}` for "public" repositories or `private-job-#{id}` for "private" ones (quotes added for GitHub Enterprise since everything is a bit private).

Oddly, other pieces of pusher based interface (which also broadcast on `private-` in Enterprise) continue to work regardless, I'm guessing from this work: https://github.com/travis-ci/travis-web/pull/601. So it seems the way we handle things is different for this end point than others?

I know we also have some of these ideas around The Merge and how Travis CI Enterprise should behave. I believe @svenfuchs and I had a conversation about the immediate handling of this, I think we opted for everything to be private. But ideally would take our cues from the GitHub instance we're connected to later. Eg If it's in private mode (have to log in to see anything) so are we, if it's in public mode (anonymous users can see repos of public projects) so are we (and they can read logs). When we get there, some changes will be needed to logs and likely web. 

For the time being - I'm happy to just lock this endpoint down as private with the rest so we can release 2.2 in Jan as this is a current blocker.

PR #1275, opened 2017-08-29
Subscribe to public repos on pro too

drogus:
When a user opens a repo that they aren't a collaborator of, we will
subscribe them to a public repo channel, so they still get updates for
the repo. This was done only on .org, but now that we plan to merge the
platforms it should work on .com version too.

The code needed to be changed a little bit, because now we need to also
care about private repositories and also we don't want to subscribe to
repositories that are already covered on user channel (ie. if the
current user is a collaborator, they would get an update through a user
channel anyway).

### DIFF
--- a/app/routes/application.js
+++ b/app/routes/application.js
@@ -30,29 +30,47 @@ export default TravisRoute.extend(BuildFaviconMixin, KeyboardShortcuts, {
   },
 
   activate() {
-    let repos;
-    if (!this.get('features.proVersion')) {
-      repos = this.get('store').peekAll('repo');
+    this.setupRepoSubscriptions();
+  },
+
+  // We send pusher updates through user channels now and this means that if a
+  // user is not a collaborator of a repo or a user is not signed in, we need to
+  // use repo channels for updates for each repo. This method ensures that a
+  // visitor is subscribed to all of the public repos in the store as long as
+  // they're not a collaborator. It also sets up an observer to subscribe to any
+  // new repo that enters the store.
+  setupRepoSubscriptions() {
+    this.get('store').filter('repo', null,
+      (repo) => !repo.get('private') && !repo.get('isCurrentUserACollaborator'),
+      ['private', 'isCurrentUserACollaborator']
+    ).then((repos) => {
       repos.forEach(repo => this.subscribeToRepo(repo));
-      return repos.addArrayObserver(this, {
+      repos.addArrayObserver(this, {
         willChange: 'reposWillChange',
         didChange: 'reposDidChange'
       });
-    }
+    });
   },
 
-  reposWillChange() {},
+  reposWillChange(array, start, removedCount, addedCount) {
+    let removedRepos = array.slice(start, start + removedCount);
+    return removedRepos.forEach(repo => this.unsubscribeFromRepo(repo));
+  },
 
   reposDidChange(array, start, removedCount, addedCount) {
-    let addedRepos;
-    addedRepos = array.slice(start, start + addedCount);
+    let addedRepos = array.slice(start, start + addedCount);
     return addedRepos.forEach(repo => this.subscribeToRepo(repo));
+  },
+
+  unsubscribeFromRepo: function (repo) {
+    if (this.pusher) {
+      this.pusher.unsubscribe(`repo-${repo.get('id')}`);
+    }
   },
 
   subscribeToRepo: function (repo) {
     if (this.pusher) {
-      const channel = `repo-${repo.get('id')}`;
-      return this.pusher.subscribe(channel);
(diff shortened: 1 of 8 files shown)


app/utils/pusher.js
```diff
@@ -63,29 +62,19 @@ TravisPusher.prototype.unsubscribeAll = function (channels) {
...
-  }
-  if (!((ref = this.pusher) != null ? ref.channel(channel) : void 0)) {
-    return this.pusher.subscribe(channel).bind_all((function (_this) {
-      return function (event, data) {
-        return _this.receive(event, data);
-      };
-    })(this));
+TravisPusher.prototype.subscribe = function (channelName) {
+  if (channelName && this.pusher && !this.pusher.channel(channelName)) {
+    this.active_channels.push(channelName);
```
clekstro: Is the order important here? i.e. does it make sense to add to the active channels only after the pusher subscription was successful?

app/utils/pusher.js
```diff
@@ -63,29 +62,19 @@ TravisPusher.prototype.unsubscribeAll = function (channels) {
...
-    return;
-  }
-  //eslint-disable-next-line
-  console.log("unsubscribing from " + channel);
-  if ((ref = this.pusher) != null ? ref.channel(channel) : void 0) {
-    return this.pusher.unsubscribe(channel);
+TravisPusher.prototype.unsubscribe = function (channelName) {
+  if (channelName && this.pusher) {
+    this.active_channels.removeObject(channelName);
+    return this.pusher.unsubscribe(channelName);
```
clekstro: Same question here.

app/routes/application.js
```diff
@@ -30,29 +30,47 @@ export default TravisRoute.extend(BuildFaviconMixin, KeyboardShortcuts, {
...
+  },
+
+  // We send pusher updates through user channels now and this means that if a
+  // user is not a collaborator of a repo or a user is not signed in, we need to
+  // use repo channels for updates for each repo. This method ensures that a
+  // visitor is subscribed to all of the public repos in the store as long as
+  // they're not a collaborator. It also sets up an observer to subscribe to any
+  // new repo that enters the store.
+  setupRepoSubscriptions() {
+    this.get('store').filter('repo', null,
```
clekstro: So this doesn't make a request itself, because `null` is passed, but the store is still being observed as other components/routes fetch this data, correct? Just want to make sure I'm understanding the flow and that not making the request was intentional, given this is happening in the ApplicationRoute.

tests/acceptance/repo/show-public-repo-test.js
```diff
@@ -0,0 +1,61 @@
...
+
+  page.visit({ organization: 'musterfrau', repo: 'a-repo' });
+
+  andThen(() => {
+    let subscribed = this.application.resolveRegistration('pusher:main').active_channels.includes(`repo-${repo.id}`);
+    assert.ok(subscribed, 'user is subscribed to a repo channel');
+  });
+});
+
+test('viewing public repo as a signed collaborator does not trigger subscription', function (assert) {
```
clekstro: signed in?

tests/acceptance/repo/show-public-repo-test.js
```diff
@@ -0,0 +1,61 @@
...
+
+  page.visit({ organization: 'musterfrau', repo: 'a-repo' });
+
+  andThen(() => {
+    let subscribed = this.application.resolveRegistration('pusher:main').active_channels.includes(`repo-${repository.id}`);
+    assert.ok(!subscribed, 'user is not subscribed to a repo channel');
+  });
+});
+
+test('viewing public repo as a signed user triggers subscription', function (assert) {
```
clekstro: signed in?

app/utils/pusher.js
drogus: At the moment I just use it for tests, so it doesn't really matter I think. In production app we should rely on pusher as a source of truth. More explanation in commit message here: https://github.com/travis-ci/travis-web/pull/1275/commits/ecf87603f06057cb200222f45f1ee0754dd340d1

app/routes/application.js
drogus: Yup, exactly, I don't want to fetch any new repos, just get anything that's loaded already and react to any changes.

tests/acceptance/repo/show-public-repo-test.js
drogus: :+1:

app/routes/application.js
clekstro: 👍

app/utils/pusher.js
clekstro: Sorry, I read that initially and then when I started the review somehow lost track. Makes sense 👍

PR #1279, opened 2017-08-31
Tags!

lislis:
screen shot 2017-08-31 at 16 24 22
screen shot 2017-08-31 at 16 37 21

Placing a tag icon where a push == tag, linking to GH tag from the build header

### DIFF
--- a/app/models/build.js
+++ b/app/models/build.js
@@ -61,7 +61,7 @@ export default Model.extend(DurationCalculations, {
 
   @computed('tag')
   isTag(tag) {
-    return tag !== null;
+    return (tag && tag.name);
   },
 
   @computed('state')
--- a/app/models/build.js
+++ b/app/models/build.js
@@ -61,7 +61,7 @@ export default Model.extend(DurationCalculations, {
 
   @computed('tag')
   isTag(tag) {
-    return Ember.isEmpty(tag);
+    return tag !== null;
   },
 
   @computed('state')
--- a/app/models/build.js
+++ b/app/models/build.js
@@ -24,6 +24,7 @@ export default Model.extend(DurationCalculations, {
   finishedAt: attr('string'),
   pullRequestNumber: attr('number'),
   pullRequestTitle: attr('string'),
+  tag: attr(),
   eventType: attr('string'),
   _config: attr(),
 
@@ -56,6 +57,11 @@ export default Model.extend(DurationCalculations, {
   @computed('jobs.[]')
   isMatrix(jobs) {
     return jobs.get('length') > 1;
+  },
+
+  @computed('tag')
+  isTag(tag) {
+    return Ember.isEmpty(tag);
   },
 
   @computed('state')
--- a/app/templates/components/build-header.hbs
+++ b/app/templates/components/build-header.hbs
@@ -50,6 +50,10 @@
           <span data-tooltip="View branch on GitHub" class="label-align inner-underline">Branch {{build.branchName}}</span>
           {{svg-jar 'icon-external-link' class="icon"}}
         {{/external-link-to}}
+
+        {{#if item.isTag}}
+          it's a tag
+        {{/if}}
       </li>
     </ul>
 
--- a/app/templates/components/build-header.hbs
+++ b/app/templates/components/build-header.hbs
@@ -47,7 +47,7 @@
       <li>
         {{#if build.isTag}}
           {{#external-link-to href=urlGitHubTag class="commit-branch-url"}}
-            {{svg-jar 'icon-branch' class="icon"}}
+            {{svg-jar 'icon-tag' class="icon"}}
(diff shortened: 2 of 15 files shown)


lislis:
still needs tests

tests/integration/components/build-header-test.js
```diff
@@ -75,6 +75,29 @@ test('render cron build', function (assert) {
   assert.ok(this.$().find('.build-title').text().match(/cron Just complete and utter joy/), 'displays cron before commit message');
 });
 
+
+test('render tag build', function (assert) {
+  let commit = {
+    subject: 'Just complete and utter joy',
```
backspace: :joy:

tests/integration/components/build-header-test.js
```diff
@@ -75,6 +75,29 @@ test('render cron build', function (assert) {
...
+    isTag: true
+  };
+
+  this.set('build', build);
+  this.render(hbs`{{build-header item=build commit=build.commit}}`);
+
+  assert.equal(this.$().find('.build-status svg title').text(), 'Tag');
+  assert.equal(this.$().find('.build-title .commit-branch').text().trim(), 'v1.0.0', 'displays tag name in title');
+  assert.ok(this.$().find('.commit-branch-url').text().match(/Tag v1.0.0/), 'displays link to tag on GH');
+  assert.ok(this.$().find('.commit-branch-url').attr('href').match(/releases\/tag\/v1.0.0/), 'url is correct');
```
backspace: This makes me wonder, are there any limitations on what characters tags can contain? [This page](https://git-scm.com/docs/git-check-ref-format) purportedly lists what’s okay in a tag name, is it possible that someone can generate a tag name that needs escaping when linking? 🤔

app/models/build.js
```diff
@@ -58,6 +59,11 @@ export default Model.extend(DurationCalculations, {
     return jobs.get('length') > 1;
   },
 
+  @computed('tag')
+  isTag(tag) {
+    return (tag && tag.name);
+  },
```
backspace: Not a big deal but this could use the [`notEmpty`](https://emberjs.com/api/ember/2.15/namespaces/Ember.computed/methods/notEmpty?anchor=notEmpty) computed property.

app/templates/components/build-header.hbs
```diff
@@ -45,11 +50,19 @@
...
+            <span data-tooltip="View tag on GitHub" class="label-align inner-underline">Tag {{build.tag.name}}</span>
+            {{svg-jar 'icon-external-link' class="icon"}}
+          {{/external-link-to}}
+        {{else}}
+          {{#external-link-to href=urlGitHubBranch class="commit-branch-url"}}
+            {{svg-jar 'icon-branch' class="icon"}}
+            <span data-tooltip="View branch on GitHub" class="label-align inner-underline">Branch {{build.branchName}}</span>
+            {{svg-jar 'icon-external-link' class="icon"}}
+          {{/external-link-to}}
+        {{/if}}
```
backspace: Seeing all these `if item.isTag`/`else`s spread about brings to mind a refactoring pattern, [Replace Conditional with Polymorphism](https://refactoring.guru/replace-conditional-with-polymorphism). I don’t know if it’s an exact match, but it could be a way to reduce some duplication. Maybe it’s not worth it at this point because there is only one conditional, not a series.

But as an example, this template could be:

```
{{#external-link-to href=item.gitHubHref class='commit-branch-url'}}
  {{svg-jar item.icon class='icon'}}
  <span data-tooltip='View {{item.type}} on GitHub' class=…>{{upcase (:laughing:) item.type}} {{item.displayName}}</span>
  {{svg-jar 'icon-external-link' class='icon'}}
{{/external-link-to}}
```

This is obviously not a realistic implementation, but you can see how it could reduce complexity in the templates. I start to think about this when I see the same conditional scattered around a codebase. But I’m just pointing this out as something you may not have heard about, it seems still under control here, just getting toward refactoring-itch territory for me 😀

Typing that made me wonder whether line 54 should have a `commit-tag-url` branch? But maybe not, if they’re sharing styles.

tests/integration/components/build-header-test.js
lislis: good point, I didn't think of that. But I guess GitHub would already make sure it's an url-able tag name?

app/templates/components/build-header.hbs
lislis: I also have the feeling that this could be done more elegantly. Eg having the request and tag icon conditional in and `event-icon` component. Also the polymorphism approach looks nice. 
So far this is all that tags are about, maybe it's fine?
I'll also leave `.commit-branch-url`, same styles and the test uses it, too.

tests/integration/components/build-header-test.js
backspace: yeah, seems like it! I have never made a tag in my whole life haha so I’m like 🤔

PR #1281, opened 2017-09-01
Switch to Ember Tooltips

backspace:
This is in progress.


backspace:
I approved the diffs because I think they’re an improvement! It looks to me that the existing tooltips were causing their elements to shift downward slightly.

PR #1289, opened 2017-09-05
Fix test warnings

clekstro:
This was obscuring the log output with errors about multiple Pretender instances.


drogus:
Looks good to me

PR #1297, opened 2017-09-09
customized and separate oss project numbers on landing page

ctrlaltjustine:
Addresses [design-tasks issue #68](https://github.com/travis-pro/design-tasks/issues/68)
To make our testing numbers more accurate, created new svgs and styles to make each number individual which will allow us to customize them when needed. Unfortunately this is static atm.


clekstro:
@ctrlaltjustine I'm not able to see the numbers in the PR env or locally. I'll have a look and see if I can track down the issue.

carlad:
Looks to be visible now :)
Before we merge I will update the number to the actual current number: 932977
I think Justin-e used the number from the dodgy cut and paste i did with the existing graphic (I only had a few numbers to cut and paste with...)

PR #1301, opened 2017-09-14
Use API V3 for logs endpoint

drogus:


drogus:
TODO:

* check error handling
* fix tests (and check if we have a test for an error)

clekstro:
@drogus I fixed the tests as I figured it made more sense to build off of this PR when working on my own issue to avoid re-requesting the same logs.

drogus:
@clekstro thanks! :D

drogus:
One thing that I just noticed is that I missed the `Authorization` header: https://github.com/travis-ci/travis-web/pull/1301/files#diff-0c1e6c749da5d33ef8ab52b874c40d6dR38

I would surely catch it in manual testing, but I'll think about catching such stuff in mirage as it should be doable.

drogus:
I think this is ready to go once the backend work is finished: https://github.com/travis-ci/travis-api/pull/561

PR #1303, opened 2017-09-14
Move twitter.svg into the correct directory

ur5us:
This asset is not used anywhere else apart from
app/templates/components/landing-default-page.hbs which expects to find the
image in /public/images/landing-page/.


backspace:
Thanks for fixing this! 😀

the-j0k3r:
and https://github.com/travis-ci/travis-web/blob/master/app/styles/app/layouts/error.scss#L71 was left in old directory :) should have updated the paths too when moving images.

PR #1307, opened 2017-09-19
Update to Ember Data 2.15

backspace:
I’ve skipped some tests with a [backtracking rerender error](https://github.com/emberjs/ember.js/issues/13948) but apart from that everything seems to be working now, so I’ll return to look at what’s going on there.


backspace:
This is working for me locally. See commit comments for details about some of the changes.

backspace:
There’s a rendering change in the dashboard snapshot, but it’s a byproduct of my having fixed the backtracking rerender bug, it’s not a Real Problem™ IMO.

PR #1309, opened 2017-09-19
Add npm lockfile

backspace:
I failed to include this in my Node 8 update PR 😱

It currently includes an unnecessary package update, I’m just making sure it works as expected when deployed to Heroku.


backspace:
Since it deployed successfully after I cleared the cache, I’ve stripped the extra commits and now it’s only the lockfile.

PR #1314, opened 2017-09-20
Run svgo against svg icons

clekstro:


clekstro:
@backspace @lislis any concerns with this? No visual diffs in percy, so I don't think there's a high risk of any resulting UI bugs.

lislis:
👍 looks like mostly reduced whitespace and removed`g` elements and attributes. Should be fine

backspace:
yeah, fine with me!

backspace:
er…

PR #1315, opened 2017-09-20
Fix fetching a single stage

drogus:
At the moment API only implements /build/:id/stages endpoint for
fetching stages info. We implemented findMany method in stage adapter in
order where we use the endpoint, but it fails when there's only one
stage in build. Ember Data won't fire findMany then, but instead it will
use findRecord, which by default would go to /stages/:id; API doesn't
support it.

This commit implements findRecord method in the stage adapter to use
/build/:id/stages endpoint, but return only one stage.


backspace:
excellent, thanks for figuring this out!

PR #1328, opened 2017-09-27
Repository filtering on profile page

clekstro:
This still needs some work, but is mostly functional.


BanzaiMan:
Hello. How can we help move this forward while Curtis is on leave? I think this is a real UX issue we would like to have fixed for our users.

BanzaiMan:
Rebasing seems to help, but the conflict remains with `package.json` and `package-lock.json`, which I am afraid is outside of my comfort zone.

config/environment.js
```diff
@@ -25,7 +25,7 @@ module.exports = function (environment) {
     },
 
     // defaults for running travis-web
-    apiEndpoint: 'https://api-staging.travis-ci.org',
+    apiEndpoint: 'https://api.travis-ci.org',
```
backspace: I wish I had a better way to accomplish this, but if your branch has `staging` in the name, you get `org-staging` and `com-staging` PR deployments too, which removes the need for this sort of thing.

drogus: This was actually only for testing locally, I just forgot to remove it. But thanks for the tip!

PR #1332, opened 2017-10-02
Adding Bianca's assets to the team page

ctrlaltjustine:
This PR addresses [design-tasks issue #73](https://github.com/travis-pro/design-tasks/issues/73)


backspace:
I found an extra image asset that I theorise was an aid in creating the hybrid flag. Everything renders the same after I delete it from the SVG, so I think it’s safely gone! 😀

PR #1341, opened 2017-10-06
Port app code over to new module syntax

clekstro:
* Apply `ember-modules-codemod` and a few manual fixes dealing with
conflicts with `ember-decorators`.

### DIFF
--- a/app/adapters/v3.js
+++ b/app/adapters/v3.js
@@ -1,3 +1,5 @@
+import { underscore } from '@ember/string';
+import { get } from '@ember/object';
 import Ember from 'ember';
 import config from 'travis/config/environment';
 import RESTAdapter from 'ember-data/adapters/rest';
@@ -55,7 +57,7 @@ export default RESTAdapter.extend({
 
   buildURL: function (modelName, id) {
     let url = [];
-    const host = Ember.get(this, 'host');
+    const host = get(this, 'host');
     const prefix = this.urlPrefix();
     const pathPrefix = this.pathPrefix(...arguments);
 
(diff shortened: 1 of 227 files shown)


app/adapters/v3.js
```diff
@@ -1,3 +1,5 @@
+import { underscore } from '@ember/string';
+import { get } from '@ember/object';
 import Ember from 'ember';
```
backspace: It looks to me that this is for `Ember.String.pluralize`, is it an oversight?

I searched the branch and many of the remaining imports are for `Ember.testing` which I assume is not possible to address yet, I don’t know too much about this… I also see `Ember.Handlebars.Utils`, `Ember.Inflector.inflector`, `Ember.merge`…

Is there an advantage to a partial conversion in this direction? Is it worth waiting until we can get rid of all our monolithic imports?

This has turned into a more meta comment than this single line hehe

clekstro: @backspace So, I was surprised to discover that you can't import `import { pluralize } from '@ember/string'`, likely because `pluralize` is actually on `DS.String` and not `Ember.String`. (How does this work lol)

I figured that it didn't hurt to move over in pieces, though I'm fine if we want to wait until everything can be removed. Your call.

backspace: oh yes, I guess it’s a “prototype extension”, makes sense. Well I don’t see a reason to not move, though I don’t know how I’ll remember to make this change hehe

drogus:
I'm going to merge this to not wait with it further as it will most likely cause conflicts on PRs, so the sooner the better.

@backspace regarding your comment, I guess there are no guarantees other than Ember.js release plan, but it's more than enough for me. And even if that ever needs to be changed in any way, I trust the community or the core team to prepare a tool to aid such a transition.

PR #1368, opened 2017-10-17
Update Ember projects to 2.16

backspace:
I removed ember-router-service-polyfill, as advised by a notification
while building.

This includes some simple line-switchings where the modules codemod
put things in a different order than the blueprints. I also replaced
testem.js with the blueprint version because our differences were
negligible, might as well reduce friction during updates.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,11 @@ cache:
     - node_modules
     - bower_components
 
+env:
+  global:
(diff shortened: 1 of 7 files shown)


backspace:
hmm the Percy diff is because somehow the team page was randomised in the snapshot, I thought I made that go away 😕

PR #1369, opened 2017-10-18
Move from jQuery timeago to npm’s timeago

backspace:
This has some test failures. I don’t know how attached to the
current wording people are. We’ll see…


backspace:
hmm I missed the live-updating part of the original OOPS

PR #1370, opened 2017-10-18
Remove deprecated alias

backspace:
This will be removed in Ember 2.17. I think it‘s the only one…
https://emberjs.com/deprecations/v2.x/#toc_controller-content-alias


drogus:
Looks good to me!

PR #1377, opened 2017-10-20
fix white-space in extra colored logs

lislis:
Turns 

screen shot 2017-10-20 at 11 43 45

into

screen shot 2017-10-20 at 11 43 50


backspace:
hah wow, nice catch!

Turbo87:
@backspace unfortunately this change is now causing all whitespace in logs to be collapsed which is a much larger issue than the colors above not working correctly. I'd suggest to revert this change and they try to fix the color issue some other way.

backspace:
eep okay thanks for letting us know 💞

PR #1381, opened 2017-10-23
Restore log whitespace style

backspace:
Thanks to @Turbo87 for letting us know about this here:
https://github.com/travis-ci/travis-web/pull/1377#issuecomment-338700704

### DIFF
--- a/app/styles/app/layouts/log.sass
+++ b/app/styles/app/layouts/log.sass
@@ -153,7 +153,6 @@
     padding: 0 15px 0 55px
     margin: 0
     min-height: 16px
-    white-space: normal
     &:hover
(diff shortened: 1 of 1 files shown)


Turbo87:
@backspace I'm a little surprised that Percy did not catch this regression. Is it not testing any output logs?

backspace:
It does, here’s the screenshot:

image

It’s difficult to exercise all possible variations in a test, but I think we should be able to adjust it so it covers this, thanks to your bug report. The logging code is a black hole of complexity 😬

Turbo87:
> The logging code is a black hole of complexity 😬

no doubt :D

thanks for addressing this so quickly! :)

PR #1382, opened 2017-10-23
Add whitespace to complex log acceptance test line

backspace:
This adds Percy detection for the regression addressed in #1381.


backspace:
meh I don’t understand, it’s not showing the extra whitespace:

image

backspace:
Here it is locally in Chrome:

image

And Firefox 54, the version Percy uses:

image

🤔 it’s like Percy ran against non-`master`, I’m perpetually confused

backspace:
I can’t explain it but #1383 has the expected Percy diff. ¯\\\_(ツ)\_/¯

PR #1395, opened 2017-11-07
Add signin route

backspace:
This is an experiment.

### DIFF
--- a/app/router.js
+++ b/app/router.js
@@ -47,7 +47,7 @@ Router.map(function () {
   this.route('first_sync');
   this.route('insufficient_oauth_permissions');
   this.route('auth');
-  this.route('signup');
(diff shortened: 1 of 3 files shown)


drogus:
So, the thing is that we already had signin route and it was removed, because we thought that it's not really used for anything and it looked like this: https://github.com/travis-ci/travis-web/commit/580324fe8c95229bed0652e1e2ec7766c312935f#diff-ec206cdb5f1d1f5541ac71d60e19fdb2 (minus the comment of course, I just took it from this commit: ). I think that with this version you wouldn't have to change `auth.js` actually.

Also, sorry, I figured out what's happening last week, but with all the things happening lately I somehow forgot to update our internal issue and to commit anything when I moved on to other things :(

PR #1397, opened 2017-11-14
Don't update records with stale data

drogus:
From the commit message:

```
Data updates doesn't arrive to the browser in any specific order.
Because of that it's possible that data in an update delivered by
Pusher is older than data that we have available locally.

This commit overrides `store._pushInternalModel` function in order to
guard against such situations by checking the `updatedAt` attribute. If
a record that is being pushed to the store is older than a record that
we already have, we will just return the existing record.
```

### DIFF
--- a/app/models/build.js
+++ b/app/models/build.js
@@ -30,6 +30,7 @@ export default Model.extend(DurationCalculations, {
   tag: attr(),
   eventType: attr('string'),
   _config: attr(),
+  updatedAt: attr('date'),
 
   repo: belongsTo('repo'),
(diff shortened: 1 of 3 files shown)


drogus:
Just for the record - I'm planning to propose a change in Ember Data to allow registering some kind of a hook for this reason, but it can take a long time before it's get accepted (and it's not certain if it will be accepted at all).

So as much as I know that each piece of code like that increases our chances of getting into issues when updating Ember Data, I think it's fine as long as it's tested and we can catch regressions (and I think the tests I added should be enough for this)

PR #1403, opened 2017-11-17
Fix `offset=0` on the profile page

drogus:
There was a bug in my custom filtering code that was improperly caching collections - we were caching by calculating id based on filter parameters, but not taking into account nested query params. This PR fixes it and then also adds `offset: 0` to `account/repositories` controller to avoid adding it to query params by default.

### DIFF
--- a/app/utils/live-paginated-collection.js
+++ b/app/utils/live-paginated-collection.js
@@ -106,4 +106,8 @@ LivePaginatedCollection.reopenClass({
   }
 });
 
+LivePaginatedCollection.toString = function () {
+  return 'LivePaginatedCollection';
+};
+
 export default LivePaginatedCollection;
--- a/app/components/repository-status-toggle.js
+++ b/app/components/repository-status-toggle.js
@@ -25,7 +25,9 @@ export default Component.extend({
 
   @computed('repository.permissions')
   admin(permissions) {
-    return permissions.admin;
+    if (permissions) {
+      return permissions.admin;
+    }
   },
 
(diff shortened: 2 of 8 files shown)


drogus:
An interesting thing to note here that it was working without `offset: 0` in the controller, because then the offset was a string and the code was not returning live paginated collection, but rather regular one, which is not cached nor autoupdated.

backspace:
wow, thanks for figuring this out. I didn’t think about it being a pagination thing. I’m glad we can get rid of the annoying `offset=0`!

app/utils/live-paginated-collection.js
```diff
@@ -106,4 +106,8 @@ LivePaginatedCollection.reopenClass({
   }
 });
 
+LivePaginatedCollection.toString = function () {
+  return 'LivePaginatedCollection';
+};
+
```
backspace: Is this for debugging help?

drogus: For tests actually, but I guess it's good to have anyway (and maybe I'll start adding it more?). With this you get something like `<LivePaginationCollection:ember1213>` instead of `<subclass of (ArrayProxy):ember1213` when you do `toString()` on an instance.

drogus: I'm checking it here: https://github.com/travis-ci/travis-web/pull/1403/files#diff-8916b1cc56744b82d8052735c3777361R242

backspace: oh yes, makes sense, I missed that 😬

PR #1404, opened 2017-11-17
Fix auto sign out

drogus:
from the commit message:

```
When a Travis token expires and a user tries to open the app, we will
sign them out. The problem is that due to changes in e71862c we're no
longer running some of the sign out steps when the sign out is initiated
from the `auth` service. This commit gets back to calling the
`afterSignOut` every time a `signOut` is initiated.
```


drogus:
I'll add some tests to it to not let it happen again, but I wanted to open a PR already.

backspace:
Thanks for looking into this!

PR #1405, opened 2017-11-17
Fix background colour bleeding

backspace:
I’d prefer to get rid of floats altogether, but that’s for
another century.


backspace:
haha well _that’s_ entirely wrong

PR #1411, opened 2017-11-29
Add filtering for triggerable branches

backspace:
It doesn’t make sense to trigger a build on a branch that doesn’t exist on GitHub.


backspace:
This includes some Percy diffs but I approved them because I don’t think they’re relevant to what’s being tested.

PR #1413, opened 2017-12-01
Remove included build.jobs from PR query

backspace:
It’s hard to say what this will do, we’ll see!


backspace:
I’m merging this for deployment on Monday! 😳

PR #1414, opened 2017-12-04
Use Menlo for monospace font stack

jarrodldavis:
The current monospace font stack, `Monaco, monospace`, causes issues on macOS since Monaco does not have a Bold variant, only Regular.

In Chrome, any text intended to be bolded isn't, thus bold and non-bold text have the same rendered font weight:

chrome-before

Safari tries to create a bold variant on-the-fly, but this breaks the intended monospace behavior:

safari-before

Menlo, on the other hand, does have a bold variant, and looks better in Chrome and Safari:

chrome-after

safari-after

This will also affect Firefox on macOS, though it renders both typefaces correctly. Other platforms should not be affected because neither typeface should be included by default (due to both being created by Apple).

### DIFF
--- a/app/styles/app/base.sass
+++ b/app/styles/app/base.sass
@@ -91,7 +91,7 @@ span.emoji-sizer
   @include resetul
 
 .monospace
-  font-family: Monaco, monospace
+  font-family: Menlo, monospace
   font-size: 13px
   line-height: 1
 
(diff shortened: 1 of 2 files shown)


backspace:
Hey, thanks for letting us know about this! I was just noticing last week that Firefox had boldness in the logs that wasn’t present in Chrome. I can’t just merge this without some internal discussion, but I’ll get that started and get back to you.

backspace:
Okay, we decided to switch to a typeface that should appear the same across platforms. Thanks for the nudge on this!

PR #1423, opened 2017-12-07
Fix log streaming in Enterprise, always use private channels. 

bnferguson:
As written up here: https://github.com/travis-ci/travis-web/pull/1269#issuecomment-349621548 

> The issue being that we configure Travis.config.pusher.secure as true globally for Enterprise, so travis-logs tries to broadcast updates on the private-job-#{id} channel for everything (https://github.com/travis-ci/travis-logs/blob/master/lib/travis/logs/pusher.rb#L20-L22) while travis-web is listening either on job-#{id} for "public" repositories or private-job-#{id} for "private" ones (quotes added for GitHub Enterprise since everything is a bit private).
>
> Oddly, other pieces of pusher based interface (which also broadcast on private- in Enterprise) continue to work regardless, I'm guessing from this work: #601. So it seems the way we handle things is different for this end point than others?
>
> I know we also have some of these ideas around The Merge and how Travis CI Enterprise should behave. I believe @svenfuchs and I had a conversation about the immediate handling of this, I think we opted for everything to be private. But ideally would take our cues from the GitHub instance we're connected to later. Eg If it's in private mode (have to log in to see anything) so are we, if it's in public mode (anonymous users can see repos of public projects) so are we (and they can read logs). When we get there, some changes will be needed to logs and likely web.
>
>For the time being - I'm happy to just lock this endpoint down as private with the rest so we can release 2.2 in Jan as this is a current blocker.

----------------

I attempted to address the issue in this PR but upon testing in Enterprise I'm not entirely sure `this.get('config.enterprise')` returns what I think or that `config.enterprise` is even available in this spot. Also it seems we bounce between `this.get('config.enterprise')`, `this.get('features.enterprise')` and `features.enterpriseVersion` (in templates). So I'm not entirely sure which makes the most sense.

What am I missing here? Any help is hugely appreciated. <3

### DIFF
--- a/app/models/job.js
+++ b/app/models/job.js
@@ -173,7 +173,7 @@ export default Model.extend(DurationCalculations, DurationAttributes, {
   @computed('repo.private', 'id')
   channelName(isRepoPrivate, id) {
     // Currently always using private channels on Enterprise
-    const usePrivateChannel = this.get('config.enterprise') || isRepoPrivate
+    const usePrivateChannel = this.get('config.enterprise') || isRepoPrivate;
     const prefix = usePrivateChannel ? 'private-job' : 'job';
     return `${prefix}-${id}`;
   },
--- a/app/models/job.js
+++ b/app/models/job.js
@@ -170,10 +170,10 @@ export default Model.extend(DurationCalculations, DurationAttributes, {
     });
   },
 
-  @computed('repo.private', 'id')
-  channelName(isRepoPrivate, id) {
+  @computed('repo.private', 'id', 'features.enterpriseVersion')
+  channelName(isRepoPrivate, id, enterprise) {
     // Currently always using private channels on Enterprise
-    const usePrivateChannel = this.get('config.enterprise') || isRepoPrivate;
+    const usePrivateChannel = enterprise || isRepoPrivate;
     const prefix = usePrivateChannel ? 'private-job' : 'job';
     return `${prefix}-${id}`;
   },
--- a/app/models/job.js
+++ b/app/models/job.js
@@ -172,7 +172,9 @@ export default Model.extend(DurationCalculations, DurationAttributes, {
 
   @computed('repo.private', 'id')
   channelName(isRepoPrivate, id) {
-    const prefix = isRepoPrivate ? 'private-job' : 'job';
+    // Currently always using private channels on Enterprise
+    const usePrivateChannel = this.get('config.enterprise') || isRepoPrivate
+    const prefix = usePrivateChannel ? 'private-job' : 'job';
     return `${prefix}-${id}`;
   },
 


bnferguson:
As a side note (and bigger than this PR) is that at Past Jobs™ it's been nice to extract these sorts of feature flags away from their implementation and using a more descriptive name like `features.forcePrivateChannels` or `features.billingEnabled` even if it just wraps a single `features.enterprise` check.

This little bit of indirection gives us a few things: '

1. It gives a descriptive name to what we're trying to accomplish when making a piece of functionality conditional
2. It groups together functionality under a specific bit of intent, so that in the case that we want to change it in the future we can search for the feature flag and not sift through the Enterprise flags at the ground level that may be in models, templates, tests and others.
3. Gives us a single point to override, test, or change assumptions around when and if something is enabled

It seems like a really silly thing at the front side of development (a single config bit that just defers to another one), but can make a huge difference in moments where things shift. I'm not always good about making sure I do this in my own projects (because it does often seem sooo silly) but it came up again as I was working so figured I'd mention it.

backspace:
Hey. I’d be into renaming things to make them more clear. I’m actually not even that familiar with what all these things are and how they interact. I can open an issue for this, because having better nomenclature can definitely be helpful!

It’s my suspicion that in this case, `this.get('config.enterprise')` isn’t what you want because it’s accessing the job’s config, not the `travis-web` config. The application configuration needs to be imported; see [here](https://github.com/travis-ci/travis-web/blob/40b110a8fabbd17b3a0cf0f48888a801d80dbedf/app/routes/application.js#L6) for an example.

Maybe you’re reaching EOD but we could pair on looking into this if you like?

bnferguson:
> Hey. I’d be into renaming things to make them more clear. I’m actually not even that familiar with what all these things are and how they interact. I can open an issue for this, because having better nomenclature can definitely be helpful!

Sweet! Yeah digging into this more it looks like the one reference to `features.enterprise` may actually be a bug we sometimes see since that's not set anywhere. And `features.enterpriseVersion` is the one we use the most and `config.enterprise` is the weird one that seems to be mostly base level things? Not sure the distinction there. 

>It’s my suspicion that in this case, this.get('config.enterprise') isn’t what you want because it’s accessing the job’s config, not the travis-web config. The application configuration needs to be imported; see here for an example.

Ahaaaa, that makes sense. I wondered. Also I went back and saw that there's some things around `@computed` so I switched it back to closer to what it was pre #1269 with `@computed('repo.private', 'id', 'features.enterpriseVersion')` testing that now. If this doesn't work will check out the application importing and yeah pairing could be good! Web is one of my biggest blind spots when it comes to dev and testing for Enterprise. 

Thanks for the help!

backspace:
> Yeah digging into this more it looks like the one reference to `features.enterprise` may actually be a bug we sometimes see since that's not set anywhere. And `features.enterpriseVersion` is the one we use the most and `config.enterprise` is the weird one that seems to be mostly base level things? Not sure the distinction there. 

eep well it’s all pretty confusing. `config` is generally set either when the application is “built” with environment variables or hard-coded depending on “environment”. `features`, I believe are mostly controlled via the API; some of them are individual to a user, some are controlled for the entire deployment.

I wonder if we could make things throw warnings if an undefined flag is accessed, but that’s pretty against the way Javascript typically works 😬 

Unfortunately I’m the lone web person right now and these are things I’ve mostly not learned about yet haha

bnferguson:
@backspace OK, looks like that `@computed` style on the feature flag got me where I needed to be! If you give the 👍 I'm happy to do the deploy dance and get it merged!

bnferguson:
> Unfortunately I’m the lone web person right now and these are things I’ve mostly not learned about yet haha

haha I feel you there (but on Enterprise). I keep unearthing new wrinkles of things written long ago. (Not to mention somehow I've become a bash and infrastructure person?!?! Well, that wasn't intentional. 🤔) 

>eep well it’s all pretty confusing. config is generally set either when the application is “built” with environment variables or hard-coded depending on “environment”. features, I believe are mostly controlled via the API; some of them are individual to a user, some are controlled for the entire deployment.

Ahhhhh that makes sense. Good to know. I'll see about addressing that one dangling reference as I think it explains a bug we've seen. Also just makes it more consistent. 

And thanks again for looking at all of this so quickly!

PR #1426, opened 2017-12-08
Change wrapped log line alignment

backspace:


rsimha-amp:
@backspace With this change, the progress indicators used by our tests are no longer being wrapped, and are impossible to read. Earlier, they would automatically wrap to the width of the browser window, and would span however many lines were necessary to display all indicator dots.

For example, see https://travis-ci.org/ampproject/amphtml/jobs/316060543#L724

Is this intentional? If so, is there a way to override the behavior and have long log lines automatically wrap to the width of the browser window?

rsimha-amp:
To clarify, I'm not seeing a horizontal scrollbar in the travis web logs, so long lines are currently unreadable beyond the browser window width.

backspace:
wow, that’s a huge regression, I’m sorry about that. This was my attempt to address some alignment problems, but I didn’t realise this side-effect. I’ll revert it and try to find another way. Thanks for letting us know, @rsimha-amp!

(I was able to see at this [older test deployment](https://bd-cousine.test-deployments.travis-ci.org/ampproject/amphtml/jobs/316060543#L724) what it used to look like.)

rsimha-amp:
@backspace Yep, getting back the old view would be nice :)

backspace:
Okay, wrapping is restored, @rsimha-amp, thanks again!

backspace:
I’m looking at another solution [here](https://github.com/travis-ci/travis-web/pull/1443).

rsimha-amp:
Thanks @backspace. Did you manage to figure out the cause of the issue I reported in https://github.com/travis-ci/travis-web/pull/1442#issuecomment-351567937 ?

PR #1427, opened 2017-12-11
Dj trusty updates q4 banner

cotsog:


app/components/job-infrastructure-notification.js
```diff
@@ -19,15 +19,21 @@ export default Component.extend({
   @equal('queue', 'builds.linux') isLegacyInfrastructure: null,
   @equal('queue', 'builds.ec2') isTrustySudoFalse: null,
 
-  @computed('job.startedAt', 'job.config')
-  isTrustyStable(startedAt, config = {}) {
-    if (config.dist === 'trusty' && config.group === 'stable') {
+  @computed('job.startedAt', 'queue', 'job.config')
```
backspace: I think this should be `job.config.dist` for the last dependent key, but maybe you had that before and it didn’t work?

backspace:
hmm I see from Slack that I shouldn’t have merged this but I won’t deploy so I think it’s okay 😬

PR #1429, opened 2017-12-11
Update to Ember 2.17

backspace:
This has a failing test related to confusing innards involving Pusher
and Mirage 😖


backspace:
There are these deprecation warnings:

image

But I gather they’re from addons, as I’ve replaced all the imports I can find.

PR #1430, opened 2017-12-12
Trusty updates Q4 banner

cotsog:
Updated timestamp to `2017-12-12T16:15:00-00:00`.


app/components/job-infrastructure-notification.js
```diff
@@ -33,10 +33,10 @@ export default Component.extend({
 
   @equal('queue', 'builds.macstadium6') isMacStadium6: null,
 
-  @computed('queue', 'job.config,dist')
-  isPreciseEOL(queue, dist) {
+  @computed('queue', 'job.config,dist', 'job.config.language')
```
backspace: ah there’s a `,` in `job.config,dist` that should probably be a `.`?

cotsog: :man_facepalming:

Thanks for spotting that one, @backspace!

PR #1440, opened 2017-12-13
Use config from V3

drogus:
This PR uses config from V3 instead of querying V2.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -7,7 +7,8 @@ module.exports = {
   parser: 'babel-eslint',
   extends: 'eslint:recommended',
   env: {
-    browser: true
+    browser: true,
+    es6: true
   },
   rules: {
     // Don't allow unused vars, but allow unused arguments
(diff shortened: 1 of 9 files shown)


drogus:
I just realized that my simplification of config loading misses the "current tab". We may need config there as well. I will need to rethink this approach as I'm now not so sure if the simplification is worth it if we need to explicitly load config in so many places.

drogus:
@backspace the problem is more about the matrix on the current tab - with the current code in this PR it won't work unfortunately :( So I think I'll have to rework this PR to get back to lazy loading whenever the config is missing. That said, it might be easier than with what I did for V2.

backspace:
oops! Understood.

drogus:
I was thinking about this issue more and while the new approach requires a bit of housekeeping (ie. preloading the config explicitly on the pages where we need it), I think I still like it more than lazy loaded version. The latter requires more async code in the controllers/components layer which makes things harder to understand. So now I'm wondering if there would be a way to make "current tab" more explicit in how it reacts to data load.

drogus:
I'm going through the code for the current tab and I feel it's getting more and more complicated and I'm tempted to rewrite it to make setting the current build more explicit.

drogus:
I ended up modifying job config fetcher to use V3 API, but I would love to rewrite all of this at some point.

PR #1442, opened 2017-12-13
Add long unbroken text for Percy

backspace:
This is meant to prevent a regression reported in #1426.


backspace:
Hey, could you open an issue [here](https://github.com/travis-ci/travis-ci/issues) for that, including a link to a log that shows that problem? That’ll make it far easier to look into.

That said, I’m on vacation for a while starting tomorrow, so I’ll have to look into it when I get back, sorry.

PR #1443, opened 2017-12-13
Change log wrapping alignment

backspace:
This is an attempt to address #1426 without breaking other things. It doesn’t work yet.


backspace:
Success!

PR #1444, opened 2017-12-14
api tracing component

igorwwwwwwwwwwwwwwwwwwww:
This is a new feature that can be enabled by visiting `/toggle-tracing`. Once enabled, it will display a little tracer button on the bottom left of the page.

screen shot 2017-12-19 at 17 02 32

Clicking this button will open up a little panel that shows all api requests that happened, and includes a link to the trace in papertrail.

screen shot 2017-12-19 at 17 08 34

The color theme was adapted from [sailorhg](http://twitter.com/sailorhg)'s [fairyfloss](http://sailorhg.github.io/fairyfloss/) theme.

The XHR replacement implementation was adapted from [the symfony profiler](http://symfony.com/blog/new-in-symfony-2-6-ajax-requests-in-the-web-debug-toolbar).

refs travis-ci/travis-api#607


backspace:
If you need to add this in more places, like `app/services/ajax` to catch non-Ember Data requests, it might make sense to treat it like a `feature`, to centralise the checking.

backspace:
re the Percy failures, did you branch off an old `master` maybe?

igorwwwwwwwwwwwwwwwwwwww:
@backspace spot on!

drogus:
I looked into the code and it seems that we not only use our own ajax functions, but sometimes jQuery directly as well. It's basically a mess and I would like to fix it by unifying all of our XHR calls. It shouldn't be really hard to do, so I might give it a shot next week. If you don't want to wait I could also help you come up with a way to make it work with jQuery (we could register a global callback and check if the request is to our API) and then also with the ajax service.

igorwwwwwwwwwwwwwwwwwwww:
I've some web-based UI, please take another look!

PR #1461, opened 2018-01-09
Change PR titles to commit messages

drogus:
When you view a PRs list you only see the PR title and no commit
message, which might be confusing as to what changes were made in the
commit.


backspace:
This still seems like a good idea to me, or at least that it’s useful to be able to see commit messages on PR build headers. Is it something we should return to, @drogus?

PR #1479, opened 2018-01-17
Fix badge pop up menu overriding the default branch

berraknil:


tests/acceptance/repo/show-test.js
```diff
@@ -0,0 +1,36 @@
...
+
+test('loading branches doesnt update the default branch on the repo', function (assert) {
+  server.logging = true;
+  page.visit({ organization: 'killjoys', repo: 'living-a-feminist-life' });
+  page.openStatusImagePopup();
+
+  andThen(() => {
+    const url = new URL(page.statusBadgeImageSrc);
+    const expectedPath = `${url.pathname}?${url.searchParams}`;
+    assert.equal(expectedPath, '/killjoys/living-a-feminist-life.svg?branch=feminist%23yes');
```
rwrede: Was just having a quick look.. too late because already merged. I know. But shouldn't that be the other way round? 

-> `assert.equal('/killjoys/living-a-feminist-life.svg?branch=feminist%23yes', actualPath);`

berraknil: pinging @drogus as I was mostly observing the process on this one

backspace: I’m not @drogus but hopefully it’s okay for me to respond… since it‘s `assert.equal`, it doesn’t _really_ matter. (If there were a third argument, it would be the text to show when the assertion fails.) I think `actualPath` is a better name, indeed. But scanning our codebase, the order is usually `assert.equal(actual, expected)` so I think the order fits that.

drogus: yes, it should totally be `actualPath` instead of `expectedPath`, good catch! I'll change the name to not make it confusing to read.

PR #1480, opened 2018-01-17
Highlight the active organization on the sidebar

berraknil:
Moved the avatar inside the link-to component so now the whole sidebar item is effected when the link is active. 

We can 
- only change the colors on hover/active,
- add underline to the text,
- change the both background color and text color (like in the example below)

layout

Will wait to hear from design team to finalize any design decisions on this, but making a PR now to create a proof of concept everyone can check out.


berraknil:
@drogus @backspace This is all [approved by the design team](https://github.com/travis-pro/team-teal/issues/2494#issuecomment-359886750) But still failing Percy diffs, there are overall design changes which we are aware of and those are expected to fail, but this bit is only showing erroneously on Percy and I am unable to reproduce it anywhere else (including various browsers on BrowserStack).

whyisthis


So looking for reviews/input before proceeding, also should I add/write tests for this new functionality in this PR as well? In that case I might need a little bit of help with test questions. Thank you both!

berraknil:
In addition to BrowserStack, also tested the visual diffs on an Ubuntu 16.04 with Chrome latest stable (which is what our build is running) results are below

ubuntuchromelatest


ubuntuchromelatestipadhorizontalres


So, although still not sure what is throwing Percy off, I am going to merge this today and if it somehow breaks on master then will try to see what is happening with it.

PR #1482, opened 2018-01-18
Only show the job duration after the job starts

berraknil:
Also fixes the wrong date-time issue on hover when the elapsed time is a negative number


berraknil:
Scrapping this for now per https://github.com/travis-pro/team-teal/issues/1508

Will probably just make a temporary fix to prevent this behavior until the overall improvements are done

35277266-3692c2d8-0057-11e8-81f0-cfdea1eec2ba

PR #1484, opened 2018-01-22
Update Ember and family to 2.18

backspace:
This is a first pass after a complex merge. I expect many
lint failures and deprecations.

I wonder whether we can just import the Airbnb ESLint config
rather than duplicating it?


backspace:
There are two duplicate deprecation warnings when running the server:

```
DEPRECATION: An addon is trying to access project.nodeModulesPath. This is not a reliable way to discover npm modules. Instead, consider doing: require("resolve").sync(something, { basedir: project.root }). Accessed from:   NPMDependencyVersionChecker (/Users/b/Documents/Travis/travis-web/node_modules/ember-cli-version-checker/src/npm-dependency-version-checker.js:11:32)
DEPRECATION: An addon is trying to access project.nodeModulesPath. This is not a reliable way to discover npm modules. Instead, consider doing: require("resolve").sync(something, { basedir: project.root }). Accessed from:   NPMDependencyVersionChecker (/Users/b/Documents/Travis/travis-web/node_modules/ember-cli-version-checker/src/npm-dependency-version-checker.js:11:32)
```

From what I can tell, these come from addons that have an old version of `ember-cli-version-checker`. I updated the addons I could… meh

I haven’t looked at deprecations in the console yet.

backspace:
This seems sufficient; I’ll merge and deploy tomorrow.

drogus:
I'm super happy to see this, because I'd like us to try out 2 new things: [Glimmer named arguments](https://github.com/emberjs/rfcs/pull/276) and [ES5 getters](https://github.com/emberjs/rfcs/pull/281)

PR #1486, opened 2018-01-22
Restore silenced deprecations

backspace:
What will happen


backspace:
hmm the router deprecation was [removed in 2.16](https://www.emberjs.com/deprecations/v2.x/#toc_ember-router-router-renamed-to-ember-router-_routermicrolib) and I probably fixed it in [the update](https://github.com/travis-ci/travis-web/pull/1368/files)! The macro one isn’t related to the Ember version, so I think we might be ready for 3.0! 🎉

PR #1495, opened 2018-01-28
Fix status icon clipping

berraknil:
This fixes https://github.com/travis-pro/team-teal/issues/2524 by making the SVG overflow property visible to prevent rectangle shaped path clipping over the icons.

### DIFF
--- a/app/styles/app/modules/status-icon.sass
+++ b/app/styles/app/modules/status-icon.sass
@@ -5,6 +5,7 @@
     stroke-linecap: round
     stroke-linejoin: round
     stroke-miterlimit: 10
+    overflow: visible
 
 .status-icon
   @extend %icon


berraknil:
This seems to be good to go other than Percy (naturally) freaking out because suddenly all of our icons are FULL instead of clipped 🎉 

I have checked the visual diffs as much as I could but there are 51 diffs in total, so any more checks before pushing are much appreciated.

berraknil:
[SVG Icons are cut off on parts of the UI](https://trello.com/c/GC0oozBw/379-svg-icons-are-cut-off-on-parts-of-the-ui)

ctrlaltjustine:
@berraknil just checked it out on my end and looks good to me! Thanks for looking into that so quickly!

berraknil:
@ctrlaltjustine Thanks for catching it in the first place and checking the fix so quickly! Will deploy the fix today :)

PR #1498, opened 2018-01-29
Fix broken pagination links containing decimals

clekstro:
Users were able to manually specify the offset in the URL, which led to rounding errors in the rendered pagination links.

This makes it simpler to use from a user's perspective and also avoids exposing something the user should not be able to influence (yet). We can always re-evaluate this later if users would find it useful, as API still supports it.

### DIFF
--- a/app/components/pagination-navigation.js
+++ b/app/components/pagination-navigation.js
@@ -77,7 +77,7 @@ export default Component.extend({
         }
       }
 
-      // ... devider unit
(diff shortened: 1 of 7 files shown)


berraknil:
[Profile pagination links can be misnumbered](https://trello.com/c/5JE9LbEE/381-profile-pagination-links-can-be-misnumbered)

PR #1500, opened 2018-01-30
[Bugfix]: Ensure tab is set as `builds` regardless of auth status

clekstro:
This should address https://github.com/travis-pro/team-teal/issues/2520, where builds for a public repo were not being loaded when clicking "Show more."


clekstro:
Looks good in the PR env. Merging.

PR #1510, opened 2018-02-05
[Reliability] Remove broken offset parameter from links

berraknil:
Since we changed the pagination from offset to page, we don't need to send an _offset=0_ to **link-to** components on templates. It breaks active org/user state for highlighting as we don't use the offset at all anymore.


berraknil:
[Remove broken offset parameter from links](https://trello.com/c/xT5YGSL0/402-remove-broken-offset-parameter-from-links)

PR #1514, opened 2018-02-05
Add better handling of OS icons

backspace:


backspace:
ah good catch, thanks. I’m not up on this process, sorry. It’s true that the Power-handling is Enterprise-specific.

PR #1518, opened 2018-02-06
Bump ember-cli-mirage to 0.4.x series

clekstro:
We've avoided upgrading for too long because of the perceived amount of work to change our relationship definitions. I'm going to try and fix this incrementally, these first few commits being the first chunk.

### DIFF
--- a/mirage/models/branch.js
+++ b/mirage/models/branch.js
@@ -1,7 +1,7 @@
 import { Model, hasMany, belongsTo } from 'ember-cli-mirage';
 
 export default Model.extend({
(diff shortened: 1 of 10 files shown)


clekstro:
I don't want to accept the current Percy issues, so I'll fix them in the morning.

clekstro:
After looking closer, this appears to not be a data/mirage issue. Pausing the relevant test shows the screenshot as expected, meaning it appears that the screenshot used by Percy is perhaps taken a bit too soon? Merging.

PR #1520, opened 2018-02-07
[Reliability]: Convert SASS syntax to SCSS

berraknil:
Related issue [here](https://github.com/travis-pro/team-teal/issues/2538) for more information


berraknil:
[Switch the codebase from SASS syntax to SCSS syntax](https://trello.com/c/7tmsQLLT/403-switch-the-codebase-from-sass-syntax-to-scss-syntax)

berraknil:
@backspace I am not even sure what is Percy even doing at this point, it is also on my todo list for this week, what is funny is it accepted the wrong diff as the correct one now (look at the profile list item on the left where token is still out of the span) and only reports an avatar change which is not even there? Anyway. Maybe best to leave this merge to tomorrow

backspace:
I think it’s fine to merge today if you want but you do you 😆

PR #1521, opened 2018-02-07
Add experiment with ember-in-viewport

backspace:
I’m still looking to replace the Bower-imported Waypoints
jQuery plugin with something. I’ve figured out why
ember-in-viewport didn’t work last time I tried:
didInsertElement was missing the super call! 😱

This still isn’t working as expected though, despite all
my fiddling attempts with the settings. It shows the
top-bar having exited the viewport immediately upon scrolling,
rather than after it’s completely disappeared. I’m pushing
just to save this for later revisiting.

The tests will fail.

### DIFF
--- a/app/components/top-bar.js
+++ b/app/components/top-bar.js
@@ -21,8 +21,6 @@ export default Component.extend(InViewportMixin, {
   classNames: ['top'],
   landingPage: false,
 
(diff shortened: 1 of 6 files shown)


backspace:
I don’t know why the Percy status isn’t updated here, but there are no diffs.

PR #1526, opened 2018-02-12
Update ember-cli-sentry

backspace:
I want to see whether this breaks anything so we’re ready
when the final version is released.


backspace:
3.0.0 is out now, I’m going to merge this! 🎉

PR #1533, opened 2018-02-14
Improve overall token UX on profile page

clekstro:
This does a few things:
* Allows users to directly copy the token on the profile page (instead of merely hiding/showing), improving the flow.
* Switches the token that is shown/copied from the "Travis token" to the "API token"
* Removes the "eye" icon for a simpler button-based UI

See https://github.com/travis-pro/team-teal/issues/2559 for more details.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -339,6 +339,8 @@ module.exports = {
       },
       globals: {
         server: true,
+        triggerCopySuccess: true,
+        triggerCopyError: true,
         signInUser: true,
         withFeature: true,
         percySnapshot: true,
--- a/app/adapters/application.js
+++ b/app/adapters/application.js
@@ -24,7 +24,7 @@ export default ActiveModelAdapter.extend({
     hash.headers['accept'] = 'application/json; version=2';
     hash.headers['X-Client-Release'] = config.release;
 
-    let token = this.get('auth').token();
+    let token = this.get('auth.token');
     if (token) {
       if (!hash.headers['Authorization']) {
         hash.headers['Authorization'] = `token ${token}`;
--- a/app/adapters/v3.js
+++ b/app/adapters/v3.js
@@ -50,7 +50,7 @@ export default RESTAdapter.extend({
 
     hash.headers = hash.headers || {};
 
-    let token = this.get('auth').token();
+    let token = this.get('auth.token');
     if (token) {
       hash.headers['Authorization'] = `token ${token}`;
     }
--- a/app/components/branch-row.js
+++ b/app/components/branch-row.js
@@ -44,7 +44,7 @@ export default Component.extend({
         }
       };
       if (this.get('auth.signedIn')) {
-        options.headers.Authorization = `token ${this.auth.token()}`;
+        options.headers.Authorization = `token ${this.auth.token}`;
       }
       let path = `${apiEndpoint}/repo/${repoId}/builds`;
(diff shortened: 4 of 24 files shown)


clekstro:
It might actually be nicer to show dots instead of the current 'hidden' text:
screen shot 2018-02-14 at 14 18 58

clekstro:
@Nat-ess @ctrlaltjustine It would be great to get your feedback on this as well. I think this could be shipped in the interim but am definitely open to improving things aesthetically.

backspace:
This is better in accessibility terms; the current UI doesn’t allow you to reveal the token with the keyboard (let alone copy), and this does! So, nice, thanks.

ctrlaltjustine:
Thanks for working on this! I agree having the copy button is going to be really helpful. Out of curiosity, if these were icons instead of buttons, could we make them inline with the token? Or is there a concern about doing that?

clekstro:
Hey @ctrlaltjustine! Thanks for taking a look :)

There were a few reasons I went with simple text buttons as opposed to just adding another icon to the current layout.

First, I wanted to avoid the current "bounciness" of having the icon position move as the state changed. I thought it was important to show some sort of success message after copying the token, which would have added to the randomness of the icon positions depending on the length of the text. We could probably get around this by aligning the icons to the right and fixing their position though.

The other reason was that I simply didn't have the icon. Also, when I tried to imagine how that would look with the tooltips as you hovered over them (think moving the mouse to the furthest right icon), I think it felt busy/cramped because of the two animations in such quick succession and that it was likely the tooltips would (briefly) overlap.

ctrlaltjustine:
@clekstro thanks for that explanation. And I definitely considered the icon alignment as a potential obstacle. @Nat-ess and I spoke about this issue last night and we could potentially do buttons with text and icons.

Would something like this make sense?
screen shot 2018-02-14 at 3 16 16 pm


Another thing of note that Natalia and I spoke about was that the hidden password styles seem smushed a bit. Is it possible to increase the width between `•••••`?

clekstro:
@ctrlaltjustine This looks amazing! I can definitely add some padding to the hidden key styles. Do you already have the copy icon you'd like used handy?

ctrlaltjustine:
Yessir, here's the icon _(zipped because svg isn't supported in issues)_

💾  [icon-copy.svg.zip](https://github.com/travis-ci/travis-web/files/1728533/icon-copy.svg.zip)

berraknil:
@clekstro as a reminder, the way the component markup is right now (link-to element being the wrapper) doesn't fix the [Percy problem](https://github.com/travis-pro/team-teal/issues/2522) as we still have interactive elements (buttons/anchors) inside an anchor tag, which apparently what broke the layout due to it being illegal in spec. 

Just pinging you here again since you might want to split the component into two, to make the anchor tag and the tokens siblings instead, though in that case the active styling won't work the way it is and will need .active class followed via Ember I believe as @drogus mentioned on Slack the other day.

clekstro:
@berraknil I think I'm going to tackle that in a separate PR. It was an issue that existed prior to this work and while it's important to fix, I don't want to hold this up in the interim.

clekstro:
@ctrlaltjustine This is now live in production. I think this looks very close to the mock up, but if there are adjustments that you feel need to be made, just let me know :)

Here's what production looks like:
screen shot 2018-02-22 at 15 31 19

PR #1537, opened 2018-02-16
Fix ember-cli-page-object import deprecations

clekstro:


backspace:
oops, I didn’t know I’d missed those, thanks!

PR #1542, opened 2018-02-20
Remove lint link

backspace:
Is this tested? WE SHALL SEE


backspace:
hah wow I actually didn’t mean to merge that, oops… but I think it’s okay, apart from the weird commentary

PR #1544, opened 2018-02-21
Migrate to npm Pusher

backspace:
Maybe this is mostly working???


backspace:
I have two `FIXME`s of question here, but it seems to be working! 🎉

app/utils/pusher.js
```diff
@@ -38,7 +39,21 @@ TravisPusher.prototype.init = function (config, ajaxService) {
...
   return this.pusher = new Pusher(config.key, {
     encrypted: config.encrypted,
-    disableStats: true
+    disableStats: true,
+    authorizer: function (channel, options) {
+      return {
+        authorize: function (socketId, callback) {
+          let channelName = channel.name;
+          // FIXME this string was Pusher.channel_auth_endpoint, where is it from?
+          TravisPusher.ajaxService.post('/pusher/auth', {
```
backspace: Here’s [the documentation](https://pusher.com/docs/authenticating_users#authEndpoint):

image

But is there a way this could differ?

drogus: I think that I was using `Pusher.channel_auth_endpoint`, because this is what Pusher uses by default, but I'd be fine with owning it ourselves, ie. we could move it to `config/environment.js` and define it ourselves to `/pusher/auth`

drogus: And responding to "But is there a way this could differ?", I guess no - we have never changed it and we basically define it in the API, so in order for it to change we would need to change the API.

So I guess even hardcoding it here is fine.

app/utils/pusher.js
```diff
@@ -156,22 +171,12 @@ TravisPusher.prototype.ignoreMessage = function (message) {
   return existingSubscription || noSubscription;
 };
 
-Pusher.SockJSTransport.isSupported = function () {
-  if (ENV.pusher.host !== 'ws.pusherapp.com') {
-    return false;
-  }
-};
-
-Pusher.channel_auth_transport = 'travis_ajax';
```
drogus: Unfortunately this needs to stay as is for now. API implements pusher auth differently than Pusher, because in the past when we were using a channel per repo approach we needed to auth in bulk. We can change it in the API, but until we do, we need a custom authorizer.

backspace: I think this is still working, they just changed the interface, so I moved it into the `authorize` method.

drogus: 👍

backspace:
I’ve compared this on the PR deployments against production and live updates look to be behaving the same. 🎉

PR #1547, opened 2018-02-22
Disambiguate feature flags

clekstro:
Move us away from a blanket `proVersion` feature flag toward more declarative ones as discussed in https://github.com/travis-pro/team-teal/issues/2433.

Also adds the `ember-truth-helpers` addon to allow for a bit more flexibility in terms of managing template state. Not all booleans or simple logical operations need be a computed property.

This keeps the current `TRAVIS_PRO` configuration via command line/build args in order to keep it compatible with our current processes, and just sets the proper config accordingly based on the logic that was in the app.


clekstro:
I need to test this out in an Enterprise build to make sure this hasn't accidentally broken anything. It would ideally be covered by tests, but I've got a bit less confidence as a result of yesterday's snafu.

app/templates/loading.hbs
```diff
@@ -8,28 +8,25 @@
...
+                <button {{action 'markBroadcastAsSeen' broadcast}} title="hide this broadcast" class="broadcast-close no-button" type="button">
+                  {{svg-jar 'icon-failed' class="icon"}}
+                </button>
+              </p>
+            </li>
+          {{else}}
+            <li><p>There are no broadcasts</p></li>
+          {{/each}}
+        </ul>
+      {{/if}}
```
backspace: wow, I don’t even know about this template, this duplication should probably be extracted? But that’s orthogonal to this PR.

app/templates/owner-loading.hbs
```diff
@@ -8,28 +8,25 @@
...
+                <button {{action 'markBroadcastAsSeen' broadcast}} title="hide this broadcast" class="broadcast-close no-button" type="button">
+                  {{svg-jar 'icon-failed' class="icon"}}
+                </button>
+              </p>
+            </li>
+          {{else}}
+            <li><p>There are no broadcasts</p></li>
+          {{/each}}
+        </ul>
+      {{/if}}
```
backspace: 😳 samesies

tests/acceptance/home/with-repositories-test.js
```diff
@@ -56,9 +56,11 @@ moduleForAcceptance('Acceptance | home/with repositories', {
   }
 });
 
-test('the home page shows the repositories', (assert) => {
+test('the home page shows the repositories', function (assert) {
   sidebarPage.visit();
 
+  // test fails without this just by adding the feature flag :tableflip:
+  andThen(() => {});
```
backspace: I wonder why this is happening to us 😞  maybe it’ll magically go away with the new test setup hehe

app/templates/loading.hbs
clekstro: Yeah, for sure. The loading templates definitely need to be consolidated.

backspace:
I forgot to say: there’s an enterprise deployment that builds the manual, so this can probably be tested on that!

clekstro:
@bnferguson @schultyy I think this is good to go, but from my testing this *will* impact Enterprise as things currently stand. Once this is merged, we'll be showing the "Running" tab when the `showRunningJobsInSidebar` flag is enabled. This is set automatically within `config/environment.js` to true when the app is built passing `TRAVIS_PRO=true`. From a product/business perspective, this is correct, but afaict Enterprise is still being built with that flag passed, which needs to be remedied.

I'd be happy to sync up to discuss how to untangle this. I vaguely recall it being related to Pusher, but you all may have more insight here.

clekstro:
The current Percy issue is because the current screenshot shows a GATEWAY TIMEOUT error. Ignore it.

bnferguson:
Thanks for the heads up! 

>  Once this is merged, we'll be showing the "Running" tab when the showRunningJobsInSidebar flag is enabled. This is set automatically within config/environment.js to true when the app is built passing TRAVIS_PRO=true. From a product/business perspective, this is correct, but afaict Enterprise is still being built with that flag passed, which needs to be remedied.

So sounds like so far this should build and nothing changes? That's awesome! What I don't completely understand is why removing `TRAVIS_PRO` from Enterprise builds. Are we trying to deprecate that tag or something similar? I've always been under the impression that Enterprise technically is a special case of Travis Pro, so assumptions there can be brought to Enterprise as well (with some adjustments, which is why the Enterprise env var exists). 

That said, I can also see wanting to make these lines harder. So Travis Pro is it's own thing and Enteprise is its own thing env wise. I'm hip to that. In which case seems like should adjust `config/environment.js` and `showRunningJobsInSidebar` to set it when Enterprise is true as well?

clekstro:
> That said, I can also see wanting to make these lines harder. So Travis Pro is it's own thing and Enteprise is its own thing env wise. I'm hip to that. In which case seems like should adjust config/environment.js and showRunningJobsInSidebar to set it when Enterprise is true as well?

That's a reasonable solution. I just always get confused when seeing `if pro but not enterprise` in the code, which is part of the reason I'd like to make a clear distinction between the two. It might be better to be ultra-specific here, i.e. defining every flag for every environment, instead of trying to write logic in the config to match current app behavior (what things look like now in the PR). I'll merge after making those changes, but definitely let me know if you see anything unexpected as a result.

clekstro:
@bnferguson So I've made the necessary changes and will be merging/deploying this tomorrow morning.
From now on, in order to enable one a feature for enterprise, you can customize [this file](https://github.com/travis-ci/travis-web/blob/0aaa1bf5bd8d76c0cd54ce31806472f494bef327/app/instance-initializers/enterprise-environment.js). That code is only run when the `TRAVIS_ENTERPRISE` flag is passed at build time, and the Enterprise config will win over the pro config if both are passed. I've included all of our current flags so that it's more explicit.

We now also have test coverage so we don't inadvertently bork this stuff going forward.

bnferguson:
This is great! Thanks so much. One question: 

` featureFlags['show-running-jobs-in-sidebar'] = false;` what does this do? I recently noticed that all running jobs seemed to be gone in master, product change? Do we show them in .com?

clekstro:
@bnferguson No product changes. Currently, that feature is only enabled on .com. 

The way this is set up, we have the base configuration that's valid for .org in [config/environment.js](https://github.com/travis-ci/travis-web/blob/master/config/environment.js#L68-L75). 

For pro and enterprise, those flags are set at boot time through a two-step process. First, we set environment-based feature flags based on the ENV vars passed at build time [here](https://github.com/travis-ci/travis-web/blob/master/config/environment.js#L79-L85). Then, at boot time, we execute the initializers and set the feature flags accordingly. You can see the pro version [here](https://github.com/travis-ci/travis-web/blob/master/app/instance-initializers/pro-environment.js#L10) (where the flag in question is enabled) and the enterprise version [here](https://github.com/travis-ci/travis-web/blob/master/app/instance-initializers/enterprise-environment.js) where it's turned off.

The main benefits of this are that I was able to get test coverage around this as well as ensure we can see per-environment which flags are enabled. It is perhaps slightly unexpected that the configs are separate spots, but I felt it worth the tradeoff.

PR #1548, opened 2018-02-22
attempt to fix api token on branches page

igorwwwwwwwwwwwwwwwwwwww:
refs #1533


drogus:
This will fix the problem, but the root of the problem is using custom `jQuery.ajax` call, it would be better to add `@service api: null` at the top of the class and then change the call to `this.get('api').get(url).then(`

PR #1554, opened 2018-02-26
Add homepage screenshot acceptance test

backspace:
This is a proof-of-concept for an approach.


backspace:
This probably doesn’t need to be automated, but if it does, this’ll work for reference.

PR #1557, opened 2018-02-28
Load feature flags from local storage

clekstro:
Currently, we make a request to the `/beta_features` API endpoint to ensure we fetch the feature flags server-side before rendering the app (which may behave differently depending on the flags enabled). Now, instead of always requesting the feature flags, this persists them to local storage after the first time they're fetched.

They're only re-fetched when visiting the `/features` page.

This state is also persisted to local storage when enabling/disabling a given flag.


clekstro:
I've tested this extensively in the PR env, don't expect any issues. Merging.

PR #1574, opened 2018-03-09
Improve UX of Show More buttons

clekstro:
Currently in production, clicking on the "Show More" buttons when on the Build History/Pull Requests tab, nothing happens visually. The data loads, but the button doesn't do anything.

It appears this is a bug with the current implementation, as it expects us to change the button text to "Loading" and disable the button once it has been clicked.

Instead, this PR takes a different approach. Clicking the button will now replace the button with the loading indicator that is shown when the page loads, i.e. our normal `{{loading-indicator}}` component. This eliminates our need to disable the button. We also now use an ember-concurrency task to use derived state and the `perform` helper to radically reduce the complexity around this button.

Before:
current

After:
future

### DIFF
--- a/app/components/show-more-button.js
+++ b/app/components/show-more-button.js
@@ -3,5 +3,5 @@ import Component from '@ember/component';
 export default Component.extend({
   tagName: 'button',
   classNames: ['showmore-button', 'button'],
-  label: 'Show More',
+  label: 'Show more',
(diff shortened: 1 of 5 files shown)


igorwwwwwwwwwwwwwwwwwwww:
awesome! 👍

PR #1578, opened 2018-03-09
Add requests acceptance tests

backspace:
I included `ember-test-selectors` here, which I remember you suggesting a while ago, @clekstro, but maybe that never happened? It just seemed weird to me to have to use `eq(3)` etc to get at the sections of a request. But I could remove it if it seems weird to have both page objects and test selectors.


clekstro:
@backspace I'm 👍 on including it. I agree that it might be a little weird to have both, but tbh I'd prefer to use test selectors over page objects in the medium to long term. If it makes this test more readable, it'd probably be better in other situations as well.

PR #1580, opened 2018-03-13
Request images from API directly instead of thru waiter redirect

clekstro:
We discussed this a long, long time ago in https://github.com/travis-pro/team-teal/issues/1077. I've tested the markdown links, but would feel more comfortable testing the other image types a bit more before merging.

### DIFF
--- a/app/services/status-images.js
+++ b/app/services/status-images.js
@@ -7,27 +7,20 @@ export default Service.extend({
   @service features: null,
 
   imageUrl(repo, branch) {
-    let prefix = `${location.protocol}//${location.host}`;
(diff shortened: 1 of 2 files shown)


igorwwwwwwwwwwwwwwwwwwww:
I think there is some branding value in keeping it under the main domain. the same way you can `git clone git@github.com/<repo>`, you can use `travis-ci.org/<repo>.svg` for status badges.

getting rid of that redirect is a super good idea though. an alternative approach we could consider is installing the nginx buildpack on web and proxying the badges instead of redirecting.

PR #1583, opened 2018-03-13
Don't upload error pages to S3

drogus:
Some errors on heroku need a link to special error pages for error pages
that don't even load the app (500 from heroku, maintenance etc). We
decided to generate those pages from travis-web, but I think that it
doesn't really make sense anymore. For some reason generating those
pages broke and maintenance page was missing styles. Instead of
debugging it I think it would be better to just keep the error pages on
S3 and modify them there if needed. It would also add a benefit of not
making travis-web's builds longer.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -48,23 +48,5 @@ script:
 after_script:
   - greenkeeper-lockfile-upload
 
-before_deploy:
-  - ASSETS_HOST=https://s3.amazonaws.com/travis-error-pages ember build --env production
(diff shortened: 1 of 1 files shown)


joecorcoran:
Might be nice to instead have a new repo for the maintenance page(s) and a script to upload directly.

drogus:
@joecorcoran yup, that would be good. I'll merge it right away anyway, to not overwrite what we have there at the moment and then I'll open an issue about it.

PR #1586, opened 2018-03-14
always lazy-load build.request

igorwwwwwwwwwwwwwwwwwwww:
fetching build request every time we fetch a repo is quite expensive, especially on the dashboard. this should improve load times across the board.

refs https://github.com/travis-ci/reliability/issues/37

### DIFF
--- a/app/adapters/build.js
+++ b/app/adapters/build.js
@@ -1,7 +1,7 @@
 import V3Adapter from 'travis/adapters/v3';
 
 export default V3Adapter.extend({
-  includes: 'build.commit,build.branch,build.request',
+  includes: 'build.commit,build.branch',
 
   pathPrefix(modelName, id, snapshot, type, query) {
     if (type === 'query' && query.repository_id) {
--- a/app/adapters/job.js
+++ b/app/adapters/job.js
@@ -1,8 +1,6 @@
 import V3Adapter from 'travis/adapters/v3';
 
 export default V3Adapter.extend({
-  includes: 'build.request',
-
   coalesceFindRequests: true,
 
   groupRecordsForFindMany(store, snapshots) {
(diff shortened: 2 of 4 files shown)


igorwwwwwwwwwwwwwwwwwwww:
I'm expecting this to requires some `include` overrides in a couple places.

backspace:
Is there any way in the API for us to find a build’s request? I couldn’t find a way, which is why I ended up adding to `includes`. While some kind of conditional `includes` could work, it would be a bit of a mess, and would require reloading a build model in the absence of a way to fetch a build’s request.

Alternately, if `request_id` were included in the build’s minimal representation, we could get the build messages that way. Since it’s already a field on that table, maybe that’s not too onerous?

igorwwwwwwwwwwwwwwwwwwww:
@backspace thanks for the extra context, we are indeed missing the `request_id`, and the minimal representation does include a bunch of fields that requires fetching the records.

I think I would like to change that to have the minimal representation of request include only the id and then include that on build by default.

joecorcoran:
The build.request is available at the moment, as long as you include it in the params. It was implemented this was rather than by default, since including the request in a list of builds (`/repo/:id/builds`) is expensive, and the API doesn't differentiate between resource representations for single/list views (e.g. show me one build vs show me all the builds).

backspace:
My suggestion is that the minimal representation of a build could include `request_id` (not the whole request), which seems near-free to me because it’s already in that table. Because of the way Ember Data works, it’s harder to re-request a build with an extra `includes`; I could do it outside Ember Data, but it’ll mean the build messages are slower to load and won’t be cached.

We don’t even care about the request’s other properties at all, we just need it to get the build’s messages.

igorwwwwwwwwwwwwwwwwwwww:
Closing this stale PR, hopefully we can revisit it in the future.

PR #1588, opened 2018-03-15
Fix Pusher in Enterprise

bnferguson:
This PR is building off of @clekstro's work. 

Moving to the NPM version of Pusher in https://github.com/travis-ci/travis-web/pull/1544 caused a few issues as some of the interfaces changed, namely around setting a custom host and path. This did not affect hosted as it uses the defaults (read: the actual Pusher service) but in Enterprise we use "slanger", a Ruby version of the Pusher server that runs in the container. Additionally it runs not at a domain but under a path, much like API. 

This PR fixes this issue by passing in the new properties while initializing the Pusher class (`wsHost` and `wsPath`). Additionally it removes some code that appears to be dead as of the move to NPM pusher. The `Pusher.getDefaultStrategy` method that we're redefining is now `Pusher.Runtime.getDefaultStrategy`. Meaning that bit of code hasn't been active for a few weeks now. So I went ahead and removed that - this is something I'd love some review on as I'm not clear on what all this bit did outside the defaults besides setting the path, which we do now another way. 

When trying to use it in the current setup, it caused some exceptions. So if we do need that bit of code, it may need some rewriting.


backspace:
Hey, thanks for looking into this. It was worrisome to me to change the Pusher version without knowing much about how it works, so I made the minimal changes I could that kept it working, knowing that there were perhaps lingering things that were no longer of use. So I’m glad you’ve found some dead code to remove!

I’m at EOD and have unreliable service so I can’t exercise this fully, but I’ll try out the PR deployment tomorrow.

bnferguson:
Absolutely! Appreciate the review! 

> It was worrisome to me to change the Pusher version without knowing much about how it works, so I made the minimal changes I could that kept it working,

Big fan of the safe approach, kinda stumbled onto the dead code myself when I couldn't figure out why path wasn't getting set. :D But feels good to get rid of it! 

> I’m at EOD and have unreliable service so I can’t exercise this fully, but I’ll try out the PR deployment tomorrow.

No worries at all! I just built an Enterprise build with this branch and if it checks out for you tomorrow and passes our QA I'll pull it into the `enterprise-2.2` branch. Thanks so much again for looking at it!

(also I just noticed: yay bower is gone! 🎉 )

backspace:
yeah, removing Bower was my main reason for this update, it’s been a long struggle! haha

It looks like live updates aren‘t working in the org-production or com-production PR deployments:

image

I think the Pusher config needs a default for `path`, or maybe the `wsPath` should only be set when `config.path` is present?

bnferguson:
> or maybe the wsPath should only be set when config.path is present?

Ahh I think this is probably the case. At least this is how the code removed worked.

bnferguson:
Ha, beat me to it. I went with: 

`wsPath: config.path ? `/${config.path}` : null,` 

Your method feels safer in the case that the implementation changes and that `null` ends up overwriting a default. 👍

app/utils/pusher.js
```diff
@@ -55,7 +55,7 @@ TravisPusher.prototype.init = function (config, ajaxService) {
   });
 
   if (config.path) {
-    this.pusher.wsPath = `/${config.path}`,
+    this.pusher.wsPath = `/${config.path}`;
```
bnferguson: image

Beat me to it again. :D

backspace:
hah I made a syntax error though 🙃

As a semi-aside, do you think it’s feasible to have an enterprise PR deployment for the instance that builds the manual? It would be great to have a more-representative environment to test things in. But there’s a lot I’m unaware of with enterprise, so maybe this is infeasible.

app/utils/pusher.js
backspace: haha it’s because I moved the original line down and didn’t notice that 😆

bnferguson:
> do you think it’s feasible to have an enterprise PR deployment for the instance that builds the manual?

Not sure I completely follow. Are you talking about like some of the test environments we have for `travis-web`? Absolutely. Even testing locally I can just point at that API and slanger deployment. However that particular one is on 2.1.X which is based off of code from a year ago. So there may be issues with current PRs going against it. But once it's on Enterprise 2.2 it should work just fine - at least for awhile. Though I don't plan on 2.3 being a year. Maybe a couple months. 

At some point I'd love to have a nightly version running somewhere of `master` builds - which would be easy to point `travis-web` PRs at.

backspace:
yeah, that’s what I mean. It would definitely help me understand the enterprise implications of `travis-web` changes. I look forward to getting something like that going!

The fix looks to be working 😀

image

bnferguson:
Testing it in Enterprise now. (Can see how I'm testing it at the moment here: https://github.com/travis-pro/travis-enterprise/issues/253#issuecomment-372883218 - sorry for the private issue)

Looks like path is getting dropped, and needs to be passed at the point `Pusher.new` is called. It won't be taken into account after. Got a rework going now - testing it and will push it up.

bnferguson:
OK, checks out on Enterprise! And from what I can tell on .com too (via local testing with `TRAVIS_PRO=true ember serve`).

backspace:
gah, I was wondering whether it could be added after the constructor, good catch!

It’s also working on the PR deployments, so 🎉

bnferguson:
Yay! Doing that deploy dance and getting it merged!

clekstro:
@bnferguson @backspace Thanks for wrapping this up!

PR #1590, opened 2018-03-17
Add asset map via ember-cli-ifa

backspace:
This is maybe less exciting now that the team page has been
extracted, but it could bring PR deployments environments
closer to the production ones. To be discussed.


backspace:
This isn’t working as I would expect, I’ll continue on it later.

backspace:
I’ll try this again some time.

PR #1591, opened 2018-03-21
Upgrade test suite to use new testing API

clekstro:
This converts the entire test suite to use the new Ember testing API's as described in [rfc232](https://github.com/emberjs/rfcs/blob/master/text/0232-simplify-qunit-testing-api.md) and [rfc268](https://github.com/emberjs/rfcs/blob/master/text/0268-acceptance-testing-refactor.md), including test helpers.

In addition, it adds `qunit-dom` for better dom assertions.

Currently, the mirage server is started automatically and available via `this.server` in all rendering tests. I'm not sure if we should keep this behavior or not, as it goes against our desire to be explicit and also probably carries a performance penalty. I may address that in a future PR, however.

I have used both available codemods ([ember-qunit-codemod](https://github.com/rwjblue/ember-qunit-codemod) and [ember-test-helpers-codemod](https://github.com/simonihmig/ember-test-helpers-codemod)) as much as I could, though as you can see from the commit history it was not possible to simply convert the test suite en masse.


clekstro:
This does not complete the transition, but does make significant progress toward cutting over. Merging in current progress for now.

PR #1598, opened 2018-03-28
Add support for GitHub Apps-managed repositories

backspace:
This is branched off #1592 with the hopes that that might be mergeable before this, but we shall see. It’s currently quite uggers but at least it’s a start!


backspace:
I’m going to close this in favour of a branch with `staging` in the name ugh

PR #1602, opened 2018-04-04
Change to Chrome beta

backspace:
This fixes the test failures that weren’t happening locally.

I don’t know why `stable` is yielding Chrome 62 but ¯\\\_(ツ)\_\/¯


backspace:
I fixed the failures in #1605 instead.

PR #1604, opened 2018-04-05
Update profile page for GitHub Apps

backspace:
Still a long way to go.

ETA this is too unwieldy and I’ll be breaking it up, but not for a while.


backspace:
I see thanks to Percy that it wasn’t a perfect breaking-up, as some images were missing for the landing page, but it’s GOOD ENOUGH

PR #1606, opened 2018-04-06
Fix services page to integrations tab

dmytrostriletskyi:
For now services page (in repository settings) is available from a `Integrations & services` side bar's item instead of `Webhook`. So fix name of the tab.

The current documentation guide:

screen shot 2018-04-06 at 5 06 54 pm

The real repository settings looks like:

screen shot 2018-04-06 at 5 08 22 pm

Also the same not actual documentation in [there](https://github.com/travis-ci/travis-web-index/blob/a36e9610f4253d7577dd96484a22f79bd6bc6f59/app/templates/getting-started.hbs). If you accept the pull request, i will provide changes for this case too.

### DIFF
--- a/app/templates/getting-started.hbs
+++ b/app/templates/getting-started.hbs
@@ -37,7 +37,7 @@
               <p>To start a build, perform one of the following:</p>
               <ul>
                 <li>Commit and push something to your repository</li>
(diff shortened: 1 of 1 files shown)


backspace:
Hello, thanks for this!

I’m going to merge this, but it won’t be deployed until next week.

You can safely ignore `travis-web-index`, which won’t survive in its current form.

PR #1607, opened 2018-04-06
Don't show Beta Features page in Enterprise

bnferguson:
In the 2.2 build of Travis Enterprise `travis-web` now shows the Beta Features page. But this is always empty as that's not a thing we support in Enterprise (and isn't in 2.1 or 2.0): 
monosnap 2018-04-06 14-23-09
image

This PR creates a feature flag to fix this and hides the navigation to this page from the accounts page.


joshk:
Great catch Brandon!

It would be awesome to expose the Dashboard somehow 🤔

bnferguson:
> It would be awesome to expose the Dashboard somehow 🤔

Yeah! Though probably not for 2.2 since we're on the last legs of QA there. Love to get this out next week. Since it _should_ be a fairly simple feature flag change I'm up for looking at it in a point release! (or just pushing for 2.3 sooner than later). 

@clekstro thanks so much with all your help on this!

PR #1613, opened 2018-04-11
Update switch appearance

backspace:
This is a place to store off the full conversion of all (?) switches in the application to a new look because #1604 has grown terribly large and I’m thinking about how I’ll break it up.


backspace:
I guess I might as well just do this the right way hehe

PR #1619, opened 2018-04-12
Add Intercom

backspace:


app/app.js
```diff
@@ -67,6 +75,12 @@ const App = Application.extend(Evented, {
         return HS.beacon.identify(userParams);
       });
     }
+  },
+
+  setIntercomUser(user) {
+    this.get('intercom').set('user.name', user.name);
+    this.get('intercom').set('user.email', user.email);
+    // FIXME createdAt?
```
joshk: we can use `user.first_logged_in_at` i think, and assign it to `createdAt`

backspace: 👍🏻 thanks, I’ve added that now. I’m having a hard time getting anything to build locally so I’ve just pushed and will see how it works on the PR deployment.

app/services/auth.js
```diff
@@ -161,7 +161,7 @@ export default Service.extend({
     if (proVersion && get(config, 'intercom.enabled')) {
       this.get('intercom').set('user.name', user.name);
       this.get('intercom').set('user.email', user.email);
-      // FIXME createdAt?
+      this.get('intercom').set('user.createdAt', user.last_synced_at);
```
joshk: ohhhhh, no no no, `first_logged_in_at`

backspace: oops thanks, I didn’t get enough sleep 😳

backspace: I finally got it running locally, looks like it’s `synced_at`:

image

backspace: er wait… `first_logged_in_at` doesn’t exist, or anything like it

backspace: (for observers: `created_at`, obvsy; like I said, underslept 😆)

PR #1631, opened 2018-04-26
Update profile page to use V3, including subscription status

backspace:
This is awaiting the proper functioning of `GET /subscriptions`.


backspace:
itshappening.gif

PR #1634, opened 2018-04-30
Add pagination query param name configurability

backspace:


backspace:
This is extracted from #1604. Homestretch!

PR #1639, opened 2018-05-01
Change token buttons to not switch routes

backspace:
It works for the view button but I don’t know how to address it for the copy button.


backspace:
I’m going to let go of getting the copy button to behave properly for now so I can merge and deploy this.

PR #1645, opened 2018-05-06
Fix SVG styles conflict (example)

ivanvotti:
This is just an example of how [the problem](https://github.com/ivanvotti/ember-svg-jar/issues/68) can be fixed with SVGO. The new SVGO version has slightly changed plugin defaults, so I had to add `{ removeViewBox: false }` to prevent viewBox deletion.


ember-cli-build.js
```diff
@@ -2,6 +2,7 @@
 
 const EmberApp = require('ember-cli/lib/broccoli/ember-app');
 const Funnel = require('broccoli-funnel');
+const SVGO = require('svgo');
```
ivanvotti: This way we use the latest SVGO in SVGJar.

package.json
```diff
@@ -80,7 +80,7 @@
     "ember-route-action-helper": "^2.0.3",
     "ember-sinon": "^1.0.1",
     "ember-source": "^2.18.2",
-    "ember-svg-jar": "^1.1.1",
+    "ember-svg-jar": "ivanvotti/ember-svg-jar#custom-svgo-module",
```
ivanvotti: This branch has the custom SVGO module feature, that will be available in the next SVGJar release.

ember-cli-build.js
```diff
@@ -56,6 +57,20 @@ module.exports = function () {
...
+    svgJar: {
+      optimizer: {
+        svgoModule: SVGO,
+        plugins: [
+          { removeViewBox: false },
+          {
+            inlineStyles: {
+              onlyMatchedOnce: false,
+              removeMatchedSelectors: true
+            }
```
ivanvotti: `inlineStyles` is the SVGO plugin that does the trick.

ember-cli-build.js
```diff
@@ -56,6 +57,20 @@ module.exports = function () {
         'public/images/stroke-icons',
         'public/images/svg'
       ]
+    },
+    svgJar: {
+      optimizer: {
+        svgoModule: SVGO,
```
ivanvotti: Injecting the latest SVGO package to use it in SVGJar instead of the old version.

backspace:
Thank you! This is really above-and-beyond in terms of responding to my issue. I have some significant work happening elsewhere in this repository that I need to get to, but I’ll keep this around so we can move to using the `svg-jar` component for all our SVGs. I look forward to the new version!

PR #1655, opened 2018-05-09
Add explicit rejection for unsupported browsers

backspace:
We don’t have a real browser support policy but this could address one if
we did! To be determined.

Thanks to those in this thread for ideas:
https://discuss.emberjs.com/t/best-practices-for-browser-version-detection/14719/4

I suspect the PR deployment won’t behave as expected, but it does locally.
We’ll see!


backspace:
wow good to know, thanks! So fast haha

My true dream would be to magically connect this to `targets.js` but it’s probably not worth it for something that changes so infrequently.

backspace:
This is quite minimal but will suffice to reject IE < 11, which is satisfactory to me for now. 🎉

PR #1666, opened 2018-05-14
Remove build stages beta banner

backspace:


backspace:
It looks like this is superseded by #1697.

PR #1671, opened 2018-05-17
Fix issue of localStorage getting cleared too often for the Enterprise License Banner

bnferguson:
Fixes https://github.com/travis-pro/travis-enterprise/issues/300 (Pardon the private issue)

Summarizing that issue, essentially the license banner comes back after a refresh even if you have dismissed it. Digging we found some bugs in one of the flows of logic where some properties got renamed: 

> Turns out it wasn't the logic that was off, but the get statements... `exceeding` and `almostExceeding` return as `undefinded`. It seems the proper properties are `exceedingSeats` and `almostExceedingSeats`? 
>
>As these are always false-y the localstorage got cleared everytime. Testing a fix to this right now. So hopefully that's the problem. 

This change has already been made in this PR. What makes this PR WIP and the feedback I need is it seems the enterprise banner component fires twice (once with the state in `destroying` and once as `inDOM`), and the `expiring` computed property comes back first `false` then `true`. So it ends up getting cleared in: https://github.com/travis-ci/travis-web/blob/master/app/components/enterprise-banner.js#L36-L38

So the local storage is getting cleared every refresh. 

My big questions are: 

1. Should this component be firing twice?
2. Why does the value of `expiring` change between it firing? Is this a promise issue or something with the API call?

### DIFF
--- a/app/components/enterprise-banner.js
+++ b/app/components/enterprise-banner.js
@@ -33,12 +33,17 @@ export default Component.extend({
         isTrial: response.license_type === 'trial',
         isPaid: response.license_type !== 'trial'
       });
-      if (!this.get('expiring')) {
-        this.get('storage').removeItem(this.get('lsLicense'));
-      }
-      if (!this.get('almostExceeding') && !this.get('exceeding')) {
-        this.get('storage').removeItem(this.get('lsSeats'));
-      }
+      console.log(this.get('expiring'))
+      console.log(this.get('daysUntilExpiry'))
+
+      console.log(this.get('almostExceedingSeats'))
+      console.log(this.get('exceedingSeats'))
+      // if (this.get('daysUntilExpiry') && !this.get('expiring')) {
+      //  this.get('storage').removeItem(this.get('lsLicense'));
+      //}
+      // if (!this.get('almostExceedingSeats') && !this.get('exceedingSeats')) {
+      //   this.get('storage').removeItem(this.get('lsSeats'));
+      // }
     });
   },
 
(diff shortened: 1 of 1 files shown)


backspace:
It would be somewhat of an overhaul, but my suggestion for dealing with this involves using the `PromiseProxyMixin` to wrap the API call in a promise and avoid even trying to show the banner before the call is complete. The `didInsertElement` thing is somewhat un-Ember-ish as it’s more procedural than declarative; I didn’t want to be annoying when this was originally being created, but this is the kind of mess one gets into when ignoring asynchrony.

You can see an example [here](https://github.com/travis-ci/travis-web/blob/768796e91946575415c147c2921cd418fccac756/app/components/build-messages-list.js#L17) where I used `PromiseProxyMixin` to pass asynchronously-fetched data to the [template](https://github.com/travis-ci/travis-web/blob/c0d3152548b3cacdc0131adb0af22187eab7bd37/app/templates/components/build-messages-list.hbs#L1). After the API call, you could construct a plain Javascript object that contains all the flags that are currently computed properties, then the template can operate on same things, but you can add an `{{#if enterpriseLicenseCheck.response}}` to wait for the data to have been fetched before showing anything. (`response` is a bit of a hackish name… oh well?? haha)

Let me know if this makes sense!

bnferguson:
Also this may end up being too much for 2.2.1 - as there's a really nasty thread leak in it - may need to either find a quicker stop gap or just not include this as one of the fixes and mark it as a known issue. 

@danishkhan - thoughts on the timing of this? I'm gonna look at maybe just removing the bit that clears the local storage for now so it sticks. It'll be less smart until we fix it correctly but also less annoying for customers.

PR #1674, opened 2018-05-23
When Enterprise has publicMode turned off, generate tokens for status images. 

bnferguson:
Fixes: https://github.com/travis-pro/travis-enterprise/issues/301 (pardon the private issue)

From that issue: 

>Looks like build images for public repositories break when Travis Enterprise is not in Public Mode (we're never in public mode right now - this is so we don't leak projects, we'll introduce it in 2.3 as a config option). 

This PR aims to fix this issue by introducing an env var that can be set that mimics `public_mode` in other projects. This var was introduced when we started merging .org and .com to determine if it should show "public" projects on github or not (Enterprise we default to not). So in this case when `public_mode` is `false` we'll generate tokens for all of the images, and not just the private repositories.


bnferguson:
> So public mode defaults to true if the flag isn't passed?

Yes. This is the one thing I wasn't completely sure on, should we default to true. So I'll go through my thinking here to see if it makes sense outside my head. :D 

This is something that comes from the .com/.org merge. In Enterprise we still treat everything as not visible to unauth'd users even if it's a public repo (this will be configurable, based on the public_mode flag in an upcoming version). So if the repository is "public" (on GitHub Enterprise for instance), you still need to be authenticated to see the logs. 

In hosted, public is public. And public mode is turned on. But in enterprise it may be on or off, so we set it and if it's set we use it. If it's not set, we default to true (eg: hosted).

bnferguson:
Also it looks like I need to add this to the `config.ru` as well to get it at runtime? Like so in the backport: https://github.com/travis-ci/travis-web/commit/3704da98943036a0d088bab65f71f3ba7553e1cc

Is this correct? Anywhere else?

backspace:
hmm yes I forgot about `waiter`, as I often do… I think that’s the only place. 😳

bnferguson:
If someone has a moment can I get a re-review? Made a few of changes to get this working. This is confirmed as fixing 2.2.0's issue - then I cherry picked the changes over.

config/environment.js
```diff
@@ -150,6 +150,12 @@ module.exports = function (environment) {
       }
     }
 
+    if (process.env.PUBLIC_MODE !== null && process.env.PUBLIC_MODE == 'false') {
```
backspace: it’s not really a big deal but `!== null` is redundant I believe? I’d write it this way:

`if (process.env.PUBLIC_MODE === 'false')`

bnferguson: 🤦‍♂️  You would be correct!

bnferguson:
> Are you able to test this out in a real environment? 

Yup. This is running on the `enterprise-2.2` branch successfully with the waiter bits. 👍

> The whole waiter business is a perpetual confusion to me so I can’t confidently say it’ll work.
It also confuses me, I feel like I relearn about it every 6 months when I need to add an env var or change one in Enterprise.  Would be good to see if we can get around it at some point without requiring a rebuild of the project after settings changes (eg, every reboot of an Enterprise server, that may not have access to the internet). Which I assume is why it exists? Not actually super familiar with how `travis-web` runs in hosted.

bnferguson:
> yes, I think that’s the major reason for its existence, but it predates my joining the company and I’m not sure I’ve ever gotten an explanation hehe

haha, right on. It feels like an older artifact somehow.

What do we run it all on in hosted? As in what serves the files? Also, I guess all of the config happens at build/deploy time which is the snag we have in Enterprise as rebuilding it all each time a config changes may not be possible (since some of these instances don't have connections to the internet, plus it makes restarting take much longer). 

Not sure how possible it is, but it'd be great to get Enterprise as close to hosted as possible.

backspace:
Being able to update the configuration without a full rebuild is definitely powerful! It may be that we want to keep that core functionality, but I think there’s a lot of other stuff happening that we could strip out. Some day…

How are assets served for Enterprise?

bnferguson:
We have nginx handing off to waiter as a reverse proxy (which runs on port 5300 internally): https://github.com/travis-pro/travis-enterprise/blob/master/src/main/etc/nginx/sites-enabled/travis.conf.tpl (pardon the private link, just an nginx config, which is actually viewable in our Enterprise container under `/usr/local/travis/etc/nginx/sites-enabled/travis.conf`) 

Actually just now noticing that static assets aren't directly served by nginx. That's kinda surprising, it's _much_ better at serving those than an application server (then again later on we'll be separating this out to work on kubernetes and the like so nginx may be going away in favor of trafik or similar)

backspace:
oh yes hmm… for hosted the `index.html` is served by Ruby on Heroku and the assets come from Fastly.

bnferguson:
Hmmm, so Ruby wise, is it also using waiter on Heroku? Or something else?

PR #1678, opened 2018-05-24
Update footer

backspace:
The footer should now be always visible!

I expect I’ll find some problems with this still.


backspace:
The footer shouldn’t be full-width on the repository 404 page but I can’t address that at the moment.

PR #1691, opened 2018-06-01
SVG style conflict internal

backspace:
This is an insider version of #1645.


backspace:
#1645 is already merged 🙃

PR #1697, opened 2018-06-12
Job names

svenfuchs:
This adds a job's name if it is present in the job's config. We are considering making this a first class attribute on jobs and add it to the api payload, but that is still to be decided. Opening this here so we can look at it.

I haven't added the name to the job view because that would need input from creative.

This replaces the "language" section with the given name:

image

Alternatively we can replace both "language" and "env" if a name is present:

image

---

~Also, I think I'd need help with figuring out why in the world this fade-out thing is visible on the name:~

image

~It is not visible for a job that doesn't have a name:~

image

~I thought it's caused by the language icon (I haven't added an icon for the name), but it's not ... even if I add the icon back, the fadeout is still visible.~

Edit: @paulthegeek fixed this ❤️

### DIFF
--- a/app/components/build-message.js
+++ b/app/components/build-message.js
@@ -40,6 +40,10 @@ export default Component.extend({
     return `downcasing <code>${escape(args.value)}</code>`;
   },
 
+  duplicate_names(args) {
+    return `duplicate job names: <code>${escape(args.value)}</code>`;
+  },
+
   edge(args) {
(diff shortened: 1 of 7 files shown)


paulthegeek:
@svenfuchs I fixed the fade-out part. I'm not sure if there was anything else that needed to be added or fixed. Let me know if there is anything I can help with.

PR #1704, opened 2018-06-21
Switch from ember-browserify to ember-auto-import

ef4:
This was a good test case to exercise [ember-auto-import](https://github.com/ef4/ember-auto-import) in a real application. It went pretty smoothly! 

The benefits here are:
 -  being able to simply `import` directly from NPM, similar to what ember-browserify could do but without the custom `npm:` syntax
 - moderately faster initial build (about 10%)
 - faster incremental build (about 30%)

There are only two minor points of interest where changes were needed:

 - the app was importing raven-js but didn't list it as a dependency. It was relying on hoisting from another package (which is not really guaranteed to work, but usually does). ember-auto-import is stricter and won't let you import things you don't list as dependencies.

 - yamljs has an open issue for having a `require('fs')`, which I worked around by providing a snippet of webpack config. Before with browserify, this would have pulled in browserify's fs browser implementation, which you almost certainly didn't really want or need.

I measured and there was no impact on bundle size (webpack and browserify do equally well on the set of libraries you're pulling in).


backspace:
This is amazing, thanks so much! The faster build times will be excellent for us. I should be able to look into this more next week.

PR #1732, opened 2018-08-01
Update Ember and Ember Data to 3.3.*?

backspace:
This is the other part of #1729 not covered by #1731. So far I’m at Ember 3.2.2 and have everything passing (locally! 🤞🏻), we’ll see how far I can get with it and Ember Data.


backspace:
On 3.3.1:

image

It’s confusing to me that the filter thing shows up when I didn’t change Ember Data at all. Rather than get back into that quagmire at the moment, which I struggled with earlier this week, I’m going to merge this and come back later for more incremental changes.

PR #1736, opened 2018-08-03
Update addons

backspace:
I’m targeting addons that produce the deprecation warning about ember-cli-babel.


backspace:
I updated some others while I was at it.

PR #1737, opened 2018-08-03
Pj 135 top bar content

paulthegeek:


joshk:
This looks great!

I think we should either remove `Dashboard` and ship this, or hold off until the Dashboard is finished.

@acnagy @ctrlaltjustine thoughts?

ctrlaltjustine:
I don't think top bar should be blocked by dashboard and don't have a problem with that link being removed until dashboard work is finished.

ctrlaltjustine:
That said, there is still some [feedback that needs to be addressed before it's ready to be shipped](https://github.com/travis-pro/design-tasks/issues/135#issuecomment-411430298).

joshk:
Sounds great to me @ctrlaltjustine , we can ship without the Dashboard

PR #1744, opened 2018-08-10
Adjust builds for more speed

backspace:
I’m going through [this](https://github.com/ember-cli/ember-cli/blob/master/PERF_GUIDE.md).

### DIFF
--- a/ember-cli-build.js
+++ b/ember-cli-build.js
@@ -4,6 +4,26 @@ const EmberApp = require('ember-cli/lib/broccoli/ember-app');
 const Funnel = require('broccoli-funnel');
 const SVGO = require('svgo');
 
+var fs    = require('fs');
+var path  = require('path');
+var types = require('node-sass').types;
+
+var svg = function (buffer) {
+  var svg = buffer.toString()
+    .replace(/\n/g, '')
+    .replace(/\r/g, '')
+    .replace(/\#/g, '%23')
+    .replace(/\"/g, "'");
+
+  return '"data:image/svg+xml;utf8,' + svg + '"';
+};
+
+var img = function (buffer, ext) {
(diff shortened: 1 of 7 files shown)


artursmirnov:
I have also noticed, that the app uses `ember-composable-helpers`. [Whitelisting](https://github.com/DockYard/ember-composable-helpers#configuration) only the helpers that we use may help to improve build performance as well.

backspace:
oh nice, thanks, I’ll check that out when I can come back to this. Sadly most of my changes haven’t made much of a difference for build times, but there are some changes and updates we should do anyway.

artursmirnov:
On one of my previous apps the `ember-cli-mirage` made a huuuge affect to build time. When I installed it and created some factories, the build time increased like twice! And I also heard the same from other ember devs. Probably it makes sense to ask Sam at Inside EmberMap about what can affect the build performance and maybe he can give some advice on improvement.

In fact, building `travis-web` with mirage takes 23s for me:
```bash
$ time ember b
Built project successfully. Stored in "dist/".
ember b  23.11s user 3.48s system 107% cpu 24.827 total
```
Without mirage it takes 19s:
```bash
$ time ember b
Built project successfully. Stored in "dist/".
ember b  19.36s user 2.78s system 111% cpu 19.898 total
```

Also production build complains about minification of `vendor.js`, which means it's too big. The way to improve it may be to sanitize the addons that we use. And, of course, the upcoming tree-shaking would help with that.

backspace:
yeah, the Mirage update I made in this branch actually did speed up the build significantly. I’m starting to wonder whether we should stop using it entirely because of the number of times I’ve had problems with slight differences between the true and Mirage APIs, most recently [here](https://github.com/travis-ci/travis-web/pull/1751). But that would be a significant undertaking.

I generated some build time graphs and Mirage was indeed quite slow.

PR #1749, opened 2018-08-16
Fix styles for disabled profile settings button

artursmirnov:


backspace:
(we talked more about that elsewhere, seems we can’t reproduce it outside Percy) so “Squash and merge” at will! 😀

artursmirnov:
It would be really handy if Percy could give some more insight on the environment. Like browser version, OS, installed fonts, zoom level, resolution, screen density, etc. Also it would be nice to have an ability to create a screenshot manually with arbitrary parameters. We should submit a feature request to them 👍

PR #1750, opened 2018-08-16
Fix build job list alignment issue

sorta:
Previously you could see this issue when viewing travis-ci/gimme.

Looks like it was due to a quirk of flexbox children with `white-space: nowrap;`. By default flex items won't become smaller than their contents. Setting `min-width: 0;` on the parent resolves the issue.

### DIFF
--- a/app/styles/app/layouts/jobs.scss
+++ b/app/styles/app/layouts/jobs.scss
@@ -238,11 +238,11 @@
     }
 
     .job-env {
-      flex: 0 0 21em;
+      flex: 0 0 21.5em;
     }
 
     .job-lang {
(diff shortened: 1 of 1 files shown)


sorta:
@backspace nice catch, yeah I think we might as well fix that here. In that case, it was set widths for some things on xl viewports that didn't take into account that one of the items has a negative margin on it. So I updated some of those widths accordingly.

backspace:
Excellent, it’s all looking good to me now, thank you! Do you want to merge or shall I? haha

sorta:
Hah, I'll do the merge. Thanks!

PR #1756, opened 2018-08-21
Refactor travis-switch to make it reusable

artursmirnov:
For https://github.com/travis-ci/product/issues/21 I need a reusable `switch` component, so I refactored it a bit and made other dependent component re-use the existing one.


app/components/travis-switch.js
```diff
@@ -1,28 +1,28 @@
...
+  active: false,
+  disabled: false,
 
-  @or('target.active', 'active') _active: null,
+  @computed('active')
+  'aria-checked'(active) {
+    return active ? 'true' : 'false';
+  },
+
+  onToggle() {},
```
backspace: In terms of developer ergonomics, what do you think of adding something like this here?

```
onToggle() {
  throw new Error('Must supply an implementation for onToggle');
}
```

It might slightly ease the work of some future person using the component 😀

artursmirnov: Yeah, I agree, totally makes sense 👍 Will push a fix shortly

artursmirnov: @backspace I added and [abstractMethod util](https://github.com/travis-ci/travis-web/pull/1756/files#diff-77c6b499af98e4d8e6c7c30d046fa488) for such cases

PR #1760, opened 2018-08-24
Email settings and unsubscribe

artursmirnov:
Issue: https://github.com/travis-ci/product/issues/21


backspace:
Looking good so far! That `paper-block` component would be useful elsewhere in the `github-apps-invitation`.

artursmirnov:
@backspace another bunch of changes :) Could you please take another look once you have a moment? Basically, UI is done. Only API integration and tests left. So it's pretty much close to the final version.

backspace:
Good to see, I guess there are some failures that preclude a PR deployment, let me know if you want to team up on figuring those out. I wonder if this is getting big enough that it would make sense to extract the `paper-block` and `loading-screen` components into their own PRs so this one can be more focused on the particular changes? Those could be merged before this one is ready.

I think you can safely delete the auto-generated tests, probably the only necessary one would be an acceptance test. (We’ve only partially migrated to application tests.) [Here](https://github.com/travis-ci/travis-web/blob/0068883a0f259a7c1a9c1cd8890fc5d6fb9afb41/tests/acceptance/repo/settings-test.js#L199-L204)’s an example where I tested for server API calls in a test, might be helpful. We use [page objects](http://ember-cli-page-object.js.org/docs/v1.14.x/) for more-readable assertions that keep selectors from being scattered across the tests.

backspace:
I thought of another thing, that some of the ESLint errors you ran into might be addressed automatically by setting up your editor with something that auto-corrects based on the rules present in the repository. I’m using Atom lately and I have it working that at least some things are fixed on save, let me know if you want to hear more about it!

artursmirnov:
Yeah, good point, thanks!

I have `eslint` integrated to my editor (I use `neovim`) and it highlights errors. It doesn't fix them automatically, but there is a shortcut for this. 

Talking about pull-request failures, it seems like `eslint` complains mostly about autogenerated tests, which I didn't approach yet. And there are some failed real tests as well. I'm going to fix those and add new ones today.

app/templates/profile.hbs
```diff
@@ -1,3 +1,13 @@
-{{#travis-layout layoutName="layouts/profile" class="profile-view"}}
-  {{outlet}}
-{{/travis-layout}}
+{{#if showLoading}}
```
backspace: Is it possible to use the “loading substate” for this kind of thing instead?

app/templates/profile.hbs
```diff
@@ -1,3 +1,13 @@
...
+  {{#travis-layout layoutName="layouts/center"}}
+    {{loading-screen center=true margin=true}}
+  {{/travis-layout}}
+{{else if showProfile}}
```
backspace: I’m guessing this kind of pseudo-routing work is because the URLs for unsubscribing are nested here but you don’t want all the profile things to show here. I wonder if it’s worth noting that somewhere? Maybe both here and in the controller?

backspace:
Something mysterious is that when I left the PR deployment for `org-production` option for a while, these eventually showed up despite me having done nothing related to email with them:

image

This is likely a result of build events coming in via Pusher, they’re likely missing some key that you’re looking for to control what displays here.

There’s an example [here](https://github.com/travis-ci/travis-web/blob/48daa8097d966eb75a41ace1f09f665578757af4/tests/acceptance/home/with-repositories-test.js#L120-L124) of where I simulate a Pusher event in an acceptance test, it might be helpful in adding assertions to test for this not happening. You can see Pusher events coming in by looking at the “frames” of the websocket in the network area of browser developer tools.

Let me know if you want to pair on digging into this, it can be pretty obscure!

app/templates/profile.hbs
artursmirnov: No, unfortunately, this kinda hack is necessary, because until the entire path is rendered the `router.currentRouteName` is different from `account.settings.unsubscribe` and there's no way to figure out if it's an unsubscribe route or not. This causes the profile layout to be rendered and only then replaced by `email-unsubscribe`. Thus, it's necessary to show loading screen until the entire path is resolved, not only the `profile` part.

artursmirnov:
All feedback and the issue with Pusher is fixed

app/templates/components/email-unsubscribe.hbs
```diff
@@ -0,0 +1,95 @@
...
+          Please confirm that you would like to unsubscribe from build emails for the
+        {{else}}
+          You are unsubscribed from build emails for the
+        {{/if}}
+        <span class="repo-slug repo-slug--bold">{{repo.formattedSlug}}</span>
+        repository.
+
+        {{#if showConfigNote}}
+          If you’re still receiving emails after unsubscribing, please check your
+          <a target="_blank" href="https://github.com/{{repo.slug}}/blob/master/.travis.yml">.travis.yml</a>
```
backspace: it’s maybe not a big deal but when I tried this just now on a test repository I got a 404 for [this link](https://github.com/backspace/travixperiments-redux/blob/master/.travis.yml) because `primary` is my default branch and `master` doesn’t exist.

artursmirnov: We had a discussion about this in [#project-unsubscribe](https://travisci.slack.com/archives/CC97UDR8R/p1536230141000100). In short, this page is not aware of a branch name or commit. If we want to make it more flexible, we have to provide a branch name in query parameter. But it may be not enough, because config may be different on a commit level. So pointing to default branch is not a rock-solid solution, but it would work in most cases.

Although, putting the `repo.defaultBranch` instead of `master` may be a good idea 👍 

What do you think?

backspace: yeah, I think even with that limitation that using the default branch is a better idea, thank you 😀

backspace:
Looks good, I wonder if it might be nice to have some kind of scroll-down/highlight situation to show the section of the page the user cares about? When visiting https://as-email-settings.test-deployments.travis-ci.org/account/preferences/unsubscribe on my current display, the switch in question is actually off the bottom of the screen.

artursmirnov:
@backspace nice suggestion, thanks! It's live.

artursmirnov:
And here's the commit to review: https://github.com/travis-ci/travis-web/pull/1760/commits/91c9b40e3461b772add424baad8888e53415d30a

backspace:
Nice, `ember-concurrency-scroll`, I hadn’t seen that before!

PR #1762, opened 2018-08-27
Add cron jobs for testing Ember betas

backspace:
I think this approach may not work 😞


backspace:
It seems to?!

The push build is showing here as pending but is [actually complete and passing](https://travis-ci.org/travis-ci/travis-web/builds/421655578) so I’m going to merge.

PR #1767, opened 2018-08-29
Fix abstract method test

artursmirnov:
It was failing for me locally. Not sure how it worked in CI.


artursmirnov:
Change is really minor, so I'm merging the PR myself, @backspace please take a look afterwards.

PR #1778, opened 2018-09-05
Update stage times alignment in wide windows

sorta:


sorta:
Ok cases like https://so-stage-time-alignment.test-deployments.travis-ci.org/backspace/travixperiments-redux should be aligning correctly now, and also the previous fix for cases like https://travis-ci.com/ezyang/pytorch-unattached/builds/81412631 should still be working.

PR #1783, opened 2018-09-10
Fix log blanks

artursmirnov:
Issue: https://github.com/travis-pro/team-teal/issues/2782


artursmirnov:
Thanks for checking! It would be nice to refactor this part someday to Ember Components, which would greatly simplify the logic. And in scope of that task we could add testing agains multiple different and complex logs.

backspace:
I think the reason this doesn’t use more Ember-y facilities is that it can be rendering up to 10k lines of log that will remain static, so the overhead for creating dynamically-controlled elements isn’t worth it. Maybe things have changed to be efficient enough since Glimmer 2 etc though…??? ¯\_(ツ)_/¯

edmorley:
Hi! Is it possible this change caused travis-ci/travis-ci#10133?

backspace:
It seems quite likely! We’ll revert this change for now and try to find another way. Thanks for letting us know about this, we can use your example to refine the solution.

PR #1784, opened 2018-09-10
Update Ember CLI to 3.4

backspace:
This skips template linting for now.


backspace:
This showed me that maybe we don’t have too far to go before we can remove jQuery as a dependency! But I’ll save that for another time.

PR #1792, opened 2018-09-12
Repo activate button for github apps

sorta:


joshk:
Oh, one thing to note, is that `.org` and Enterprise don't use GitHub Apps, does this fallback to the old way for them?

sorta:
Yes it should fallback to the old way for `.org`. Not sure about Enterprise, I'll do some testing to verify.

backspace:
I think [this line](https://github.com/travis-ci/travis-web/pull/1792/files#diff-096e8a56e2f38bc2b06dd1206dc06eccR10) that checks for `config.githubApps`, makes it work 😀

joshk:
Ahhhhhh, awesome! Super High 5!

sorta:
Ok I tested and confirmed that `org` and Enterprise fall back to the old method. I've also added in the changes to the manage button.

joshk:
awesome! well, feel free to ship at your convenience! :)

sorta:
Excellent! Will get that going :smiley:

PR #1806, opened 2018-09-19
Add touch icon support

clekstro:
This should address https://github.com/travis-pro/team-teal/issues/2818.

Previously, the icon looked like this:
before

Now, after generating the icon (and then adjusting the favicon in the second version), we get this:

image-1

I've purposely left off some icons to avoid inflating the size of the `index.html` (firefox os, opera, yandex iirc), but we can discuss that further if there are concerns.


ember-cli-build.js
```diff
@@ -54,6 +55,45 @@ module.exports = function () {
...
+          //
+          android: true,              // Create Android homescreen icon. `boolean` or `{ offset, background, mask, overlayGlow, overlayShadow }`
+          appleIcon: true,            // Create Apple touch icons. `boolean` or `{ offset, background, mask, overlayGlow, overlayShadow }`
+          appleStartup: true,         // Create Apple startup images. `boolean` or `{ offset, background, mask, overlayGlow, overlayShadow }`
+          coast: false,               // Create Opera Coast icon. `boolean` or `{ offset, background, mask, overlayGlow, overlayShadow }`
+          favicons: true,             // Create regular favicons. `boolean` or `{ offset, background, mask, overlayGlow, overlayShadow }`
+          firefox: false,             // Create Firefox OS icons. `boolean` or `{ offset, background, mask, overlayGlow, overlayShadow }`
+          windows: true,              // Create Windows 8 tile icons. `boolean` or `{ offset, background, mask, overlayGlow, overlayShadow }`
+          yandex: false               // Create Yandex browser icon. `boolean` or `{ offset, background, mask, overlayGlow, overlayShadow }`
```
artursmirnov: 😟

clekstro: @artursmirnov Could you elaborate? I specifically asked for feedback on what to support, so if you're concerned about leaving something off, I'm happy to hear it :)

artursmirnov: @clekstro I think that your choice totally makes sense 👍 I'm using Yandex Browser myself on both mobile and desktop, but I think I'm pretty much unique on this in Travis audience. It just doesn't worth adding because of very low market share :)

clekstro:
I was seeing some weird behavior in the PR envs. Neither iOS nor Android was working as expected. However, when running in production mode locally, I was able to see that things work on both.

Here's what Android would look like:

screen shot 2018-09-19 at 16 49 07

I know we've had issues (e.g. build badges) with those envs in the past, so I'd be curious if anyone has concerns about this?

I'm a little hesitant to merge for a few other reasons:

1. The generated `index.html` is generating links that aren't correct -- relative paths that are interpreted by the app as missing repositories when requested.

2. The generated icons appear in the `public/` folder directly (we might be able to change this, not sure), and while it generates android icons, they're not actually linked in the `index.html`. This is very confusing...

3. I'm not sure what the performance impact of this is (if at all). But development build times were definitely slower (though we could probably only do this in production somehow). 

4. We could probably do more research and just add the images we'd like/the head html ourselves and avoid an addon altogether. I'm not sure how often we'd want to change these things, so I'd be curious how we feel about the tradeoff...

/cc @backspace @artursmirnov @sorta @paulthegeek

artursmirnov:
@clekstro seems like the addon is skipping the `assets` part in generated image paths. It may work in the apps that don't use wildcard routes, but Travis-web does use it. My guess, this is the reason of not loading images.
Android takes image urls from manifest file, that's why they are not in `index.html`
I would give this addon a try https://emberobserver.com/addons/ember-web-app. It works in a similar way, but does not generate images in build time, so it should not affect builds performance.

clekstro:
@artursmirnov I've made the change you've suggested and it appears to have cleaned things up significantly (less unexpected/invalid HTML output). I've looked at this on both iOS and Android and things still look good.

Clicking on some of the relative links directly still leads to bad behavior, i.e.:
```html
<link rel="manifest" href="/manifest.webmanifest">
```
but it doesn't seem to impact the what the user sees afaict.

What are your thoughts on merging now and addressing this (if possible) in a subsequent PR?

artursmirnov:
@clekstro I think this sounds great! Thanks for working on this, can't wait to try it :)

PR #1812, opened 2018-09-24
Account url changes

artursmirnov:
Issue: https://github.com/travis-ci/product/issues/44


backspace:
I noticed just now that switching between organisations causes the whole display to flash (but _not_ switching to the user). Maybe that‘s an okay price to pay for this improvement, but it does feel a bit more jarring than the current implementation. Do you think it would be too much of a hassle to change the hierarchy so that doesn’t happen?

flashing

artursmirnov:
@backspace thanks for this catch! This flash is caused by the re-render of navigation due to different routes. Although, seems like I found a way to make this re-render instant. Please take another look once you have a moment.

backspace:
I love it!!!! Thanks for figuring that out 🎉

joshk:
This is fantastic!

Just to confirm, all redirections work as expected?

artursmirnov:
@joshk yes, all cases that I could imagine, and here's the test that covers redirects: https://github.com/travis-ci/travis-web/blob/39dc0085a652fb09bb0cc34db7495d6c62910d19/tests/acceptance/profile/redirect-test.js

joshk:
@artursmirnov 👍 on deploying

I think we should give Support a heads up. Would you like me to do it? (I'm about to head to bed) Otherwise you are free to co-ordinate with them.

PR #1813, opened 2018-09-24
Update to Ember Tooltips 3

backspace:
[This issue](https://github.com/sir-dunxalot/ember-tooltips/issues/273) is causing the tests to fail, so it’s omgtoosoon for us. I also saw some rendering problems I’ll hopefully be able to report. (Commit to force PR deployments forthcoming.)


backspace:
[Here](https://github.com/sir-dunxalot/ember-tooltips/issues/275) is an issue about the rendering problems.

maxfierke:
@backspace As of ember-tooltips@3.1.0, we now support a `popperContainer` option that you may use to render the tooltips in `<body>` or wherever else you may like, if you would like to revisit upgrading!

PR #1815, opened 2018-09-25
Fix typo in job log page

mohd-akram:


backspace:
Thank you! This should be deployed later today.

PR #1816, opened 2018-09-25
Reduce builds request amount

artursmirnov:
Issue: https://travisci.slack.com/archives/C1T1E0F4L/p1537870032000100

This is a temporary solution until the API is fixed properly.

### DIFF
--- a/app/models/repo.js
+++ b/app/models/repo.js
@@ -89,7 +89,8 @@ const Repo = Model.extend({
   builds(id) {
     const builds = this.store.filter('build', {
       event_type: ['push', 'api', 'cron'],
-      repository_id: id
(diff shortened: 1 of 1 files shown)


backspace:
hmm it seems likely to me that the things I added to the `includes` for the “my builds” query are possibly slowing down other routes. It might be possible to tweak the build adapter to only conditionally include the things needed for “my builds” (jobs, specifically). Do you want to try that, or I could?

artursmirnov:
@backspace thanks for your feedback! You're right, I just checked and it appears that `build.jobs` slows down the requests. I'm not sure, I understand the "my builds" enough, could you please take a look how to best optimize this query? Thanks!

PR #1821, opened 2018-09-26
Revert ember-try version change

backspace:
This reverts #1819 as it’s broken the cron tests against Ember
and Ember Data betas and I can’t figure it out at the moment.


backspace:
I ran the cron-y script locally and the problem from the newer `ember-try` wasn’t present. 🎉

PR #1824, opened 2018-09-27
Experiment with build.canCancel not needing jobs

backspace:


backspace:
This may still be a good idea, but who would ever know?

PR #1826, opened 2018-09-29
Add YAML view

backspace:
omgtoosoon, obvs


backspace:
I expect the tests to fail, my attempt to clean up the hideous `messagesRequest.messages` thing seems to have led to test-only failures. Frustrating. I may have to return things to the previous interface despite my hatred for it.

backspace:
The PR build says it‘s still running but also says `Done. Your build exited with 0.`, so I’m merging.

PR #1829, opened 2018-10-05
Extend test coverage to cover recent regression

clekstro:
A user reported in https://github.com/travis-ci/travis-ci/issues/10112
that we were overwriting spaces when colored text was printed to the
console in logs.

I wasn't able to reproduce, but wanted to extend our current coverage in
to keep us from accidentally breaking this in the future.

As part of this work, I have added a new page object helper to easily
reference the entire text of a given line. Including this unfortunately
required modifying the scope of the page object's collection, so this
ended up being a bit more invasive than I would have liked. But it seems
worth it given the ability to assert on more complex lines in the future,
given that we were previously only able to assert against the contents of 
the first (and second) span.

### DIFF
--- a/tests/acceptance/home/with-repositories-test.js
+++ b/tests/acceptance/home/with-repositories-test.js
@@ -161,7 +161,7 @@ test('Pusher events change the main display', function (assert) {
       id: job.id,
       number: 0,
       final: false,
-      _log: '\u001B[0K\u001B[33;1mThe first line'
+      _log: '\u001B[0K\u001B[33;1mThe first line\n'
     });
   });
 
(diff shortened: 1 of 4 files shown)


artursmirnov:
@clekstro This may be relevant:
We had an issue with empty lines in log: https://github.com/travis-pro/team-teal/issues/2782 which I have fixed by squashing multiple spaces into one. But this fix had unexpected consequences for some other users https://github.com/travis-ci/travis-ci/issues/10133, so @backspace reverted it. I feel like this issue was also caused by my "fix" and has been resolved by the revert.

clekstro:
@artursmirnov Thanks, that's good context.

PR #1834, opened 2018-10-09
Fix repeated scrolling issue on certain in-progress jobs

sorta:
A possible way to address the repeated scrolling issue on some in-progress jobs . I'm not sure yet whether this approach would cause other issues though.

### DIFF
--- a/app/components/log-content.js
+++ b/app/components/log-content.js
@@ -38,6 +38,9 @@ Log.Scroll.prototype = $.extend(new Log.Listener(), {
       $('#main').scrollTop(0);
       let offset = element.offset();
       let scrollTop = ((ref = offset) != null ? ref.top : void 0) - (window.innerHeight / 3);
+      if (isNaN(scrollTop)) {
+        return;
+      }
       return $('html, body').scrollTop(scrollTop);
     }
   }


app/components/log-content.js
```diff
@@ -38,6 +38,9 @@ Log.Scroll.prototype = $.extend(new Log.Listener(), {
       $('#main').scrollTop(0);
       let offset = element.offset();
       let scrollTop = ((ref = offset) != null ? ref.top : void 0) - (window.innerHeight / 3);
+      if (isNaN(scrollTop)) {
+        return;
+      }
```
artursmirnov: Wouldn't [scrollIntoView](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView) work for this case instead of jQuery manipulations? It could just scroll to last log line. Browser support should not be an issue, only IE does not support smooth scrolling, but there is a [workaround](https://stackoverflow.com/a/51070207)

sorta: Yeah I think that's a good idea! I don't think it's something I'll have a lot of time to look into / try out this week though. But yeah I'm generally in favor of things that can reduce our reliance on jQuery.

PR #1835, opened 2018-10-10
Platform migration UI

clekstro:
See https://github.com/travis-pro/team-teal/issues/2838 for more details.


app/templates/components/github-apps-repository.hbs
```diff
@@ -4,9 +4,43 @@
...
 
-{{#unless repository.active_on_org}}
+{{#if repository.active_on_org}}
+  {{#if repository.permissions.migrate}}
+    {{#if repository.migrationStatus}}
+      <div
+        data-test-migration-status={{repository.migrationStatus}}
+        class="migration-status {{repository.migrationStatus}}"
+      >
+      {{#if (eq repository.migrationStatus "migrating")}}
```
artursmirnov: I would avoid using helpers wherever possible due to performance issues (helpers calculate values eagerly and don't cache results).

app/templates/components/github-apps-repository.hbs
```diff
@@ -4,9 +4,43 @@
...
+        <span>{{svg-jar "stage-passed" class="icon"}}</span>
+      {{/if}}
+      {{#if (eq repository.migrationStatus "failed")}}
+        <span>{{svg-jar "stage-errored" class="icon"}}</span>
+      {{/if}}
+      </div>
+    {{else}}
+      <a
+        data-test-migrate-repository-button
+        href="#"
```
artursmirnov: In some browsers links with `herf="#"` make page jump to top. It may be safer to remove `href` and add `pointer` class.

app/styles/app/layouts/profile.scss
```diff
@@ -668,6 +668,43 @@ header.space-between {
...
+        color: white;
+
+        .icon {
+          @include colorSVG(white);
+        }
+      }
+
+      .success {
+        background-color: white;
+        color: white;
```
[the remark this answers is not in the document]
clekstro: There isn't any text, just the icon, so this might be unnecessary.

app/models/repo.js
```diff
@@ -148,6 +159,13 @@ const Repo = Model.extend({
       then(data => this._convertV3SettingsToV2(data['settings']));
   },
 
+  startMigration() {
+    const url = `/repo/${this.get('id')}/migrate`;
+    return this.get('api').post(url).then(() => {
```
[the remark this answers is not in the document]
clekstro: I can try removing the `.get`s, but we should probably aim for some consistency (at least on a file-by-file basis) to avoid confusion. You're right that we could use a task, I just didn't bother because we didn't need any derived state.

artursmirnov:
Regarding the UI, I would make "Migrate" button a bit more prominent, as it looks like disabled:

image

app/styles/app/layouts/profile.scss
```diff
@@ -853,3 +890,9 @@ header.space-between {
     }
   }
 }
+
+.repository-migration-modal {
+  #migration-modal-cancel {
```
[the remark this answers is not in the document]
clekstro: Addressed in 51d8af974.

clekstro:
@artursmirnov When you have both the user feature flag and the `migrate` permission, the text opacity is set to `1` again. Agree that this is confusing, will have a look in the morning.

app/templates/components/github-apps-repository.hbs
drogus: I sometimes avoid helpers for clarity, but I don't think that we need to think about performance in such cases. Checking equality will not add any considerable overhead.

app/models/repo.js
```diff
@@ -42,6 +43,16 @@ const Repo = Model.extend({
   }),
   _branches: hasMany('branch'),
 
+  // We do not include an error state as part of an API response. This is only
+  // transmitted via a Pusher message for now.
```
drogus: I don't understand this comment. Even if an error state is not included in an API response, wouldn't it be updated when a pusher message comes in? Ie. I'm not sure what are the implications of that comment.

app/styles/app/layouts/profile.scss
```diff
@@ -668,6 +668,43 @@ header.space-between {
       }
     }
 
+    &.migratable {
+      opacity: 1.0;
+
+      .migration-status {
+        width: 66.5156px;
```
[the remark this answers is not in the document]
drogus: I have never considered fractional values for pixels, but it seems that it might actually change something: https://stackoverflow.com/a/9080685/661200

That's a minor thing (as it will most likely only be visible with high zoom levels), but it's an interesting thing to know.

app/templates/components/github-apps-repository.hbs
drogus: Regarding clarity, what I meant is that `#if isMigrating` or `#if migrationSucceeded` looks better than `eq`.

artursmirnov: @drogus I agree about clarity, but disagree about performance. Even in this case, this template is a part of component, which is used in a loop over user's GitHub Apps repositories. Thus, the value would be recalculated on each re-render of each repository. It's ok if a user has 10 repositories, but in cases with hundreds of repos on low-end devices this overhead may be pretty noticable. Also it's hard to predict the frequency of re-renders.

app/models/repo.js
clekstro: Deleted in 59a31110.

app/templates/components/github-apps-repository.hbs
drogus: I'd say that without benchmarking you don't really know if you're making things worse or better, but in the same time I can't prove anything either.

clekstro: Relevant: https://simplabs.com/blog/2017/03/21/on-computed-properties-vs-helpers.html. 
From that article, it seems like you mostly need to worry about eager evaluation *of the arguments*, so it doesn't seem like this would be a bad use case for a helper. I'll leave it as-is for now and we can address the topic more generally in a subsequent PR if we decide to avoid these types of template helpers altogether.

app/components/repository-migration-modal.js
```diff
@@ -0,0 +1,12 @@
+import Component from '@ember/component';
+import { action } from 'ember-decorators/object';
+
+export default Component.extend({
+  classNames: ['repository-migration-modal'],
+
+  @action
+  startMigration() {
+    this.get('repository').startMigration();
+    return this.get('onClose')();
```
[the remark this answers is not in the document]
clekstro: TL;DR: I'm not sure if adding no-ops makes things any more clear.

First, there's no connection between defining a property on a component and what arguments it can be passed. I can define `repository: null` on the component and that could mean it's a default value for state management used by a CP somewhere else within the component or that it's a placeholder to be overwritten by a passed-in value; it's quite ambiguous. I'd guess that 99% of the time, a component is representing a section of UI and tied to the DOM via its template, so you'd find its public API in how it's invoked (or, in the case of an addon, in its documentation), which is inevitably done in the template. If we wanted to be strict, we'd have to define something that checked passed params, which would involve overwriting a different framework hook. So I'm not sure this helps in that regard.

Second, in the case of `onClose()`, if we define the no-op and don't pass our own `onClose` action in the template, the no-op is invoked (instead of receiving an error that it doesn't exist). I find the error message more explicit about what is actually required, as users are forced to fix the exception to proceed.

@drogus @artursmirnov That said, if both of you are of a different opinion, I'm happy to make the change.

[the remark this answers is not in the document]
artursmirnov: @clekstro thank you for clarification! It totally makes sense. That was just an opinionated suggestion from me :)

app/templates/components/account-repositories.hbs
```diff
@@ -59,15 +59,26 @@
...
             {{svg-jar 'icon-megaphone' class='icon icon-megaphone'}}
-            <span class="label-align">
-              The following repositories cannot be migrated to travis-ci.com at this time because they are
-              currently active on our legacy platform travis-ci.org. This feature will be available shortly.
-              Please <a href='https://docs.travis-ci.com/user/open-source-on-travis-ci-com/#Existing-Open-Source-Repositories-on-travis-ci.org'>read our docs on open source migration</a> to learn more.
+            <span data-test-locked-github-apps-repositories-notice class="label-align">
+              {{#if account.allowMigration}}
+                Authorized users can now migrate the following repositories
+                currently active on our legacy platform travis-ci.org to travis-ci.com.
+                Please <a href='https://docs.travis-ci.com/user/open-source-on-travis-ci-com/#Existing-Open-Source-Repositories-on-travis-ci.org'>read our docs on open source migration</a> to learn more.
```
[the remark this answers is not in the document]
clekstro: Fixed in 32c64631.

clekstro:
Fixed the opacity issue in 79ead8759.

PR #1839, opened 2018-10-12
Fix config fetching logic

artursmirnov:
Context: https://travisci.slack.com/archives/CD38XGFGU/p1539344280000100

Example: https://travis-ci.org/junit-team/junit5/jobs/440495645/config - config does not load. 

As per investigation, `job-config-fetcher` service fetched builds by `job.build.id`, but `build` is an async relationship on `job`, which caused querying builds with `id: undefined` for unresolved builds.

### DIFF
--- a/app/adapters/job.js
+++ b/app/adapters/job.js
@@ -1,7 +1,7 @@
 import V3Adapter from 'travis/adapters/v3';
 
 export default V3Adapter.extend({
-  includes: 'build.request,build.commit,build.created_by',
+  includes: 'build.request,build.commit,build.created_by,job.config',
 
   coalesceFindRequests: true,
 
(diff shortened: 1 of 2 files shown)


app/adapters/job.js
```diff
@@ -1,7 +1,7 @@
 import V3Adapter from 'travis/adapters/v3';
 
 export default V3Adapter.extend({
-  includes: 'build.request,build.commit,build.created_by',
+  includes: 'build.request,build.commit,build.created_by,job.config',
```
drogus: I think that this line should be reverted. The whole idea of fetching config lazily is to not always include it as it is adding an extra work for the server and we only need it when viewing a job.

drogus:
I think that the diagnosis may be slightly wrong here. Even if a build is loaded asynchronously, a build id should be always available when a job is loaded. This is also true when a build is still a promise object, fetching `build.id` should still work.

PR #1843, opened 2018-10-15
Update the 'owner' page repos list

sorta:


joshk:
This is looking great!

I think the whitespace above and below the text needs to be trimmed a little. @Nat-ess can you take a look when you have a moment?

Nat-ess:
Thanks for this!

@joshk is right, the spacing below and above is larger than what we want it to be. This container should be the same height and have the same padding as "Active Repositories" on dashboard ( `21px` padding all around). I also notice that the icons on either side of the commit ref are underlined when you hover over them, and it should only be the ref copy thats underlined here.

artursmirnov:
@sorta the design update looks great! Thanks for working on this! Although I have noticed some minor alignment issues:

image

It's on Chrome `Version 70.0.3538.67 (Official Build) (64-bit)`, Windows 10.

Nat-ess:
Thanks for catching that @artursmirnov.

sorta:
I agree, thanks for catching that @artursmirnov ! I'll look into this today, it should be an easy thing to fix.

PR #1845, opened 2018-10-16
Add notice about support for windows builds

artursmirnov:
Suggestion: https://travisci.slack.com/archives/C1FAZPJNP/p1539297161000100

Here's how it looks like:
image

The link is only shown for `windows` jobs. Here's an example of one: https://as-windows-support-link-v2.test-deployments.travis-ci.org/yargs/yargs/jobs/440203815

### DIFF
--- a/app/templates/components/job-infrastructure-notification.hbs
+++ b/app/templates/components/job-infrastructure-notification.hbs
@@ -16,7 +16,10 @@
 {{#if isWindows}}
   {{#notice-banner}}
     Windows builds are in early access stage. Please head to the
-    <a href='https://travis-ci.community/c/windows' target='_blank'>Travis CI Community</a>
+    {{external-link-to
+      content='Travis CI Community'
+      href='https://travis-ci.community/c/windows'
+    }}
     forum to get help or post ideas.
   {{/notice-banner}}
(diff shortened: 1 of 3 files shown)


app/templates/components/job-infrastructure-notification.hbs
```diff
@@ -12,3 +12,11 @@
     {{/if}}
   {{/if}}
 {{/unless}}
+
+{{#if isWindows}}
+  {{#notice-banner}}
+    Windows builds are in early access stage. Please head to the
+    <a href='https://travis-ci.community/c/windows' target='_blank'>Travis CI Community</a>
```
clekstro: You could use the `{{external-link-to}}` here to shorten things up and add the necessary `rel` attrs. See https://www.jitbit.com/alexblog/256-targetblank---the-most-underestimated-vulnerability-ever/ for more details of why we added that component.

artursmirnov: Wow, I never thought about that! Thanks for sharing the article. And even Same-origin policy seems to [allow `location` changes](https://developer.mozilla.org/en-US/docs/Web/Security/Same-origin_policy#Window)

PR #1846, opened 2018-10-16
Update Ember etc to 3.5

backspace:
This isn’t quite ready as we’re held back from Broccoli 2 by `ember-cli-inline-images`. I’ll add another commit to change to [my fork](https://github.com/drogus/ember-cli-inline-images/pull/1) which should fix that.


backspace:
Percy reveals that emoji aren‘t rendering properly in build stage names, so this needs some further attention.

PR #1867, opened 2018-10-31
Fix apps redirect issue

sorta:
This should address an issue where the Github Apps installation redirect page was not redirecting correctly.

### DIFF
--- a/app/controllers/github-apps-installation.js
+++ b/app/controllers/github-apps-installation.js
@@ -19,7 +19,7 @@ export default Controller.extend({
 
   startPolling() {
     this.initialDelayPromise().then(() => this.fetchPromise().then(installation => {
-      this.transitionToRoute('account', installation.owner.login);
+      this.transitionToRoute('account');
     }));
   },
 


backspace:
Nice fix, thanks. I did have [this test](https://github.com/travis-ci/travis-web/blob/master/tests/acceptance/auth/github-apps-installation-redirect-test.js) for this functionality that it would be nice to bring back to prevent regressions on this, I gave up trying to make it work consistently during the rush to release this feature 😐

sorta:
@backspace Thanks! I agree, it would be great to bring that test back. It would have helped prevent this situation. Unfortunately, I'm not sure that it's something I have a lot of time to work on at the moment, but maybe I'll create a separate issue for it?

artursmirnov:
@sorta, @backspace, I'm merging and releasing this fix due to urgency. Let's add a test as a next iteration (I can handle it in case if Sophie is busy)

sorta:
Awesome @artursmirnov , thanks for merging this!

PR #1870, opened 2018-11-01
Add ember-tooltips to ignore list for Greenkeeper

artursmirnov:
Starting from [v3.0.0](https://github.com/sir-dunxalot/ember-tooltips/releases/tag/3.0.0) `ember-tooltips` migrated to another positioning library ([popper.js](https://popper.js.org) instead of deprecated [tether](https://github.com/HubSpot/tether)). This changed the DOM structure for tooltips. Now they are rendered "in-place" instead of "on body". And since in our layout in many places we use `overflow: hidden`, "in-place" tooltips become invisible. I spent some time trying to find a workaround, but didn't find any simple solution. So I suggest to freeze the dependency for Greenkeeper for now, until we find a solution, or the add-on changes settle and authors provide a proper way of handling such cases.


backspace:
yeah, I was trying to figure out possibilities in #1813 but the workaround they told me about didn’t seem to work. I’m not sure what else to do, so I’m fine with ignoring the newer versions for now, as much as the deprecation warnings annoy me.

I added another commit to ignore `ember-decorators` as we’re [stuck](https://github.com/travis-pro/team-teal/issues/2514) on the old version for a while.

PR #1871, opened 2018-11-01
Inconsistent repos list

artursmirnov:
Issue: https://github.com/travis-pro/team-teal/issues/2824

### DIFF
--- a/app/routes/owner/repositories.js
+++ b/app/routes/owner/repositories.js
@@ -1,10 +1,9 @@
 import TravisRoute from 'travis/routes/basic';
 import config from 'travis/config/environment';
-import { service } from 'ember-decorators/service';
-import { computed } from 'ember-decorators/object';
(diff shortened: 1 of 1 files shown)


app/routes/owner/repositories.js
```diff
@@ -1,10 +1,9 @@
 import TravisRoute from 'travis/routes/basic';
 import config from 'travis/config/environment';
-import { service } from 'ember-decorators/service';
-import { computed } from 'ember-decorators/object';
+import { inject as service } from '@ember/service';
 
 export default TravisRoute.extend({
-  @service auth: null,
+  features: service(),
```
backspace: It would match the rest of the codebase more to use `@service features: null` instead here, I think, did that not work for some reason?

artursmirnov: Ah, right, thanks for pointing 👍 that's a habit :)

PR #1882, opened 2018-11-07
Update to ember-cli-sentry 4.0.0 again

backspace:
Maybe this will work now?


backspace:
I’ve included this in #1920, where it actually works!

PR #1892, opened 2018-11-19
New Help page

artursmirnov:
Issue: https://github.com/travis-ci/product/issues/53


app/templates/components/travis-status.hbs
```diff
@@ -1,6 +1,6 @@
-<h3 class="footer-title">Travis CI Status</h3>
-<ul>
-  <li><div class="status-circle {{appLoading.fetchTravisStatus.last.value}}">Status:</div>
-      <a href="http://www.traviscistatus.com/">Travis CI Status</a>
-  </li>
-</ul>
+<div class="status-circle">
+  &#9679;
```
artursmirnov: Hey @backspace, what do you think about this change? It draws black dot as a symbol instead of styled div. This makes the dot size to depend on `font-size` and scale along with the label. Are there any downsides or possible regressions that comes to your mind? Thanks!

backspace: I can’t think of any problems with that approach, seems like a good solution to me. Maybe while you’re working on this you could add a tooltip or just `title` that shows the word for the status, if available? That would make it work better for people with trouble distinguishing colours. Better than that would be an icon, but that would need more input from others.

artursmirnov: @backspace thanks for response and for idea! I'll add the status description to tooltip 👍

artursmirnov: @backspace it's live in the PR deployment 🚀

backspace: Nice, it looks great to me, thank you!

artursmirnov:
@backspace this PR is mostly ready, could you please take a look once you have a chance?

app/components/zendesk-request-form.js
```diff
@@ -0,0 +1,108 @@
...
+    const { fullName: name, email } = this.auth.currentUser;
+    const { subject, description: body } = this;
+
+    try {
+      return yield $.post(`${apiHost}/${createRequestEndpoint}`, {
+        request: {
+          requester: { name, email },
+          subject,
+          comment: { body }
+        }
```
backspace: I haven’t been able to log in to Zendesk but the autoreply I received showed this as a subject:

image

That stray `}` is curious, should this maybe be `comment: body`? Just a guess.

app/routes/help.js
```diff
@@ -0,0 +1,4 @@
+import Route from '@ember/routing/route';
+
+export default Route.extend({
+});
```
backspace: I think this could be deleted 😀

app/templates/components/forms/form-select.hbs
```diff
@@ -0,0 +1 @@
+{{yield}}
```
backspace: The  `no-newline` icon here but not in the file before suggests to me your editor isn’t listening to `.editorconfig`, is that something you have set up?

app/components/zendesk-request-form.js
backspace: I was able to log in finally and the `}` isn’t there. So I’m not sure what the origin is. Maybe it’s not a problem with this code!

app/templates/components/travis-form.hbs
```diff
@@ -0,0 +1,10 @@
+<form onsubmit={{action 'submit' }} class="travis-form">
+  {{yield (hash
+    validators=validators
+    hasErrors=hasErrors
+    isValid=isValid
+    submit=(action 'submit')
+    input=(component 'forms/form-field' form=this fieldComponentName='forms/form-input')
+    textarea=(component 'forms/form-field' form=this fieldComponentName='forms/form-textarea')
+  )}}
+</form>
```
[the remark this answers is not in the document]
artursmirnov: @backspace thanks, I'm happy that you like it :) I was thinking about refactoring other forms too, but thought that it would inflate this PR. So I created a separate issue: https://github.com/travis-pro/team-teal/issues/2898 - I'm planning to implement it right after the new help page project.

app/components/travis-form.js
```diff
@@ -0,0 +1,52 @@
...
+  tagName: '',
+
+  validators,
+
+  erroredFields: filterBy('fields', 'isError', true),
+  validFields: filterBy('fields', 'isValid', true),
+
+  hasErrors: gt('erroredFields.length', 0),
+
+  isValid: computed('fields.@each', 'validFields.@each', function () {
```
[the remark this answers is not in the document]
artursmirnov: I'm not exactly sure, what's the difference between the two, although I had some problems using `[]` in the early days, so it's more like a habit :) Anyways, you're absolutely right that `.length` would work even better here 👍

app/helpers/combine-validators.js
```diff
@@ -0,0 +1,12 @@
...
+export function combineValidators(validators) {
+  return validators.reduce((previousValidator, currentValidator) => (
+    (value) => {
+      const currentValidatorResult = currentValidator(value);
+      return currentValidatorResult === true ? previousValidator(value) : currentValidatorResult;
+    }
+  ), () => true);
+}
+
+export default helper(combineValidators);
```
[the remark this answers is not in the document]
artursmirnov: The idea was to mostly use it in templates for forms definition:

```handlebars
{{#travis-form
  as |form|
}}
  {{form.input
    value=value
    validator=(combine-validators
      (form-validator form.validators.presence 'Please fill this field')
      (form-validator form.validators.regexp '[0-9]{5}' 'This field should hold 5 digits')
    onChange=(action (mut value))
  }}
{{/travis-form}}
```

I'm working on tests, will push some today, probably they will make the use cases clearer.

app/components/zendesk-request-form.js
artursmirnov: That's interesting! Thanks for pointing, I'll test more around this issue. But I don't think it's relevant to this code, because according to [Zendesk API](https://developer.zendesk.com/rest_api/docs/support/requests#create-request) `body` is nested inside `comment` 🤔

app/routes/help.js
artursmirnov: This is auto-generated by Ember CLI. Wouldn't it make sense to keep such things, in order to make sure that everything works as it was designed? For instance, you won't be able to inject a controller unless you have it explicitly generated (even empty). I can't imagine any problems with routes at the moment, but in sake of stability and upgradability I usually strive to stick with framework defaults. Does it make sense in your opinion?

app/templates/components/forms/form-select.hbs
artursmirnov: Hmm 🤔 It should be set up (I use vim), but I'll double-check.

app/templates/components/travis-form.hbs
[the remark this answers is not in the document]
backspace: Great! I agree that it’s better to keep separated.

app/routes/help.js
backspace: oh yes, I almost never use generators so I forget about these things haha. It’s not a huge deal to keep this.

PR #1895, opened 2018-11-23
Display language as shell for shell

joepvd:
Currently, with `language: bash`, _no language set_ is displayed in the build matrix overview and job overview. This should be set to _Shell_.


artursmirnov:
@joepvd looks good to me, thanks for the fix! 👍

PR #1896, opened 2018-11-23
Fix blank build at root

backspace:
This is a bit broken but it’s a start…


backspace:
Since I reverted the `.travis.yml`-showing PR, I’ll just open a new one with its branch.

PR #1897, opened 2018-11-23
Update to x-select 4

backspace:
The helper seems broken in a component integration test…


backspace:
I’ve folded this into #1914 and will merge that when the beta period is over. 🎉

PR #1915, opened 2018-12-04
Reduce test noise

backspace:
There’s still a bit to do here, but they’re quieter already.


backspace:
I think the ≈10 lines left are maybe inescapable, as they’re produced by tests that are deliberately causing errors. Better than the 200+ before I started this!

PR #1922, opened 2018-12-11
Pj 52 better waiting

paulthegeek:
#### What does this PR do?
- Create a UI for job waiting stages 
 
#### Related Issue
[52](https://github.com/travis-ci/product/issues/52)

#### How this PR makes you feel?
waiting


tests/acceptance/job/basic-layout-test.js
```diff
@@ -63,6 +63,211 @@ test('visiting job-view', function (assert) {
...
+    assert.ok(jobPage.createdBy.avatarSrc.startsWith('/images/favicon-gray.png'));
+
+    assert.notOk(jobPage.hasTruncatedLog);
+    assert.ok(jobPage.hasWaitingStates);
+    assert.equal(jobPage.stageOneText, 'Job received');
+    assert.equal(jobPage.stageTwoText, 'Queued');
+    assert.equal(jobPage.stageThreeText, 'Booting virtual machine');
+    assert.equal(jobPage.stageOneWaitingText, '1');
+    assert.equal(jobPage.stageTwoWaitingText, '2');
+    assert.equal(jobPage.stageThreeWaitingText, '3');
```
backspace: Since these strings are hard-coded in the template, there’s not much point adding assertions for them.

tests/acceptance/job/basic-layout-test.js
```diff
@@ -63,6 +63,211 @@ test('visiting job-view', function (assert) {
...
+    assert.equal(jobPage.branch, 'acceptance-tests', 'displays the branch');
+    assert.equal(jobPage.message, 'Running acceptance-tests This is a message', 'displays message');
+    assert.equal(jobPage.state, '#1234.1 received', 'displays build number');
+
+    assert.equal(jobPage.createdBy.href, '/testuser');
+    assert.equal(jobPage.createdBy.text, 'Mr T');
+    assert.ok(jobPage.createdBy.avatarSrc.startsWith('/images/favicon-gray.png'));
+
+    assert.notOk(jobPage.hasTruncatedLog);
+    assert.ok(jobPage.hasWaitingStates);
```
backspace: I think you can ignore `hasTruncatedLog`, since the log isn’t even showing.

I noticed that there’s no `assert.notOk(jobPage.hasWaitingStates)` anywhere. When adding a boolean check like this, there should always be at least one assertion for both `true` and `false` values. So you could add a check when the job is `started` or the like that `jobPage.hasWaitingStates` is false.

tests/acceptance/job/basic-layout-test.js
```diff
@@ -63,6 +63,211 @@ test('visiting job-view', function (assert) {
...
+
+    assert.equal($('head link[rel=icon]').attr('href'), getFaviconUri('yellow'), 'expected the favicon data URI to match the one for created');
+
+    assert.equal(jobPage.branch, 'acceptance-tests', 'displays the branch');
+    assert.equal(jobPage.message, 'Running acceptance-tests This is a message', 'displays message');
+    assert.equal(jobPage.state, '#1234.1 received', 'displays build number');
+
+    assert.equal(jobPage.createdBy.href, '/testuser');
+    assert.equal(jobPage.createdBy.text, 'Mr T');
+    assert.ok(jobPage.createdBy.avatarSrc.startsWith('/images/favicon-gray.png'));
```
backspace: Of the six assertions here, I think the only one you need is on `jobPage.state`.

tests/acceptance/job/basic-layout-test.js
```diff
@@ -63,6 +63,211 @@ test('visiting job-view', function (assert) {
...
+    assert.notOk(jobPage.stageTwoLoading);
+    assert.notOk(jobPage.stageThreeLoading);
+    assert.notOk(jobPage.stageOneLoaded);
+    assert.notOk(jobPage.stageTwoLoaded);
+    assert.notOk(jobPage.stageThreeLoaded);
+    assert.ok(jobPage.stageOneLoadingLineInactive);
+    assert.ok(jobPage.stageTwoLoadingLineInactive);
+    assert.notOk(jobPage.stageOneLoadingLineActive);
+    assert.notOk(jobPage.stageTwoLoadingLineActive);
+  });
```
backspace: If you used `generatePusherPayload` here, you could collapse all this into one test, rather than having to set up all the necessary models for each step of the process. You could have the various `job:whatever` events come in via fake Pusher, like in the “Pusher events change the main display” test, then it would be clear that this test is about checking all the permutations of the element classes.

tests/pages/job.js
```diff
@@ -27,6 +27,23 @@ export default create({
...
+  stageOneWaitingText: text('.job-waiting-stages-container .stage-1-container span.waiting-stage-1-text'),
+  stageTwoWaitingText: text('.job-waiting-stages-container .stage-2-container .waiting-stage-2-text'),
+  stageThreeWaitingText: text('.job-waiting-stages-container .stage-3-container .waiting-stage-3-text'),
+  stageOneText: text('.job-waiting-stages-container .stage-1-message'),
+  stageTwoText: text('.job-waiting-stages-container .stage-2-message'),
+  stageThreeText: text('.job-waiting-stages-container .stage-3-message'),
+  stageOneLoadingLineInactive: hasClass('loading-line-1', '.loading-line-container-1 span'),
+  stageTwoLoadingLineInactive: hasClass('loading-line-2', '.loading-line-container-2 span'),
+  stageOneLoadingLineActive: hasClass('loading-line', '.loading-line-container-1 span'),
+  stageTwoLoadingLineActive: hasClass('loading-line', '.loading-line-container-2 span'),
```
backspace: I think it could lead to more clear and expressive test assertions to structure this using what the page object library calls [components](http://ember-cli-page-object.js.org/docs/v1.15.x/components). It could look something like this:

```
waitingStates: {
  scope: '.job-waiting-states-container',

  one: {
    scope: '.stage-1-container',
    isInactive: hasClass('not-loaded-ring'),
    isLoading: hasClass('spinner', 'svg'),
    isLoaded: hasClass('stage-loaded-ring'),
  },
  
  firstLine: {
    scope: '.loading-line-container-1 span',
    isInactive: notHasClass('loading-line'),
    isActive: hasClass('loading-line'),
  },

…
}
```

Then the assertions can be more concise and descriptive:

```
# created

assert.ok(jobPage.waitingStates.isVisible);

assert.ok(jobPage.waitingStates.one.isLoading);
assert.ok(jobPage.waitingStates.firstLine.isInactive);
assert.ok(jobPage.waitingStates.two.isInactive);
assert.ok(jobPage.waitingStates.secondLine.isInactive);
assert.ok(jobPage.waitingStates.three.isInactive);

# change state to queued

assert.ok(jobPage.waitingStates.one.isLoaded);
assert.ok(jobPage.waitingStates.firstLine.isActive);
assert.ok(jobPage.waitingStates.two.isLoading);
…

# change state to started

assert.ok(jobPage.waitingStates.isHidden);
```

Does that make sense?
    
I avoided the name “stage” here because it’s confusing to me to have two different concepts being called stages.

app/templates/components/dashboard-row.hbs
```diff
@@ -35,7 +35,7 @@
       <p class="row-content color">
         {{status-icon status=repo.currentBuild.state}}
         {{#link-to "build" repo.owner.login repo.name repo.currentBuild.id title="Last build on the repo"}}
-        <span class="label-align">#{{repo.currentBuild.number}} {{repo.currentBuild.state}}</span>
+        <span class="label-align">#{{repo.currentBuild.number}} {{humanize-state repo.currentBuild.state}}</span>
```
backspace: There should be an assertion somewhere in the dashboard tests that reflects this changed template.

tests/pages/job.js
backspace: also I see some `show-stage-message` and `hide-stage-message` classes that have no corresponding assertions, I guess for mobile? Those could be included here too, and then in the assertions.

app/models/job.js
```diff
@@ -1,3 +1,4 @@
+
 /* global Travis */
```
backspace: You could remove this to reduce diff noise.

app/templates/components/log-content.hbs
```diff
@@ -1,9 +1,54 @@
...
+    <div class="log-notice">Back to the queue...</div>
+  {{/if}}
+
+  {{#if job.inQueue}}
+    <div class="log-notice">Hang tight, this job is next...</div>
+  {{/if}}
+
+  {{#if job.toBeCreated}}
+    <div class="log-notice">Things are being created...</div>
+  {{/if}}
```
backspace: Are these being used? I couldn’t find them showing in the PR deployment.

tests/acceptance/job/basic-layout-test.js
backspace: (I did this lower down where I left a comment on the structure of the page object)

tests/pages/job.js
backspace: I think these can be deleted now that they’re covered by the more structured objects?

tests/acceptance/dashboard/repositories-test.js
```diff
@@ -171,6 +171,37 @@ module('Acceptance | dashboard/repositories', function (hooks) {
...
+      currentBuild: createBuild,
+    });
+
+
+    await visit('/dashboard');
+
+    assert.equal(page.activeRepos.repos[1].lastBuild, '#3 received');
+
+    percySnapshot(assert);
+  });
```
backspace: I think it’s not necessary to have an entire test for this, which is probably a response to my comment [here](https://github.com/travis-ci/travis-web/pull/1922#discussion_r243409023)? I think you could just change [this assertion](https://github.com/travis-ci/travis-web/blob/c014656463fc83bf9dffecbc09da4fbe14da3978/tests/acceptance/dashboard/repositories-test.js#L232) to check for the `created`-to-`received` conversion.

backspace:
The Percy diff shows two tests are missing snapshots, which is strange to me… do you know what that’s about? I don’t see them in the changes at all.

backspace:
I think maybe the problem with Percy is that you don’t need to `import` the `percySnapshot` function… that only needs to be imported when it’s the new-style `module` test vs the older `moduleForAcceptance` test.

PR #1923, opened 2018-12-11
Add handling for free plan

backspace:


backspace:
Note to @joshk to pass this on this if it ever actually happens.

PR #1924, opened 2018-12-11
Remove route action helper addon

backspace:
This is throwing deprecations as of Ember 3.6. It’s not that
widely-used in the application and route actions aren’t that
favoured in the RFC discussion, so we decided to remove it.

https://github.com/emberjs/rfcs/pull/394


artursmirnov:
This is awesome, thanks @backspace! 👍

PR #1933, opened 2018-12-19
Fix date-formatted titles

backspace:
This only addresses one but there was no test for this previously
and I broke it in #1915, so I’m going to continue on this tomorrow
to add assertions for other `title` attributes I can find.


backspace:
I’ll have to return for the other `titles` as I’m out of time before a break, but I _will_ return 😆

PR #1937, opened 2019-01-10
help page emails dropdown

artursmirnov:
This is a second phase of https://github.com/travis-ci/product/issues/53
It introduces a dropdown with a list of all known emails in support request form.


app/components/forms/form-select.js
```diff
@@ -0,0 +1,49 @@
...
+    }
+    return classes.join(' ');
+  }),
+
+  onChange() {},
+  onFocus() {},
+  onBlur() {},
+  onInit() {},
+
+  onchange: computed(function () {
```
backspace: Are there both `onChange` and `onchange` because this is inheriting from the addon component? Maybe there could be a comment to explain that, otherwise it’s pretty mysterious.

I haven’t thought deeply about this but is it possible this could be accomplished via composition rather than inheritance? If not, no big deal, it just worries me to be extending such a complex component, it seems like things could break more easily.

app/adapters/user.js
```diff
@@ -14,6 +14,6 @@ export default V3Adapter.extend({
   // This overrides the parent implementation to ignore the query parameters
   queryRecord(store, type, query) {
     let url = this.buildURL(type.modelName, null, null, 'queryRecord', query);
-    return this.ajax(url, 'GET', { data: {include: 'owner.installation' }});
+    return this.ajax(url, 'GET', { data: {include: query.included}});
```
backspace: I like this parameterisation of `include`… in previous work I’ve added yet another item to the `include`, which has felt wasteful to me because so many things are being returned for all possible queries. If I need to do this again, I’ll be able to do something like this to only include things when they’re needed, so thanks!

app/services/auth.js
```diff
@@ -12,6 +12,8 @@ import URLPolyfill from 'travis/utils/url';
 
 const proVersion = config.featureFlags['pro-version'];
 
+let includes = [];
```
backspace: I’m wondering why this is out here rather than in the `refreshUserData` function; probably a good reason, maybe worth explaining?

app/components/forms/form-select.js
artursmirnov: 1) You are right, `onchange` is the event handler exposed by the `ember-power-select` component and `onChange` is there mostly to align with other `travis-form` fields. Although, after I changed `travis-form` to expose `form-field` directly (which is responsible for exposing public events), it doesn't make sense. `form-field` can bind to `ember-power-select` native events internally. I'll change that, thanks for catching! 👍 

2) I also prefer composition over inheritance in regards with components (esp. external ones). But official add-on documentation suggest inheritance as a preferred way: https://ember-power-select.com/cookbook. I would suggest to keep it this way for now and see how it works. In case you have any issues with power select in future, please let me know.

app/services/auth.js
artursmirnov: This is for consistency reasons. Kind of a "singleton"-ish behaviour - the array is shared between all instances and `refreshUserData` calls. Each call to `refreshUserData` that has new `includes`, appends missing ones to this array and never deletes. This is to allow requested data survive future refreshes. Let's say, on the Help page the app requests `user` with `user.emails` included and shows them on UI. Something in background (let's say, pusher event or something else) calls `refreshUserData` via `auth` service without `user.emails`, which will override the data, remove `emails` and break the Help page UI. The current implementation aims to avoid such cases. 
Does it makes sense in your opinion? Or would you prefer single-request enhancements?

app/components/forms/form-select.js
backspace: ah well if the documentation says to do it this way, I agree we should do it this way hehe thanks

app/services/auth.js
backspace: oh, making sure it doesn’t override is a great reason I didn’t consider. Maybe some commentary would help communicate this to future visitors to this code, like:

```
// Ensures future user-fetching doesn’t erase emails attribute
let persistentIncludes = [];
```

or something?

artursmirnov: Absolutely! Thanks, I'll add a comment 👍

artursmirnov:
@backspace all feedback has been implemented, thanks!

PR #1945, opened 2019-01-30
Add branch search search

backspace:
This is #1941 with a proper branch name for staging deployments 🙃


tests/acceptance/repo/settings-test.js
```diff
@@ -264,7 +277,42 @@ module('Acceptance | repo settings', function (hooks) {
...
+
+      return newCron;
+    });
+
+    await settingsPage.cronForm.branch.click();
+    assert.equal(settingsPage.cronForm.branch.options.length, 2);
+
+    await settingsPage.cronForm.branch.search('another');
+    // FIXME the options aren’t what I would expect…
+    // assert.equal(settingsPage.cronForm.branch.options.length, 1);
```
backspace: My desire with these assertions is to first test what shows before any search has happened (the current plan is to show branches sorted in alphabetic order) and then show the results of a search. Looking at the Mirage output, the search _is_ happening, but the popup element doesn’t contain the results I’d expect. It’s strange because the code does seem to work in production.

backspace:
This is pretty close to ready, it needs:
• refinement of the branches that show before search to be alphabetic
• design tweaks
• maybe show the popup on tab rather needing another keypress? 🤔
• see my comment in the acceptance test file
• error-handling, which should have already been present!

Here’s the design:

image

artursmirnov:
@backspace thank you so much for the summary!

artursmirnov:
This is implemented in https://github.com/travis-ci/travis-web/pull/2024

PR #1950, opened 2019-02-12
Update generation of redirect URL

backspace:
Did this break with Ember 3.7? 🤔

I’d prefer this to be tested but I don’t know how that would work…


backspace:
I think there’s a way to test this but I’ll merge it for now for an immediate fix and spend more time understanding and adding a test.

PR #1951, opened 2019-02-12
Fix redirect in IE 11 and add test

backspace:
Redirection from an auth-requiring page is currently broken in IE 11,
which lacks the URL class. This restores the URL-generation to something
closer to what existed before #1950, but uses the URL polyfill instead.

It also includes an automated test that should hopefully prevent a
regression on the auth-requiring redirection functionality.


artursmirnov:
Hey @backspace, wouldn't plain old `encodeURIComponent(window.location.href)` work here?

backspace:
The goal there is to access the `origin`, so I don’t think `encodeURIComponent` would help with that, _but_ `window.location.origin` would…? I’m not sure how this whole `URL` thing happened in the first place 🙃

artursmirnov:
I think it was because of a poor browser support for `location.origin` (https://developer.mozilla.org/en-US/docs/Web/API/Window/location#Browser_compatibility). But anyways, I don't see any point to extract `origin` at all. As I understand the initial code, it then constructed the complete url again, didn't it? And essentially, what we want is to redirect user back to where he left off, right? Although, these are just my assumptions, and I may be missing some details 🤷‍♂️

backspace:
it’s a good question, I guess I’ve been dealing with this in a superficial “I want to duplicate the existing behaviour” way without understanding why it is the way it is, which is what I often find myself doing… I figure there must be a reason it was implemented as it was, because otherwise it seems nonsensical. There‘s not much testing around this as it exceeds the boundaries of Ember acceptance tests.

I’ll see what I can figure out from the history. I know that I originally added a URL polyfill, but I don’t remember what I encountered before that.

artursmirnov:
Closing this PR as the issue has been fixed in another one.

PR #1953, opened 2019-02-13
Implement insights visibility settings

sorta:
These updates allow for getting and setting of account and organization insights visibility settings. Not quite finished yet but getting there.

This PR Makes no sense without the one for the owner insights page https://github.com/travis-ci/travis-web/pull/1878


app/components/profile-nav.js
```diff
@@ -31,4 +31,9 @@ export default Component.extend({
     return !enterprise && !!billingEndpoint;
   }),
 
+  isOrganizationAdmin: computed('model', function () {
```
backspace: I don’t know how likely this is, but I wonder whether the dependent key should be `model.permissions.admin` instead. Presumably `model` will never change but it could be that the permissions change and get refetched somehow…? Definitely an edge case situation though. (I wouldn’t expect `isOrganization` to ever change though.)

This is likey overkill in this situation but sometimes I like to break things down to be able to make use of composed macros:

```
isOrganization: alias('model.isOrganization'),
hasAdminPermissions: alias('model.permissions.admin'),

isOrganizationAdmin: and('isOrganization', 'hasAdminPermissions'),
```

In this way, the dependent keys are just there, and the composed property will be recalculated whenever one changes. BUT I don’t think this is necessary in this case… haha hopefully this isn’t overboard-commentary 🙃

app/templates/account/settings.hbs
```diff
@@ -76,4 +76,91 @@
...
     {{/if}}
   </section>
+  {{#if features.proVersion}}
+    <section class="active-visiblity-settings" data-test-active-visiblity-settings>
+      <h3>Insights Visibility Settings</h3>
+      <div class="insights-settings">
+        {{!-- <div class="name" data-test-insights-settings-title>Insights</div> --}}
+        <p class="description" data-test-insights-settings-description>
+          Make more informed decisions about your development workflow using your build Insights.<br />
+          {{#link-to 'owner.repositories' account.login (query-params tab='insights')}}View {{account.name}}'s Insights{{/link-to}}
```
backspace: Further to my commentary above, this kind of string-construction is something I’d like to see an assertion of.

app/templates/account/settings.hbs
```diff
@@ -76,4 +76,91 @@
...
+          </li>
+        </ul>
+      </div>
+    </section>
+  {{/if}}
 </section>
+
+{{#if (and features.proVersion isShowingInsightsVisibilityModal)}}
+  {{#modal-dialog
+    toggleInsightsVisibilityModal=(action 'toggleInsightsVisibilityModal')
```
backspace: hmm now that I see you’re passing this action to another component, my commentary about mutating in the template makes less sense… it could perhaps be done with `let` to declare the action in the template, but I’m not sure that’s too much of an improvement

I tried this out and had an accessibility observation: I think the keyboard focus should be moved to the first button when the modal is shown. It seems [currently untenable](https://github.com/yapplabs/ember-modal-dialog/issues/236) to remove the possibility of focus on elements behind the modal, but I think it should be possible to focus the button so at least focus is in the correct general area?

app/templates/components/profile-nav.hbs
```diff
@@ -81,6 +81,12 @@
             {{#link-to 'organization.repositories' model }}Repositories{{/link-to}}
           </li>
 
+          {{#if isOrganizationAdmin}}
```
backspace: I tried the org PR deployment and this tab shows even though no actual settings are present, maybe this needs a `proVersion` check too?

app/templates/organization/settings.hbs
```diff
@@ -0,0 +1,113 @@
...
+          <li class='travis-radio-list__item'>
+            <label for="organization_insight_members">Only allow organization members to see insights from your private builds</label>
+            <input
+              type="radio"
+              name="organization_insight_vis"
+              class="travis-radio-input"
+              id="organization_insight_members"
+              value="members"
+              checked={{eq privateInsightsVisibility 'members'}}
+              onchange={{action (mut privateInsightsVisibility) 'members'}}
```
backspace: hah well re my `toggle` commentary, I see that you’re mutating in-template here, so you already know about that… good call here!

app/templates/organization/settings.hbs
```diff
@@ -0,0 +1,113 @@
...
+    </section>
+  {{/if}}
+</section>
+
+{{#if (and features.proVersion isShowingInsightsVisibilityModal)}}
+  {{#modal-dialog
+    toggleInsightsVisibilityModal=(action 'toggleInsightsVisibilityModal')
+    targetAttachment="center"
+    translucentOverlay=true
+  }}
```
backspace: I’m getting the “should this be extracted” itch here again… but I’m just not sure about it! What do you think?

app/templates/account/settings.hbs
```diff
@@ -76,4 +76,91 @@
...
+      {{action 'toggleInsightsVisibilityModal'}}
+    >Cancel</a>
+
+    <a
+      data-test-insights-visibility-modal-confirm-button
+      href="#"
+      role="button"
+      class="button button--blue"
+      {{action (perform setPrivateInsights)}}
+    >Confirm</a>
```
backspace: Given that these are both being styled as and being used as buttons, rather than links (with no destination), maybe they could just be buttons?

app/components/profile-nav.js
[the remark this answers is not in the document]
sorta: That makes a lot of sense, and looks a lot cleaner/nicer too! Thanks :D

app/templates/account/settings.hbs
sorta: I think if I switch the actions to all use the toggle helper, I might be able to avoid needing to pass the action in. Or maybe at least pass in the toggle helper version of the action? I'll try some things out and see how it goes.

And I definitely will try to see what I can do about keyboard focus on the modal. That's an excellent point, thanks!

app/controllers/account/settings.js
```diff
@@ -43,5 +65,24 @@ export default Controller.extend({
...
+    try {
+      yield this.preferences.set('private_insights_visibility', val);
+      this.flashes.clear();
+      this.flashes.success(`Your private build insights are now ${val}.`);
+    } catch (err) {
+      this.flashes.clear();
+      this.flashes.error('Something went wrong and your insights settings were not saved.');
+    }
+    this.set('isShowingInsightsVisibilityModal', false);
+  }).restartable(),
```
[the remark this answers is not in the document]
backspace: in case you haven’t already started on an acceptance test, I wanted to point out that we’re in the midst of a slow migration to a newer format for acceptance tests… [this](https://github.com/travis-ci/travis-web/blob/f340ea6f1acc565ddbf35c350add6ad347fee5ea/tests/acceptance/dashboard/repositories-test.js) is an example of the new standard, which involves the use of `async` and `await` rather than the `andThen` blocks. Since we still have a lot of acceptance tests that use the old style, it’s not disastrous to add another, but I feel like when adding new ones, we might as well use the more modern style.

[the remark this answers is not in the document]
backspace: For some reason I thought [this acceptance test](https://github.com/travis-ci/travis-web/pull/1878/files#diff-0e18780d33a062ccccacf76d2383ad51) was using the old style but I was wrong, so clearly you already know about this, sorry for the unnecessary commentary 😳

app/templates/organization/settings.hbs
sorta: I've gone back and forth on it. It _does_ seem silly to have 2 of almost the exact same thing. At this point, I'm leaning toward trying to do some sort of extraction, though I haven't decided where to draw the lines. Maybe like the radio button list, save button, and modal together make sense as a component. Then at the very least some of the duplication that is in the templates can be reduced.

app/controllers/account/settings.js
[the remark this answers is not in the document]
sorta: Hey no worries! :) I'd rather have too much commentary/communication than not enough. I believe there's at least one place where I was using the old style, so it's good to know that we're trying to move away from that. And things that I do "already know", I don't always remember haha. So it's good to have reminders, and I really appreciate all the feedback I've received so far :D

app/templates/organization/settings.hbs
sorta: I did end up extracting out a radio button list component and it is helping to keep things a bit more DRY

app/routes/organization/settings.js
```diff
@@ -1,7 +1,22 @@
...
+      this.transitionTo('organization.repositories', this.modelFor('organization'));
+    }
+  },
+
+  model() {
+    const organization = this.modelFor('organization');
+    if (organization.permissions.admin !== true) {
+      this.transitionTo('organization.repositories', organization);
+    }
+    const preferences = this.store.query('preference', { organization_id: organization.id });
```
artursmirnov: It seems like `preferences` is a good case for `organization.hasMany('preference', { async: true })`, WDYT? It would allow for better request/caching handling and proper payload extraction (e.g. camelize property keys)

app/templates/account/settings.hbs
```diff
@@ -24,8 +24,8 @@
   {{#if scrollToEmail}}
     {{scroll-here}}
   {{/if}}
-  <section class="email-settings" data-test-email-settings>
-    <h3>Email notifications</h3>
+  <section id="email-settings" class="email-settings" data-test-email-settings>
+    <h3 id="email-notifications">Email notifications</h3>
```
artursmirnov: Why do sections need `id`? I didn't notice any uses of them. Generally, I would avoid using `id`s, as it's hard to track their uniqueness on the page with dynamic views. And also Ember sets `id`s for some nodes for it's own purposes.

app/templates/components/profile-nav.hbs
```diff
@@ -81,6 +81,12 @@
             {{#link-to 'organization.repositories' model }}Repositories{{/link-to}}
           </li>
 
+          {{#if showOrganizationSettings}}
+            <li data-test-org-settings-tab>
+              {{#link-to 'organization.settings'}}Settings{{/link-to}}
```
artursmirnov: Doesn't it require `organization` model to be provided, similar to the Repositories link above? I mean, there may be multiple organizations in the list, and usually it's mandatory to pass a model to resolve the route. Otherwise, how would it know, for which org to show the settings?

app/templates/organization/settings.hbs
```diff
@@ -0,0 +1,18 @@
+<section class='organization-settings'>
+  {{#if (and features.proVersion organization.permissions.admin)}}
```
artursmirnov: I would suggest to move this condition to controller giving it a name something like `showOrganizationSettings`.

app/templates/organization/settings.hbs
```diff
@@ -0,0 +1,18 @@
...
+    <section id="insights-visibility-settings" class="insights-visibility-settings" data-test-insights-settings>
+      <h3 data-test-insights-settings-title>Insights Visibility Settings</h3>
+      <div class="insights-settings">
+        <p class="description" data-test-insights-settings-description>
+          Make more informed decisions about your development workflow using your build Insights.<br />
+          {{#link-to 'owner.repositories' organization.login (query-params tab='insights')}}View {{organization.name}}'s Insights{{/link-to}}
```
artursmirnov: I would suggest to split such long lines, for better readability.

app/templates/components/profile-nav.hbs
sorta: Huh yeah I think you're right. It's weird that it went unnoticed, but maybe it's because I'm only the admin/owner of 1 organization. Either way, I'll update this to pass in `model`

app/templates/account/settings.hbs
sorta: @MariadeAnton requested that we be able to give out `#` links to specific sections, but yeah with the dynamic nature of the views maybe it is precarious. Maybe there's a better way to achieve the same thing?

app/components/visibility-setting-list.js
```diff
@@ -4,6 +4,7 @@ import { reads, empty, not, lt, gt, equal, and } from '@ember/object/computed';
 import KeyboardShortcuts from 'ember-keyboard-shortcuts/mixins/component';
 
 export default Component.extend(KeyboardShortcuts, {
+  tagName: 'div',
```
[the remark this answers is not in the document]
sorta: Change made :)

app/components/profile-nav.js
```diff
@@ -22,6 +23,12 @@ export default Component.extend({
 
   showSubscriptionStatusBanner: and('checkSubscriptionStatus', 'model.subscriptionError'),
 
+  isOrganization: alias('model.isOrganization'),
```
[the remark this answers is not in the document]
sorta: Ahh cool, makes sense. Change made 😄

app/templates/components/profile-nav.hbs
sorta: Change made 👍

app/templates/organization/settings.hbs
sorta: Yeah that sounds good to me. Change made :)

app/templates/organization/settings.hbs
sorta: Change made :)

app/routes/organization/settings.js
sorta: Yeah that seems like maybe the better way to do it. I haven't worked a lot with ember-data relationships yet, what would be involved?

app/templates/account/settings.hbs
artursmirnov: We have a `scroll-here` component for such purposes. Here's an example: https://github.com/travis-ci/travis-web/blob/master/app/templates/help.hbs#L1

Basically, first you define anchor query parameters in controller:
```js
export const SECTION = {
  EMAIL_SETTINGS: 'email-settings',
  INSIGHTS: 'insights'
}

export default Constroller.extend({
  queryParams: ['section'],
  section: null,
  
  isEmailSettingsSection: equal('section', SECTION.EMAIL_SETTINGS),
  isInsightsSection: equal('section', SECTION.INSIGHTS)
})
```

And then, depending on query parameter, you can define the places to scroll to in the template:
```hbs
{{#if isEmailSettingsSection}}{{scroll-here}}{{/if}}
...
{{#if isInsightsSection}}{{scroll-here}}{{/if}}
```

artursmirnov:
@sorta I have briefly tested the UI and found this:
image

Seems like the organization doesn't have a `name`, only `login`.

artursmirnov:
Also I would suggest to remove the outline for focused elements:
image

We don't support keyboard navigation properly anyway (try to `tab` through the page), but in my opinion it greatly decreases user experience for the most of our users. At least for me, it annoys me so much. 

I agree, that keyboard navigation may be useful, and there are many people who use it, so probably we should think about adding proper support for it (e.g. add keyboard shortcuts for navigation and common actions, etc.) and do the fields highlighting that fits naturally in our UI.

app/templates/account/settings.hbs
sorta: Nice! I'll switch to that.

sorta: Updated to use `section` query parameter and `scroll-here` 😄

app/routes/organization/settings.js
artursmirnov: In general, first I would create a model (as soon as we already have `preference` model for user preferences, I would give it a name like `orgPreference`). Then set it as a relation on `organization` model: `preferences: hasMany('orgPreference', { async: true })`. Then on the route model I would try to call it:
```js
model() {
  return this.modelFor('organization').get('preferences'); // async relationships return promises
}
```
and see how it goes.
Apparently, in this case it would fail, because it would try to fetch `/orgPreferences` endpoint, so we have to tweak adapter: generate an adapter for `orgPreference`, make it extend `V3Adapter`, add `pathForType` hook and return the correct url, override other hooks (described in the EmberData docs for JSONAPIAdapter) to make sure, that it makes proper requests. Then I would take a look how it extracts data - if it looks good, then just leave it as is, otherwise, do the same as for Adapter but for Serializer.

Although, since the release is planned for today, probably it would make sense to leave it as it is for now and maybe approach it sometime later.

sorta:
I've made an update to address the situation where an organization doesn't have a name, thanks for catching that.

RE: The keyboard navigation outline, I'm a little conflicted. I personally find them annoying as well, but keyboard navigation is an accessibility issue. I'm not sure how important it is to how many people, but people with access needs are so often left behind. I'd like to start improving accessibility when and where we can.

A possible compromise could be to tone the style down so it's not so intense. There are also some possible options to try to differentiate between keyboard and mouse focus in this [article](https://css-tricks.com/keyboard-only-focus-styles/) and its comments, although we'd have to keep in mind how autofocus is implemented.

artursmirnov:
I agree about the accessibility, and, of course, totally support its improvement in `travis-web`. Although, I'm pretty sure, that _visual_ appearance of focused field, does not affect it in any way. Screen readers do not consider it, they mostly rely on HTML semantics and ARIA attributes (and ignore CSS at all). I understand that accessibility is not only screen readers, but also allowing users to avoid using mouse, which is unfortunately also _currently_ not possible in `travis-web`. It's just impossible to navigate through the app without mouse. I'm sure that we have to address these issues, but in the meanwhile it just doesn't make sense to me to decrease usability for everyone while not _actually_ improving it for accessibility. I didn't mean to abandon accessibility in any way (I'm sorry if it sound so), but believe that it must be done in a proper way - when it's _actually_ helpful and doesn't decrease user experience for other users. WDYT?

sorta:
Yeah those are all good points. I understand that position and I don't necessarily disagree. It just feels bad to give any ground on stuff like this sometimes, y'know. It would be nice if keyboard nav was a thing on one component as a start, but it's pointless if they can't even get to the page, yeah? Especially since it does adversely affect things for most users.

Ultimately, it would be super easy to reintroduce focus styles in a future PR if deemed appropriate, so yeah I think for now we can remove them.

sorta:
Oh and FWIW I never thought that you meant to abandon accessibility :)

app/routes/organization/settings.js
sorta: Thanks for all this info, it's super helpful! The release got pushed back to tomorrow, so if I end up with time for it I'll try to fit this in.

artursmirnov:
Yeah, I feel the same 😄I was thinking about the keyboard navigation issues so often, probably it would make sense to file a product issue for that.

app/routes/organization/settings.js
sorta: Ok I think given where things are at right now, I don't think there will be time to do this before launch. So yeah I think for now let's leave this as is. I'm already planning on writing up a few issues after the launch, so maybe I'll write one up for this as well.

artursmirnov:
@sorta It looks great to me, thank you! 🚀

PR #1962, opened 2019-03-01
Update job banner to message PHP default jobs

MariadeAnton:
For upcoming PHP default update from 5.5 to 7.2

While I was there, also updated precise banner to recommend upgrading to Xenial

I need help evaluating there's not a pinned php version, rather than just the language check, is that feasible?


kchetri222:
software jobs https://foundthejob.com/category/it-software-jobs/

PR #1964, opened 2019-03-07
[WIP] Owner insights with ChartJS

sorta:
Continuing the work started in https://github.com/travis-ci/travis-web/pull/1878 , here is where I'll be working on the migration from Highcharts to Chart.JS

I thought I might create a separate branch/PR for this work in order to make it easy to compare the new PR deployments against the old ones.


sorta:
So Chart.JS ended up not having as thorough documentation as I would have liked. Also, the ember addon is not as actively maintained as the one for C3. For these reasons and a couple other small things, I think I'll close this PR and instead focus on migrating to C3.

PR #1949, opened 2019-02-12
[WIP][MOSS] add migrate tab

artursmirnov:
Issue: https://github.com/travis-pro/team-teal/issues/2915

PR #1971, opened 2019-03-13
Refactor pagination component

artursmirnov:
A small refactoring of pagination component, mostly for simplicity and stability. Also extracted from https://github.com/travis-ci/travis-web/pull/1949/

It also fixes the issue with pagination on narrow screens (it wraps and looks broken)

### DIFF
--- a/app/components/pagination-navigation.js
+++ b/app/components/pagination-navigation.js
@@ -1,56 +1,43 @@
 import Component from '@ember/component';
 import { computed } from '@ember/object';
-import { alias } from '@ember/object/computed';
+import { reads } from '@ember/object/computed';
 
 export default Component.extend({
   tagName: 'nav',
   classNames: ['pagination-navigation'],
 
   queryParam: 'page',
-  pagination: alias('collection.pagination'),
+  outer: 1,
+  inner: 2,
+
+  pagination: reads('collection.pagination'),
 
   prevPageNumber: computed('pagination.{currentPage,isFirst}', function () {
-    let page = this.get('pagination.currentPage');
-    let isFirst = this.get('pagination.isFirst');
-
+    const { currentPage, isFirst } = this.pagination;
     if (!isFirst) {
-      return page - 1;
+      return currentPage - 1;
     }
-    return undefined;
   }),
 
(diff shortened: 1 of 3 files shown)


app/components/pagination-navigation.js
```diff
@@ -1,57 +1,44 @@
...
-      let numberOfPages = this.get('pagination.numberOfPages');
-      let perPage = this.get('pagination.perPage');
-      let currentPage = this.get('pagination.currentPage');
-      let offset = this.get('pagination.offset');
-      let innerWindow = this.get('innerWindow');
-      let outerWindow = this.get('outerWindow');
-
-      let thresholdDisplayAll = ((outerWindow + 1) * 2) + (innerWindow + 1);
+      const { outer, inner, pagination } = this;
+      const { numberOfPages, perPage, currentPage, offset } = pagination || {};
```
sorta: More condensed and readable. Love it!

app/styles/app/modules/pagination.scss
```diff
@@ -5,7 +5,7 @@
   ul {
     display: flex;
     align-items: baseline;
-    flex-wrap: wrap;
+    flex-wrap: nowrap;
```
sorta: Hmm yeah the way it was wrapping before didn't look great, but doesn't this cause the outer edges of the pagination to go offscreen when there are a lot of items on a narrow screen? For example https://as-pagination-refactoring.test-deployments.travis-ci.org/travis-ci

artursmirnov: Hmm, you're right 🤔But I think it should rather collapse than wrap.. I'll think about that.

artursmirnov: It appears that the best way would be to change `inner` and `outer` parameters depending on window dimensions. But this seems like a significant change, so I'll just revert it to wrap for now and file a separate issue.

sorta: Cool that sounds good to me. I agree that the current wrapping isn't great and some sort of collapse sounds like a good solution. But yeah that sounds like a significant change, so moving it into a separate issue sounds good 👍 Thanks!

PR #1979, opened 2019-03-19
[MOSS] migrate tab v2

artursmirnov:
Issue: travis-pro/team-teal#2915

This is new implementation according to the new requirements.


artursmirnov:
Hey @sorta, even though, there could be some more changes (hopefully minor 🤞) due to feedback implementation, could you please take a look when you have time?

sorta:
Hi @artursmirnov, sure thing! I'll start taking a look at this

app/styles/app/modules/popup.scss
```diff
@@ -64,6 +64,20 @@
...
+  padding: 0;
+
+  .icon {
+    margin: 0;
+  }
+}
+
+.modal-controls {
+  text-align: right;
+  margin-top: 2em;
```
sorta: I would normally prefer `rem` units for stuff like this, but I do notice that directly above and elsewhere in this file there's a lot of usage of `em` units, so I think in this case it makes sense to leave this as-is for consistency. And I can also definitely imagine cases where it is preferable to use `em` units.

But it has me thinking about the overall unit usage strategy for the codebase. I'm not sure if we have any existing guidelines for that kind of thing, but if not it might be worth working as a team to come up with some.

app/mixins/route/owner/migrate.js
```diff
@@ -0,0 +1,34 @@
...
+
+  model({ page }) {
+    this.setProperties({ page });
+  },
+
+  afterModel(model) {
+    if (model && !model.error) {
+      const repos = model.githubAppsRepositoriesOnOrg;
+      repos.switchToPage(this.page);
+      repos.on(PAGE_CHANGED, page => {
```
sorta: Neat 😄

.eslintrc.js
```diff
@@ -258,9 +258,6 @@ module.exports = {
     // enforce operators to be placed before or after line breaks
     'operator-linebreak': 0,
 
-    // enforce padding within blocks
-    'padded-blocks': [2, 'never'],
-
```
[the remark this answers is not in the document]
artursmirnov: Yep, right, this removes the limitation. But you're right, probably it makes sense to tweak the rule instead of deleting it.

app/styles/app/modules/popup.scss
artursmirnov: Oh, it's a good point! Thanks! I don't think there's any consistency wrt units in our current styles 😅But it totally makes sense to relate such dimensional values to base font size instead of immediate value 👍I also think that we should try to improve things as we go, so I'll definitely change this one.

sorta: Cool, that sounds good to me 😄 And yep, right now there is no consistency on it haha

artursmirnov: Fixed ✔️

.eslintrc.js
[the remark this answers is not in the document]
artursmirnov: Fixed ✔️

artursmirnov:
Hey @sorta, we're aiming to release this PR tomorrow. Could you please take another look at the recent changes?

sorta:
Hi @artursmirnov, absolutely can do :+1: I'll start taking a look at this

artursmirnov:
Thanks for the review, @sorta 👍

PR #1983, opened 2019-03-20
Add red button

sorta:
Modeled after `.button--green`. We have buttons of various colors, why not red as well? Split off from https://github.com/travis-ci/travis-web/pull/1953

### DIFF
--- a/app/styles/app/modules/buttons.scss
+++ b/app/styles/app/modules/buttons.scss
@@ -181,6 +181,18 @@
   }
 }
 
+.button--red {
+  @extend %button;
+
+  border: none;
+  color: white;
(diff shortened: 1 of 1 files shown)


sorta:
Yes! That would be nice :) I hope that eventually we'll be able to work on refactoring stuff like that and addressing tech debt

artursmirnov:
I don't think we will ever have dedicated time for tech debt. At least I would not wait for it. I'm generally more in favour of constant improvement, like if you work on the code and you feel that you can improve it - do so. Here's a good talk on the topic by Sarah Mai at EmberConf 2017 https://www.youtube.com/watch?v=UDp6eDCi1Vo

PR #1984, opened 2019-03-22
Add Insights tabs component

sorta:

Adds an insights tabs component for selecting intervals.

### DIFF
--- a/app/components/insights-tabs.js
+++ b/app/components/insights-tabs.js
@@ -1,23 +1,13 @@
 import Component from '@ember/component';
 import { DEFAULT_INSIGHTS_INTERVAL, INSIGHTS_INTERVALS } from 'travis/services/insights';
 
-export const INSIGHTS_TABS = [
-  { slug: INSIGHTS_INTERVALS.WEEK, title: INSIGHTS_INTERVALS.WEEK.capitalize() },
-  { slug: INSIGHTS_INTERVALS.MONTH, title: INSIGHTS_INTERVALS.MONTH.capitalize() },
-];
+export const INSIGHTS_TABS = Object.values(INSIGHTS_INTERVALS).map(slug => ({ slug, title: slug.capitalize() }));
 
 export default Component.extend({
   tagName: 'ul',
   classNames: ['travistab-nav travistab-nav--insights insights-tabs'],
 
   selectedTab: DEFAULT_INSIGHTS_INTERVAL,
-  setTab: () => {},
 
   tabs: INSIGHTS_TABS,
-
-  actions: {
-    setInsightTab(selection) {
-      this.setTab(selection);
-    }
-  }
 });
--- a/app/components/insights-tabs.js
+++ b/app/components/insights-tabs.js
@@ -1,0 +1,23 @@
+import Component from '@ember/component';
+import { DEFAULT_INSIGHTS_INTERVAL, INSIGHTS_INTERVALS } from 'travis/services/insights';
+
+export const INSIGHTS_TABS = [
+  { slug: INSIGHTS_INTERVALS.WEEK, title: INSIGHTS_INTERVALS.WEEK.capitalize() },
+  { slug: INSIGHTS_INTERVALS.MONTH, title: INSIGHTS_INTERVALS.MONTH.capitalize() },
+];
+
+export default Component.extend({
+  tagName: 'ul',
+  classNames: ['travistab-nav travistab-nav--insights insights-tabs'],
+
+  selectedTab: DEFAULT_INSIGHTS_INTERVAL,
+  setTab: () => {},
+
+  tabs: INSIGHTS_TABS,
+
+  actions: {
+    setInsightTab(selection) {
+      this.setTab(selection);
+    }
+  }
+});
--- a/app/components/insights-tabs.js
+++ b/app/components/insights-tabs.js
@@ -1,13 +1,11 @@
 import Component from '@ember/component';
-import { DEFAULT_INSIGHTS_INTERVAL, INSIGHTS_INTERVALS } from 'travis/services/insights';
+import { INSIGHTS_INTERVALS } from 'travis/services/insights';
 
 export const INSIGHTS_TABS = Object.values(INSIGHTS_INTERVALS).map(slug => ({ slug, title: slug.capitalize() }));
 
 export default Component.extend({
   tagName: 'ul',
   classNames: ['travistab-nav travistab-nav--insights insights-tabs'],
-
-  selectedTab: DEFAULT_INSIGHTS_INTERVAL,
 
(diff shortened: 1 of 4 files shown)


app/components/insights-tabs.js
```diff
@@ -0,0 +1,23 @@
+import Component from '@ember/component';
+import { DEFAULT_INSIGHTS_INTERVAL, INSIGHTS_INTERVALS } from 'travis/services/insights';
+
+export const INSIGHTS_TABS = [
+  { slug: INSIGHTS_INTERVALS.WEEK, title: INSIGHTS_INTERVALS.WEEK.capitalize() },
```
artursmirnov: I would suggest to capitalize titles with css `text-transform: uppercase`. In this case, `INSIGHTS_TABS` could be simplified to plain array and thus to:
```javascript
INSIGHTS_TABS = Object.values(INSIGHTS_INTERVALS)
```

app/components/insights-tabs.js
```diff
@@ -0,0 +1,23 @@
...
+  { slug: INSIGHTS_INTERVALS.MONTH, title: INSIGHTS_INTERVALS.MONTH.capitalize() },
+];
+
+export default Component.extend({
+  tagName: 'ul',
+  classNames: ['travistab-nav travistab-nav--insights insights-tabs'],
+
+  selectedTab: DEFAULT_INSIGHTS_INTERVAL,
+  setTab: () => {},
```
artursmirnov: This could be just `setTab() {}`. Your variant works too, this one just a bit shorter 😄

app/templates/components/insights-tabs.hbs
```diff
@@ -0,0 +1,7 @@
+{{#each tabs as |tab|}}
+  <li class="insights-tab {{if (eq tab.slug selectedTab) 'active'}}">
+    <a data-tab={{tab.slug}} {{action 'setInsightTab' tab.slug}}>
```
artursmirnov: Would it make sense to bind tabs to query params? In this case you would be able to make the link with 
```handlebars
{{#link-to (query-params interval=tab.slug) }}
  {{tab.title}}
{{/link-to}}
```
Here's an example: https://github.com/travis-ci/travis-web/blob/master/app/templates/components/profile-nav.hbs#L57-L93

Also, it's recommended to use closure actions instead of old bubbling ones: 
```javascript
onclick={{action 'setInsightTab' tab.slug}}`
```

app/components/insights-tabs.js
sorta: In this case I kind of wanted the text markup itself to be capitalized rather than using `text-transform`, but yeah maybe that's a little silly.

I also manually constructed the array to ensure the correct order, not realizing that object key order can now be more officially relied upon in ES6+ https://stackoverflow.com/questions/5467129/sort-javascript-object-by-key/31102605#31102605 huh TIL

artursmirnov: It's totally up to you, that was just an idea, that I had during the review 😄 If you feel, that current implementation works better - that's absolutely okay! We're all senior developers and don't require guidance. The code reviews are generally aim to bring another point of view and to share knowledge, that all of us have a lot of, but maybe in different areas.

sorta: Absolutely 😄 Although, impostor syndrome being what it is, I sometimes don't _feel_ like a senior dev haha.

I think what I'm gonna end up doing in this case is `INSIGHTS_TABS = Object.values(INSIGHTS_INTERVALS).map(slug => ({ slug, title: slug.capitalize() }));`. It is nice to loop through rather than have another manually defined array, but it still keeps the capitalization within the markup.

I have found your review comments very helpful and I've learned a lot from them. 😄 Especially in terms of Ember stuff, which I didn't really use much before joining Travis CI.

app/components/insights-tabs.js
sorta: Nice, I like shorter :) Although, with the note below about binding to query params, I guess we may not even need this

app/templates/components/insights-tabs.hbs
sorta: Ooh yes, binding to query params makes sense to me, I think I'll switch to that.

sorta: Oh and good to know about closure actions 👍 Thanks!

sorta:
Thanks for the review @artursmirnov ! It should be ready for another look when you get a chance :)

app/templates/components/insights-tabs.hbs
```diff
@@ -0,0 +1,7 @@
+{{#each tabs as |tab|}}
+  <li class="insights-tab {{if (eq tab.slug selectedTab) 'active'}}">
+    {{#link-to (query-params timeInterval=tab.slug) }}
```
artursmirnov: I'm not sure if it would work in this case, but worth mentioning that when the query parameter is active on URL, the link-to helper adds active class to the link automatically. Could it be used here instead of manually setting active on li?

sorta: Oooh nice, I'll test it out and see if it would work in this case.

sorta: It works, but breaks the component test that checks that active states are set correctly. Do you know of a way to set up or pass in query params for such a test? Or maybe it would make sense to move that test out to the acceptance test for the route? I'm leaning toward moving it out to the acceptance test, wdyt?

sorta: Yeah I really like having the query-params helper automatically handle stuff, so I think that's what I'm gonna go with.

artursmirnov: Query parameters are only available on route and constructor level. So you could get the `timeInterval` property value on constructor and pass it to the component (in case if it needs to know the state). In test you could assign this property directly:

```
  queryParams: ['timeInterval'],
  timeInterval: DEFAULT_INSIGHTS_INTERVAL
});
```
```
  tagName: 'ul',
  classNames: ['travistab-nav travistab-nav--insights insights-tabs'],

   tabs: INSIGHTS_TABS,
   activeTab: DEFAULT_INSIGHTS_INTERVAL
});
```
```
module('Integration | Component | insights-tabs', function (hooks) {
  setupRenderingTest(hooks);

   test('display', async function (assert) {
    await render(hbs`{{insights-tabs activeTab='month' }}`);

     assert.dom('.insights-tabs').exists();
    assert.dom('.insights-tabs .insights-tab').exists({ count: Object.values(INSIGHTS_INTERVALS).length });
    assert.dom('.insights-tabs .insights-tab .active').hasText('month');
  });
});
```

Although, I have noticed that you got rid of this logic entire, which works as well :)

sorta: Yeah that's what I suspected. Thanks!

PR #1985, opened 2019-03-22
Add insights date display component

sorta:

Adds an insights component for displaying a date range.

### DIFF
--- a/app/components/insights-date-display.js
+++ b/app/components/insights-date-display.js
@@ -1,0 +1,32 @@
+import Component from '@ember/component';
+import { inject as service } from '@ember/service';
+import { computed } from '@ember/object';
+import moment from 'moment';
+import { DEFAULT_INSIGHTS_INTERVAL } from 'travis/services/insights';
+
+export default Component.extend({
+  tagName: 'span',
+  classNames: ['insights-dates'],
+
+  insights: service(),
+
+  interval: DEFAULT_INSIGHTS_INTERVAL,
+
+  dates: computed('interval', function () {
+    return this.insights.getDatesFromInterval(this.interval);
+  }),
+
+  startDate: computed('dates', function () {
+    if (this.dates.length === 2) {
+      return moment(this.dates[0]).format('MMMM DD, YYYY');
+    }
+    return '...';
+  }),
+
+  endDate: computed('dates', function () {
+    if (this.dates.length === 2) {
+      return moment(this.dates[1]).format('MMMM DD, YYYY');
+    }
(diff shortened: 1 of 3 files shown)


app/components/insights-date-display.js
```diff
@@ -0,0 +1,32 @@
...
+export default Component.extend({
+  tagName: 'span',
+  classNames: ['insights-dates'],
+
+  insights: service(),
+
+  interval: DEFAULT_INSIGHTS_INTERVAL,
+
+  dates: computed('interval', function () {
+    return this.insights.getDatesFromInterval(this.interval);
```
artursmirnov: 👍

app/components/insights-date-display.js
```diff
@@ -0,0 +1,32 @@
...
+  }),
+
+  startDate: computed('dates', function () {
+    if (this.dates.length === 2) {
+      return moment(this.dates[0]).format('MMMM DD, YYYY');
+    }
+    return '...';
```
artursmirnov: You could use `dates.firstObject` and `date.lastObject` here:

```javascript
startDate: computed('dates.firstObject', function () {
  const { dates = [] } = this;
  return dates.firstObject ? moment(dates.firstObject).format('MMMM DD, YYYY') : '...';
})
```

artursmirnov: Otherwise, the property should depend on `dates.[]` - because it depends not only on `dates` property itself, but also on its content.

app/components/insights-date-display.js
```diff
@@ -0,0 +1,32 @@
...
+
+  interval: DEFAULT_INSIGHTS_INTERVAL,
+
+  dates: computed('interval', function () {
+    return this.insights.getDatesFromInterval(this.interval);
+  }),
+
+  startDate: computed('dates', function () {
+    if (this.dates.length === 2) {
+      return moment(this.dates[0]).format('MMMM DD, YYYY');
```
artursmirnov: I would suggest to extract format string to a constant as it seems to be reused in multiple times, including test.

app/components/insights-date-display.js
sorta: Ahh cool, that makes sense! Updated to use `dates.firstObject`

app/components/insights-date-display.js
sorta: Good idea, change made :)

sorta:
Hi @artursmirnov thanks for the review! This should be ready for another look whenever you get a chance 😄

PR #1986, opened 2019-03-22
Add build-count and insights-glance components

sorta:

This adds a build-count component that generates a graph displaying build count data, along with some stats. It is one of 4 very similar components, which all utilize the insights-glance component, which handles a lot of the shared functionality.

### DIFF
--- a/app/components/build-count.js
+++ b/app/components/build-count.js
@@ -1,0 +1,94 @@
+import Component from '@ember/component';
+import { inject as service } from '@ember/service';
+import { computed } from '@ember/object';
+import { pluralize } from 'ember-inflector';
+import { reads, equal, or } from '@ember/object/computed';
+import { task } from 'ember-concurrency';
+import { DEFAULT_INSIGHTS_INTERVAL } from 'travis/services/insights';
+
+export default Component.extend({
+  classNames: ['insights-build-count'],
+  private: false,
+  interval: DEFAULT_INSIGHTS_INTERVAL,
+  owner: null,
+
+  insights: service(),
+
+  // Current Interval Chart Data
+  requestData: task(function* () {
+    return yield this.insights.getChartData.perform(
+      this.owner,
+      this.interval,
+      'builds',
+      'sum',
+      ['count_started'],
+      {
+        calcAvg: true,
+        private: this.private,
+      }
+    );
+  }),
+  chartData: reads('requestData.lastSuccessful.value'),
+  builds: reads('chartData.data.count_started.plotValues'),
+  labels: reads('chartData.labels'),
+
+  isLoading: reads('requestData.isRunning'),
+  isEmpty: equal('totalBuilds', 0),
+  showPlaceholder: or('isLoading', 'isEmpty'),
+
+  // Total / average
+  totalBuilds: reads('chartData.data.total'),
+  avgBuilds: reads('chartData.data.average'),
+
+  totalBuildText: computed('totalBuilds', function () {
+    if (typeof this.totalBuilds !== 'number') { return '\xa0'; }
+    return this.totalBuilds.toLocaleString();
+  }),
+
+  // Previous interval chart data
+  requestPastData: task(function* () {
+    return yield this.insights.getChartData.perform(
+      this.owner,
+      this.interval,
+      'builds',
+      'sum',
+      ['count_started'],
+      { startInterval: -2, endInterval: -1, calcTotal: true, calcAvg: true, private: this.private }
+    );
+  }),
+  pastIntervalData: reads('requestPastData.lastSuccessful.value'),
+  prevTotalBuilds: reads('pastIntervalData.data.count_started.total'),
+
+  // Percent change
+  percentChangeTitle: computed('prevTotalBuilds', 'interval', function () {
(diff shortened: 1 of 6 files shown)


app/components/insights-glance.js
```diff
@@ -0,0 +1,87 @@
...
+
+  title: '',
+  statistic: '',
+
+  delta: 0,
+  deltaTitle: '',
+  deltaText: '',
+
+  labels: [],
+  values: [],
```
artursmirnov: Properties of anything inherited from `Ember.Object` or `Ember.Mixin` (so basically literally anything in Ember 😆) should never be initialized by values of complex types (arrays, objects), because it may cause leaking state between instances. Here are more details on the topic: https://dockyard.com/blog/2015/09/18/ember-best-practices-avoid-leaking-state-into-factories

The idea is to either set defaults to such properties in `init` or use computed properties, e.g.:
```javascript
values: computed(() => [])
```

app/components/insights-glance.js
```diff
@@ -0,0 +1,87 @@
...
+  tooltip: {
+    position: (data, width, height, element) => {
+      let top = -50;
+      let left = (element.getAttribute('width') - width) / 2;
+      return ({ top, left });
+    },
+    format: {
+      value: d3format(','),
+    }
+  },
```
artursmirnov: The previous comment also applies to all of the properties above that are initialized with objects. I would suggest to wrap them to computed properties.

app/components/insights-glance.js
```diff
@@ -0,0 +1,87 @@
...
+
+  labels: [],
+  values: [],
+  datasetTitle: 'Data',
+  centerline: null,
+
+  showPlaceholder: or('isLoading', 'isEmpty'),
+
+  // Chart component data
+  data: computed('values', 'labels', 'datasetTitle', function () {
```
artursmirnov: I think the property depends on the content of the arrays, so it should be `'values.[]', 'labels.[]'`

app/templates/components/insights-glance.hbs
```diff
@@ -0,0 +1,27 @@
...
+      {{loading-indicator}}
+    {{else}}
+      {{statistic}}
+    {{/if}}
+  </div>
+  {{#if delta}}
+    <div
+      class="insights-glance-delta"
+      data-dir="{{if (gte delta 0) '+' '-'}}"
+      title="{{deltaTitle}}"
```
artursmirnov: It's generally not necessary to wrap handlebars placeholders into quotes, compiler takes care of that.

sorta: Ahh nice, good to know :) Change made

app/components/insights-glance.js
sorta: Ahh cool, that makes sense! Change made, thanks 😄

app/components/insights-glance.js
sorta: Ahh okay, that's good to know! Thanks for the explanation and link, that makes a lot of sense. I'll go ahead and wrap these in computed properties 😄

app/components/insights-glance.js
sorta: Change made :)

sorta:
Thanks for the review @artursmirnov ! This should be ready for another look when you have time 😄

app/components/build-count.js
```diff
@@ -0,0 +1,94 @@
...
+  requestPastData: task(function* () {
+    return yield this.insights.getChartData.perform(
+      this.owner,
+      this.interval,
+      'builds',
+      'sum',
+      ['count_started'],
+      { startInterval: -2, endInterval: -1, calcTotal: true, calcAvg: true, private: this.private }
+    );
+  }),
```
artursmirnov: I would suggest to add `drop` modifier to this task, because it's getting called whenever component arguments change, which may happen often, and in this case, it will hit API multiple times for the same data. Here's the link that explains how concurrency modifiers work: http://ember-concurrency.com/docs/task-concurrency

app/components/build-count.js
```diff
@@ -0,0 +1,94 @@
...
+      this.interval,
+      'builds',
+      'sum',
+      ['count_started'],
+      {
+        calcAvg: true,
+        private: this.private,
+      }
+    );
+  }),
```
artursmirnov: And here it also may make sense to add a concurrency modifier

app/components/insights-glance.js
```diff
@@ -0,0 +1,87 @@
...
+      ],
+      colors: {
+        [this.datasetTitle]: '#666',
+      },
+    };
+  }),
+
+  // Chart component options
+  legend: computed(() => ({ show: false })),
+  size: computed(() => ({ height: 50 })),
```
artursmirnov: I like how you wrapped the body with braces 👍

app/components/build-count.js
sorta: Nice, that makes sense. Added 😄

app/components/build-count.js
sorta: Change made :)

sorta:
Wow that was fast 🚀 Thanks again for reviewing! It should be ready for another look when you have time 😄

PR #1990, opened 2019-03-26
Responsive repository header

weban:
Fix for issue 2841
https://github.com/travis-pro/team-teal/issues/2841

### DIFF
--- a/app/styles/app/layouts/repo.scss
+++ b/app/styles/app/layouts/repo.scss
@@ -28,13 +28,6 @@
   .repo-title {
     margin-right: 1.2em;
 
-    @media #{$small-only} {
-      margin-bottom: 0.5em;
-      h1 {
-        font-size: 24px!important;
-      }
-    }
-
     h1 {
       margin: 0;
       font-size: 36px;
@@ -49,6 +42,20 @@
         &:active {
           border-color: $asphalt-grey;
         }
+      }
+    }
+
+    @media #{$small-only} {
+      margin-bottom: 1em;
+      width: 100%;
+
+      h1 {
+        font-size: 20px;
+        white-space: nowrap;
+        overflow: hidden;
+        text-overflow: ellipsis;
+        display: inline-block;
+        width: calc(100% - 40px);
       }
     }
   }
@@ -68,9 +75,10 @@
   display: flex;
 
   @media #{$small-only} {
+    margin-top: 0.5em;
     flex-flow: column-reverse;
     .travistab-nav li {
-      margin-right: 1em;
+      margin-right: 0.5em;
     }
   }
 
@@ -82,6 +90,12 @@
 
   .option-button {
     width: 10em;
+
+    @media #{$small-only} {
+      position: absolute;
+      top: 45px;
+      right: 0;
+    }
 
     .option-dropdown a {
       padding-left: 2.4em;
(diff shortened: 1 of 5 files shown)


weban:
@artursmirnov thanks for the review. What is the min page with that we should support

artursmirnov:
@weban currently we don't have any rules about supported browsers / screen sizes. We have been thinking about the need to define them for a long time now, but never got to that point yet. Generally, I strive to check how it looks in the major mobile viewports. Chrome DevTools suggests them out of the box:
image

Hopefully, some day we'll define them based on the target audience analysis 🤞

sorta:
@artursmirnov In the past I've sometimes done truncation with `...`, but more recently I've felt that I would rather see all of the text even if it wraps/breaks. I'm generally not a personal fan of side scrolling, but I'm not really opposed to it either. I usually tend to lean towards avoiding wrapping when possible, but if necessary then at least trying to make things wrap gracefully. 

In the case of the header, a user with a long user name and a long repo title will run out of space sooner, so at some point it will have to wrap or scroll. If you put `display: flex` on `.repo-title`, then you at least get only the text breaking and not wrapping away from the icon, which I think looks nicer.

For the tabs, I think it is a little weird for tabs specifically to be on multiple lines, so yeah in that case I think a side scrolling solution could be the way to go. I could see the [Priority Navigation pattern](https://css-tricks.com/the-priority-navigation-pattern/) working well too, but maybe it's a bit much for something that only wraps at pretty small widths.

And yes, there's a bunch of stuff like supported browsers policy that tends to end up on the backburner. Once Insights launches I might take a little time to at least make sure that we have team issues for some of those things.

app/styles/app/layouts/repo.scss
```diff
@@ -24,20 +24,31 @@
...
-
-      &:hover,
-      &:active {
-        border-color: $asphalt-grey;
+    margin-right: 1.2em;
+
+    @media #{$small-only} {
+      margin-bottom: 0.5em;
+      h1 {
+        font-size: 24px!important;
```
sorta: I try to avoid use of `!important` whenever possible. It's a big jump in specificity that could make the codebase harder to maintain. 

An easy way to address this could be to move this media query below the `h1` definition a few lines down, since stuff defined later takes precedence. In this case you could split it into 2 separate media queries or keep it as 1, whichever makes sense to you.

You could also add a class to raise the specificity of this definition without resorting to `!important`, but that seems a little weirder to me.

app/styles/app/modules/loading-indicator.scss
```diff
@@ -49,10 +49,6 @@
   padding: 0;
 }
 
-.inline-block {
-  display: inline-block;
-}
-
```
sorta: Nice, I think it makes a lot of sense to move this out of here and into base.scss 👍

app/styles/app/layouts/repo.scss
sorta: Oh here are some links on this issue:
https://uxengineer.com/css-specificity-avoid-important-css/
https://css-tricks.com/strategies-keeping-css-specificity-low/

weban:
Please check current version and decide with one is better - i mean cut title with ... or keep original title.

weban:
Thanks for the review!

artursmirnov:
@weban changes has been deployed to production 👍

PR #1995, opened 2019-04-02
[MOSS] beta ui

artursmirnov:
Issue: https://github.com/travis-pro/team-teal/issues/2903


artursmirnov:
@sorta thanks for the review! I have pushed a bit more changes with feedback implementation. Could you please take a look once you have some time?

app/models/owner.js
```diff
@@ -63,25 +64,23 @@ export default Model.extend({
...
-  migrationBetaRequests: computed(function () {
-    return this.store.peekAll('beta-migration-request').filter(request =>
-      request.owner === this || request.organizations.includes(this)
-    );
-  }).volatile(),
+  fetchBetaMigrationRequestsTask: task(function* () {
+    const data = yield this.ajax.getV3(`/user/${this.accounts.user.id}/beta_migration_requests`);
+    this.store.pushPayload('beta-migration-request', data);
+    return this.store.peekAll('beta-migration-request');
+  }),
```
sorta: Nice, I think switching this to an ember-concurrency task makes a lot of sense :+1:

artursmirnov: Yeah, `volatile` properties don't recalculate when `store` changes, so it caused issues. It was a weird decision from the beginning 😆

artursmirnov:
@sorta thanks for review!

app/models/owner.js
sorta: Haha yeah I've been there for sure. You write something in a way that makes sense in the moment, and then later look back and go "Why on Earth did I do it this way?" :laughing:

artursmirnov:
Hey @sorta, we're going to release this PR soon, could you please take another look to it once you have some time?

sorta:
Absolutely can do @artursmirnov! I'll take a look at this

app/services/flashes.js
```diff
@@ -55,33 +55,18 @@ export default Service.extend({
...
+      const type = Object.keys(flash)[0];
+      const { message, preamble = messageTypeToPreamble[type], aboveOverlay } = flash[type];
+      const icon = messageTypeToIcon[type];
+      const closeButton = messageTypeToCloseButton[type];
+      const item = { type, message, icon, preamble, closeButton, aboveOverlay };
+
+      this.flashes.unshiftObject(item);
+
+      if (!closeButton) this.removeFlash(item);
+    });
```
sorta: This is much nicer, I like it! :+1:

PR #1996, opened 2019-04-04
Reinstate day of week on tooltips for the build status chart

sorta:
Fairly straightforward, addresses this issue: https://github.com/travis-ci/travis-insights/issues/71

Not urgent to review, whenever you have time for it is fine 👍

### DIFF
--- a/app/components/build-status-chart.js
+++ b/app/components/build-status-chart.js
@@ -86,7 +86,7 @@ export default Component.extend({
   axis: computed(() => ({
     x: {
       type: 'timeseries',
-      tick: { format: '%b %e' },
+      tick: { format: '%A, %b %e' },
     },
     y: {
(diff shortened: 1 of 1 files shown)


sorta:
Just noticed that percy is catching some stuff that's always going to be different. I might look into preventing that https://docs.percy.io/docs/percy-specific-css

artursmirnov:
Oh, that seems to be caused by dynamically generated data in tests 🤔

sorta:
Yes, although the dynamic data in the tests _should_ always result in the same pattern, because it just loops through an array of preset values. But what's unavoidable is that the dates will always be different. Anyway, I've set up a separate PR for looking into solving the snapshot issue.

sorta:
Thanks! 😄

PR #1997, opened 2019-04-04
Add a way to hide things from percy snapshots. Apply to insights.

sorta:
I noticed in https://github.com/travis-ci/travis-web/pull/1996 that Percy was flagging for review when nothing in the snapshot should have really been different. It seems due to changing dates on the x-axis as well as tiny variations in drawing the line charts. This PR should address that and also provides a way to hide other things from snapshots if we need to in the future.

Reference: https://docs.percy.io/docs/percy-specific-css

### DIFF
--- a/app/styles/app/layouts/insights.scss
+++ b/app/styles/app/layouts/insights.scss
@@ -294,12 +294,12 @@ $med-small-up: "only screen and (min-width: 38em)";
 
   // Odyssey C3 customizations
   .c3-axis {
-    @include snapshotHide;
-
     .domain {
       stroke: $axis-line-color;
     }
     .tick {
+      @include snapshotHide;
+
       line {
         stroke: $axis-tick-color;
       }
--- a/app/styles/app/layouts/insights.scss
+++ b/app/styles/app/layouts/insights.scss
@@ -294,6 +294,8 @@ $med-small-up: "only screen and (min-width: 38em)";
 
   // Odyssey C3 customizations
   .c3-axis {
+    @include snapshotHide;
+
     .domain {
       stroke: $axis-line-color;
     }
--- a/app/templates/components/insights-glance.hbs
+++ b/app/templates/components/insights-glance.hbs
@@ -22,6 +22,6 @@
   {{#if showPlaceholder}}
     <hr class="insights-glance__chart-placeholder" data-test-insights-glance-chart-placeholder />
   {{else}}
-    {{c3-chart class='chart-component' data=data axis=axis legend=legend grid=grid size=size point=point tooltip=tooltip}}
+    {{c3-chart class='chart-component snapshot-hide' data=data axis=axis legend=legend grid=grid size=size point=point tooltip=tooltip}}
   {{/if}}
 </div>
--- a/app/components/insights-date-display.js
+++ b/app/components/insights-date-display.js
@@ -9,7 +9,7 @@ export const INSIGHTS_DATE_RANGE_PLACEHOLDER = '...';
 
 export default Component.extend({
   tagName: 'span',
-  classNames: ['insights-dates'],
+  classNames: ['insights-dates', 'snapshot-hide'],
 
   insights: service(),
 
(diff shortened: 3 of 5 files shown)


app/styles/app/layouts/insights.scss
```diff
@@ -1,339 +1,341 @@
...
+    }
+  }
+
+  // Odyssey C3 customizations
+  .c3-axis {
+    .domain {
+      stroke: $axis-line-color;
+    }
+    .tick {
+      @include snapshotHide;
```
sorta: This line here is the actual intended change for this file.

Ugh, I guess some of these files had CRLF line endings. I did some dev through Windows Subsystem for Linux, and I thought I had my editor set to always do LF line endings, but I guess not. Anyway, I've recently altered my setup such that I'm going to always be using my macbook. And I'm thinking I'll open up a PR to convert any other files that have CRLF line endings.

app/styles/app/layouts/insights.scss
```diff
@@ -1,339 +1,341 @@
-$axis-line-color: $dry-cement;
```
sorta: Ugh, so I guess some of these files had CRLF line endings. I did some dev through Windows Subsystem for Linux, and I thought I had my editor set to always do LF line endings, but I guess not. Anyway, I've recently altered my setup such that I'm going to always be using my macbook. And I'm thinking I'll open up a PR to convert any other files that have CRLF line endings.

app/templates/components/insights-glance.hbs
```diff
@@ -1,27 +1,27 @@
...
+      {{svg-jar 'icon-dropdown-arrow' class="icon"}}
+      <span class="insights-glance-delta__stat" data-test-insights-glance-delta-stat>{{deltaText}}</span>
+    </div>
+  {{/if}}
+</div>
+<div class="insights-glance__chart" data-test-insights-glance-chart>
+  {{#if showPlaceholder}}
+    <hr class="insights-glance__chart-placeholder" data-test-insights-glance-chart-placeholder />
+  {{else}}
+    {{c3-chart class='chart-component snapshot-hide' data=data axis=axis legend=legend grid=grid size=size point=point tooltip=tooltip}}
```
sorta: The intended change for this file is to add this class `snapshot-hide`.

app/styles/app/layouts/insights.scss
artursmirnov: How did it go with WSL? The last attempt I found source maps to not being working properly because of different filesystem root (I guess it was something like `/c/something` inside VM vs `C:\something` outside)

sorta: It went fine, a lot better than previous experiences I've had developing non-windows stuff on windows hah. I didn't notice a source maps issue, but I wasn't paying close attention to it and that makes sense, I know it does some weird stuff with filesystem paths.

The problem that I ran into is that it was slow, at least for what we do.  Ember CLI build/serve took a much longer time to finish than on OSX/linux. I've heard that filesystem operations can be slow on WSL, especially if there are a lot of them at once.

sorta:
Thanks! 😄 

Yeah I noticed that, It's weird. But I checked just now and charts look fine in FF, IE, Edge, Safari. I know the C3 component tries to size itself, maybe it doesn't do it quickly enough for Percy?

app/styles/app/layouts/insights.scss
artursmirnov: Did you try this approach for Windows builds performance improvement https://cli.emberjs.com/release/appendix/windows/ ? Just enabling symlinks makes it much better, even close to Mac/Linux performance. The reason for that is that Broccoli (Ember CLI build pipeline tool) heavily uses filesystem. Basically, every single build step is a copy of the app and dependencies. On POSIX-compatible filesystems it uses symlinks, which are fast and transparent. But in Windows symlinks are not available without some magic, thus Broccoli has to COPY the entire app on each step. Also in this case `tmp` folder may take several Gb of space.

artursmirnov:
Thanks for checking!

app/styles/app/layouts/insights.scss
sorta: Oooh, neat! I tried a few different things, but I don't remember doing anything with symlinks. Maybe I'll give those suggestions a try tomorrow 😄

PR #1998, opened 2019-04-05
Update style

weban:
MS Edge fix for missing users dropdown

### DIFF
--- a/app/styles/app/modules/navigation.scss
+++ b/app/styles/app/modules/navigation.scss
@@ -240,8 +240,11 @@ $nav-line-height: 35px;
 }
 
 ul #profile-menu-link:hover > ul,
-ul #profile-menu-link:focus-within > ul,
 ul #profile-menu-link ul:hover {
+  display: block;
+}
+
+ul #profile-menu-link:focus-within > ul {
   display: block;
 }
 
--- a/app/templates/components/header-links.hbs
+++ b/app/templates/components/header-links.hbs
@@ -60,17 +60,15 @@
       {{#if auth.signedOut}}
         <button class="auth-button signed-out" type="button" {{action 'signIn'}}>Sign in with GitHub</button>
       {{/if}}
+      {{#if auth.signingIn}}
+        <button class="auth-button signing-in" type="button">Signing In {{loading-indicator inline=true}}</button>
+      {{/if}}
       {{#if auth.signedIn}}
         {{#link-to "account" data-test-header-account-link class="navigation-anchor navigation-profile-link signed-in"}}
           {{user-avatar url=user.avatarUrl name=user.fullName size=40}}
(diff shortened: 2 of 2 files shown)


app/styles/app/modules/navigation.scss
```diff
@@ -239,10 +239,10 @@ $nav-line-height: 35px;
...
 }
 
-ul #profile-menu-link:hover > ul,
-ul #profile-menu-link:focus-within > ul,
-ul #profile-menu-link ul:hover {
-  display: block;
+#profile-menu-link:hover {
+  ul {
+    display: block;
+  }
```
sorta: This seems like a sensible reorganization of these selectors. I'm guessing it was `:focus-within` that was breaking things for IE? Unfortunately, removing it seems to have broken keyboard navigation established in https://github.com/travis-ci/travis-web/pull/1787. I would prefer to have this working for both IE and keyboard navigation. I bet that adding a separate selector below this one for `:focus-within` would do the trick. So maybe something like:

```
#profile-menu-link {
  &:hover ul {
    display: block;
  }
  &:focus-within ul {
    display: block;
  }
}
```

weban: focus-within is a pseudo class and it doesnt work on ie
https://caniuse.com/#search=focus-within

sorta: Yeah I'm assuming that it won't work in IE/edge. But if we have it as a separate selector below, then at least it'll work for chrome/FF/safari. If you want you could wrap it in a `@supports` feature query, but I think it should be unnecessary, as browsers that don't understand it would ignore it. Does that make sense?

weban: IE/EDGE does not ignore it - style isn't correctly handled (is skipped) . I'll try to find some good solution

weban: @sorta you have right, thanks1

weban: In 2nd commit I removed duplicated condition for  auth.signedIn

sorta: Ahh ok cool, thanks! I'll take a look at this now.

app/templates/components/header-links.hbs
```diff
@@ -60,17 +60,15 @@
...
           {{user-avatar url=user.avatarUrl name=user.fullName size=40}}
           {{svg-jar 'icon-dropdown-arrow' class="icon"}}
         {{/link-to}}
-      {{/if}}
-      {{#if auth.signingIn}}
-        <button class="auth-button signing-in" type="button">Signing In {{loading-indicator inline=true}}</button>
-      {{/if}}
-      {{#if auth.signedIn}}
-      <div class="menu-separator"></div>
+        <div class="menu-separator"></div>
```
sorta: This reorganization makes a lot of sense to me 👍

PR #1999, opened 2019-04-05
Set line endings to LF

sorta:
Setting line endings to LF rather than CRLF. Nothing else should be changed in these files. And I _think_ this should be all of them.


app/styles/app/modules/visibility-setting-list.scss
```diff
@@ -1,103 +1,103 @@
...
-  }
-
-  &__controls .button {
-    margin: 0 1rem 0 0;
-  }
-}
\ No newline at end of file
+
+
+.visibility-setting-list {
```
artursmirnov: Is the spacing on the top intentional? I'm fine with it in general, just wanted to bring your attention here, in case it's not.

sorta: Huh yeah I don't remember when/how that came about. I don't think it's intentional, and it's unnecessary, so I might remove it.

sorta:
Thanks for the review! 😄

PR #2001, opened 2019-04-05
Add animation to modal dialogs

artursmirnov:
Extraction from https://github.com/travis-ci/travis-web/pull/1995

### DIFF
--- a/package.json
+++ b/package.json
@@ -89,6 +89,8 @@
     "emoji-js": "^3.4.0",
     "eslint-plugin-ember": "^5.2.0",
     "fuzzysort": "^1.0.1",
+    "liquid-fire": "^0.29.5",
+    "liquid-wormhole": "^2.1.5",
     "loader.js": "^4.7.0",
     "node-sass": "^4.9.3",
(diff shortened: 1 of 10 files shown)


package.json
```diff
@@ -89,6 +89,8 @@
     "emoji-js": "^3.4.0",
     "eslint-plugin-ember": "^5.2.0",
     "fuzzysort": "^1.0.1",
+    "liquid-fire": "^0.29.5",
+    "liquid-wormhole": "^2.1.5",
```
sorta: Neat, I hadn't heard of these 👍

artursmirnov: Yeah, Liquid Fire was a default tool to make Ember apps animated for years. It seems like during the last year it tends to be replaced by [Ember Animated](https://ember-animation.github.io/ember-animated/), which is basically its successor, created by the same author. Although, our current `modal-dialog` component out of the box supports `liquid-fire`, so I decided to go with it :)

artursmirnov:
@sorta thanks for the review! 👍

PR #2003, opened 2019-04-05
Increase form label text size

artursmirnov:
Issue: https://github.com/travis-ci/bugs-questions-comments-ideas/issues/119

### DIFF
--- a/app/styles/app/modules/forms.scss
+++ b/app/styles/app/modules/forms.scss
@@ -33,7 +33,7 @@ textarea {
 }
 
 .form-label {
-  font-size: 10px;
+  font-size: 1rem;
   color: $cement-grey;
   text-transform: uppercase;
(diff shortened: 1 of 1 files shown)


sorta:
Although, I just noticed that the overlay background seems out of place in Percy. It doesn't seem to happen in browser for me, but might be worth double checking.

artursmirnov:
@sorta thanks for review 😄

artursmirnov:
re: overlay, I think, it's because of the animation. Liquid Fire performs animations via Javascript, so I assume, that Percy takes screenshot during the animation. Let's see how it goes, and if it would show different diffs, then I'll disable animations for testing.

sorta:
Cool, yeah that makes sense and sounds like a good plan 👍

PR #2004, opened 2019-04-08
Use raw configs instead of yaml config endpoint

weban:
Issue #2941

### DIFF
--- a/app/components/raw-config.js
+++ b/app/components/raw-config.js
@@ -1,20 +1,18 @@
 import Component from '@ember/component';
 import { computed } from '@ember/object';
+import { later } from '@ember/runloop';
 
 export default Component.extend({
   copied: false,
 
   buttonLabel: computed('copied', function () {
-    let copied = this.get('copied');
-    if (copied) {
-      return 'Copied!';
-    }
-    return 'Copy to clipboard';
+    return this.get('copied') ? 'Copied!' : 'Copy to clipboard';
   }),
 
   actions: {
     copied() {
       this.set('copied', true);
+      later(() => this.set('copied', false), 3000);
     }
   }
 });
--- a/app/components/raw-config.js
+++ b/app/components/raw-config.js
@@ -1,0 +1,20 @@
+import Component from '@ember/component';
+import { computed } from '@ember/object';
+
+export default Component.extend({
+  copied: false,
+
+  buttonLabel: computed('copied', function () {
+    let copied = this.get('copied');
+    if (copied) {
+      return 'Copied!';
+    }
+    return 'Copy to clipboard';
+  }),
+
+  actions: {
+    copied() {
+      this.set('copied', true);
+    }
+  }
+});
--- a/app/components/raw-config.js
+++ b/app/components/raw-config.js
@@ -1,20 +1,34 @@
 import Component from '@ember/component';
 import { computed } from '@ember/object';
+import { inject as service } from '@ember/service';
+import { isInternal, presentedPath, fileNameWithoutSha } from 'travis/utils/format-config';
 import { later } from '@ember/runloop';
 
 export default Component.extend({
   copied: false,
+  baseYmlName: '.travis.yml',
+  externalLinks: service(),
 
   buttonLabel: computed('copied', function () {
-    return this.get('copied') ? 'Copied!' : `Copy ${this.fileName}`;
+    let source = this.get('rawConfig.source');
+    return this.get('copied') ? 'Copied!' : `Copy ${fileNameWithoutSha(source)}`;
   }),
 
-  fileName: computed('rawConfig', function () {
-    let sourcePath = this.get('rawConfig').source;
-    let versionedSourceName = sourcePath.substring(sourcePath.lastIndexOf('/') + 1);
-    if (!versionedSourceName.length) { return sourcePath; }
-    let sourceName = versionedSourceName.substring(0, versionedSourceName.lastIndexOf('@'));
-    return sourceName.length ? sourceName : versionedSourceName;
+  filePath: computed('rawConfig.source', 'slug', function () {
+    let source = this.get('rawConfig.source');
+    let name = fileNameWithoutSha(source);
+    if (name === this.baseYmlName) { return name; }
+
+    return presentedPath(source, this.get('slug'));
+  }),
+
+  fileUrl: computed('rawConfig.source', 'slug', 'build.branchName', function () {
+    let source = this.get('rawConfig.source');
+    let slug = this.get('slug');
+    if (isInternal(source, slug)) { return null; }
+
+    let branchName = this.get('build.branchName');
+    return this.get('externalLinks').githubFile(slug, branchName, fileNameWithoutSha(source));
   }),
 
   actions: {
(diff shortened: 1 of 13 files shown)


tests/acceptance/config/yaml-test.js
```diff
@@ -5,11 +5,17 @@ import { percySnapshot } from 'ember-percy';
...
+let config = `
 language: jortle
 sudo: tortle
 `;
 
+let rawConfigs = [
+  {
+    config: config
+  }
+];
```
artursmirnov: It would also be nice to have a test for multiple configs - to verify they all rendered properly

weban: Thanks for reviewm @artursmirnov . Updates coming soon:)

app/templates/components/header-links.hbs
```diff
@@ -65,12 +65,7 @@
           {{user-avatar url=user.avatarUrl name=user.fullName size=40}}
           {{svg-jar 'icon-dropdown-arrow' class="icon"}}
         {{/link-to}}
-      {{/if}}
-      {{#if auth.signingIn}}
-        <button class="auth-button signing-in" type="button">Signing In {{loading-indicator inline=true}}</button>
-      {{/if}}
-      {{#if auth.signedIn}}
```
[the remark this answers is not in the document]
weban: My mistake, didnt switch the branch.

app/components/raw-config.js
```diff
@@ -0,0 +1,20 @@
...
+
+export default Component.extend({
+  copied: false,
+
+  buttonLabel: computed('copied', function () {
+    let copied = this.get('copied');
+    if (copied) {
+      return 'Copied!';
+    }
+    return 'Copy to clipboard';
```
artursmirnov: I would suggest to shorten this entire method to be:
```js
return this.copied ? 'Copied!' : 'Copy to clipboard';
```

app/components/raw-config.js
```diff
@@ -0,0 +1,20 @@
...
+  }),
+
+  actions: {
+    copied() {
+      this.set('copied', true);
```
artursmirnov: I would suggest to revert it back to `false` after a few seconds:
```js
this.set('copied', true);
later(() => this.set('copied', false), 3000);
```
Here's the [later](https://api.emberjs.com/ember/3.9/functions/@ember%2Frunloop/later) docs

app/models/request.js
```diff
@@ -54,6 +45,11 @@ export default Model.extend({
     return eventType === 'pull_request';
   }),
 
+  uniqRawConfigs: computed('raw_configs', function () {
```
artursmirnov: This could be done with [computed macro](https://api.emberjs.com/ember/3.9/functions/@ember%2Fobject%2Fcomputed/uniqBy)
```js
uniqRawConfigs: uniqBy('raw_configs', 'source'),
```

app/components/raw-config.js
weban: Old moved code, but agree its worth to refactor

app/components/raw-config.js
```diff
@@ -0,0 +1,18 @@
+import Component from '@ember/component';
+import { computed } from '@ember/object';
+import { later } from '@ember/runloop';
+
+export default Component.extend({
+  copied: false,
+
+  buttonLabel: computed('copied', function () {
+    return this.get('copied') ? 'Copied!' : 'Copy to clipboard';
```
artursmirnov: Just as a note, `.get` is not mandatory anymore. In most cases you can just `this.copied`. In rare cases when `.get` is required, Ember will tell you so in browser console. No need to fix it here, we still have a lot of `.get`s all over the app, it's just for future reference.

artursmirnov:
@weban please merge the PR whenever it's ready for release to production.

app/components/raw-config.js
```diff
@@ -0,0 +1,40 @@
...
+import { isInternal, presentedPath, fileNameWithoutSha } from 'travis/utils/format-config';
+import { later } from '@ember/runloop';
+
+export default Component.extend({
+  copied: false,
+  baseYmlName: '.travis.yml',
+  externalLinks: service(),
+
+  buttonLabel: computed('copied', function () {
+    let source = this.get('rawConfig.source');
```
artursmirnov: This computed property seems to depend on `rawConfig.source` too. So it should be listed in dependent keys so that the property re-compute properly:
```js
buttonLabel: computed('copied', 'rawConfig.source', function () {
})
```

app/templates/components/raw-config.hbs
```diff
@@ -0,0 +1,20 @@
+<div class='inner-yaml-container'>
+  <div class='file-name'>
+    {{#if filePath}}
+      {{#external-link-to href=fileUrl}}
+        {{svg-jar 'job-name-icon' class='icon' width=14 height=14}} <span class='vertical-align'>{{filePath}}</span>
+      {{/external-link-to}}
+    {{else}}
+      {{svg-jar 'job-name-icon' class='icon' width=14 height=14}} <span class='vertical-align'>{{filePath}}</span>
```
artursmirnov: `filePath` seems to always be empty here, isn't it?

app/utils/format-config.js
```diff
@@ -0,0 +1,15 @@
+import formatSha from 'travis/utils/format-sha';
+
+const slugRe = (slug) => (new RegExp(`^${slug}/`));
+
+const fileName = (source) => (source.match(/([^\/]*)(@\w{16})?$/)[1]);
```
artursmirnov: 1. I would suggest to set `source` to empty string by default for error safety: `(source = '') =>`

2. I would suggest another approach:
```js
const fileName = (source = '') => (source.split('@').firstObject || '').split('/').lastObject;
```
This would allow to avoid complex regexp and hardcoded hash length.

app/utils/format-config.js
```diff
@@ -0,0 +1,15 @@
...
+
+export const fileNameWithoutSha = (source) => (fileName(source).replace(/(@\w{16})$/, ''));
```
artursmirnov: It seems like `.replace` is doing nothing here, because `fileName` function already cleans up the hash, isn't it?

Also please use function declarations instead of function expressions for exported functions:
```js
export function fileNameWithoutSha(source = '') {
  return fileName(source);
}
```
Function declarations have some advantages over lambdas, e.g. function name, which my simplify debugging.

artursmirnov: Maybe even just export `fileName` function calling it something like extractFileName? What do you think?

app/utils/format-config.js
weban: About 2nd suggestion - current solution is more secure (case: file without sha but with @ in name)

app/utils/format-config.js
weban: About 1st one nope because fileName returns name with @ and sha

app/utils/format-config.js
artursmirnov: @weban I see your point. Although, I don't like the hardcoded hash length. It may potentially break if API format changes. And in this case it will also produce JS error as the regexp will result in `null`, from which we will then try to get `[1]`. Let's try to find more reliable and flexible solution.

PR #2005, opened 2019-04-08
Add PythonDefault banner,  Python version switch from 2.7 to 3.6

MariadeAnton:


MariadeAnton:
As a reference, this is an example build using the default Python version: https://travis-ci.org/MariadeAnton/build-pushes/jobs/517294766

PR #2006, opened 2019-04-08
Disable liquid fire animations for testing environment

artursmirnov:
This change addresses the issue with Percy showing random diffs caused by modal dialog animation.


artursmirnov:
Thanks for the review, @sorta 👍

PR #2011, opened 2019-04-10
Ensure that nav menu styles do not get merged during minification

sorta:
After deploying https://github.com/travis-ci/travis-web/pull/1998 to staging, I noticed a difference in behavior of the navigation in IE/Edge between the deployed and locally run versions. I suspect that the CSS minifier / optimizer is merging together some nearly identical rules during the build for deployment. Testing out this hypothesis.

### DIFF
--- a/app/styles/app/modules/navigation.scss
+++ b/app/styles/app/modules/navigation.scss
@@ -245,6 +245,7 @@ ul #profile-menu-link ul:hover {
 }
 
 ul #profile-menu-link:focus-within > ul {
+  opacity: 1;
   display: block;
 }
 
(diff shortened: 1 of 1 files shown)


sorta:
I think this confirms my suspicions. Compare:

https://so-trick-minifier.test-deployments.travis-ci.org/
https://ieissue.test-deployments.travis-ci.org/
https://staging.travis-ci.org/

The style declaration could have been anything to differentiate it from the rules above, but I think `opacity: 1` makes the intent clear.

sorta:
Thanks for reviewing! 😄

PR #2012, opened 2019-04-12
Fix an issue where public builds always show on initial load.

sorta:
This PR is intended to address this issue: https://github.com/travis-ci/travis-insights/issues/75


sorta:
I'm thinking about whether there's a better way to structure the public/private switching and loading, but for now this should be a relatively simple and straightforward way to address this bug.

sorta:
Thanks for reviewing @artursmirnov ! :smile:

PR #2014, opened 2019-04-12
[Insights] improve overlay messaging

sorta:
A work in progress, intended to address this issue: https://github.com/travis-pro/team-teal/issues/2939

PR #2017, opened 2019-04-16
Update owner routes to use ember-data to retrieve owner model

sorta:
Split from https://github.com/travis-ci/travis-web/pull/2014.

Currently, the owner model on the owner routes is manually pulled in through an ajax call in the route, rather than through ember-data. Architecturally, this is not ideal and inconsistent with the rest of the app.


sorta:
This PR isn't particularly urgent so it can wait until the beta UI (https://github.com/travis-ci/travis-web/pull/1995/) stuff is merged. I can see that a little bit of this PR should be changed once that merge is done.

artursmirnov:
Ugh, sorry @sorta this PR totally got out of my head :( Please, don't hesitate to ping me in such cases!

app/models/owner.js
```diff
@@ -22,6 +22,10 @@ export default Model.extend({
   githubId: attr('string'),
   education: attr('boolean'),
   fullName: or('name', 'login'),
+  permissions: attr(),
+  type: attr('string'),
+  isUser: equal('type', 'user'),
+  isOrganization: equal('type', 'organization'),
```
artursmirnov: Oh, I like this type extraction! Probably it would also make sense to remove these properties from `user` and `organization` models.

sorta: ooh yeah that's a good idea

sorta:
No worries @artursmirnov ! I got distracted last week with some higher priority work and also the hiring stuff, but yeah it's good to know I can ping if needed :smile:

PR #2019, opened 2019-04-29
Org help form submit to new forum topic

sorta:
A quick prototype of another possible way to address this issue: https://travisci.assembla.com/spaces/Web/tickets/16/details

This solution involves changing the submit button into a link to pre-fill a new forum topic, for example like this: http://travis-ci.community/new-topic?title=topic%20title

There's an awkward issue that can happen when trying to do so if you aren't logged into the forums. Basically, the authorization popup can tend to be blocked, since it isn't triggered directly by a user action.

So for now, this is only a prototype and I haven't updated any of the tests etc.

### DIFF
--- a/app/components/zendesk-request-form.js
+++ b/app/components/zendesk-request-form.js
@@ -8,6 +8,7 @@ import moment from 'moment';
 import config from 'travis/config/environment';
 
 const { apiHost, createRequestEndpoint } = config.zendesk;
+const { community } = config.urls;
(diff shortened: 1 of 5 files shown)


sorta:
Closing this in favor of https://github.com/travis-ci/travis-web/pull/2020 . If we end up wanting to implement this approach in the future, we can either reopen or start a new branch & PR.

PR #2021, opened 2019-05-07
Fix issue where date change can cause percy to flag a visual change

sorta:
I noticed that percy is flagging the insights page again, based on dynamic content. Since the date range is the current 30 days, different dates in the date display component will have different widths, which causes the content on the right to move.


sorta:
Hah yep pretty simple. Thanks for the review!

PR #2022, opened 2019-05-08
Add allow config imports for private repositories

weban:
https://github.com/orgs/travis-pro/projects/31

### DIFF
--- a/app/controllers/settings.js
+++ b/app/controllers/settings.js
@@ -47,6 +47,12 @@ export default Controller.extend({
       || settings.hasOwnProperty('auto_cancel_pull_requests');
   }),
 
+  showAllowConfigImportsSwitch: computed('model.settings', 'repo.private', function () {
+    let settings = this.get('model.settings');
+    let isPrivate = this.get('repo.private');
+    return isPrivate && settings.hasOwnProperty('allow_config_imports');
+  }),
+
   migratedRepositorySettingsLink: computed('repo.slug', function () {
     let slug = this.get('repo.slug');
     return this.get('externalLinks').migratedToComSettingsLink(slug);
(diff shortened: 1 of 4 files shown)


app/controllers/settings.js
```diff
@@ -47,6 +47,12 @@ export default Controller.extend({
       || settings.hasOwnProperty('auto_cancel_pull_requests');
   }),
 
+  showAllowConfigImportsSwitch: computed('model.settings', 'repo.private', function () {
+    let settings = this.get('model.settings');
+    let isPrivate = this.get('repo.private');
+    return isPrivate && settings.hasOwnProperty('allow_config_imports');
```
artursmirnov: For future reference, I would suggest to avoid relying to API data shape for UI behavior. It's out of our control and it's not covered by any tests as we don't call real API in tests. This means, it may change someday and UI will just break silently. Also it would be really hard to track and debug such errors.

tests/acceptance/repo/settings-test.js
```diff
@@ -360,6 +360,31 @@ module('Acceptance | repo settings', function (hooks) {
...
+    await settingsPage.visit({ organization: 'org-login', repo: 'repository-name' });
+
+    assert.ok(settingsPage.configImportsSection.exists, 'expected auto-cancellation section to exist');
+    assert.ok(settingsPage.allowConfigImports.isActive, 'expected allow_config_imports pushes to be present and enabled');
+
+    const settingToRequestBody = {};
+
+    server.patch(`/repo/${this.repository.id}/setting/:setting`, function (schema, request) {
+      settingToRequestBody[request.params.setting] = JSON.parse(request.requestBody);
+    });
```
artursmirnov: @weban could you please clarify a bit the purpose of this `patch` hook? I can't find any assertions against `settingToRequestBody` and the hook doesn't return anything.

PR #2024, opened 2019-05-14
Crons form update

weban:
Work in progress


artursmirnov:
It takes too much time to build. This usually happens when the tests have errors in asynchronous operations. In such cases test runner waits for async operation to finish (which never happens because of error) or the timeout (10000ms per test iirc).

@weban I think the revert of `form-field` component and `save` task changes should solve it.

app/components/add-cron-job.js
```diff
@@ -1,45 +1,68 @@
...
+  options: ['Always run', 'Do not run if there has been a build in the last 24h'],
+
+  currentCronJobsBranches: mapBy('repository.cronJobs', 'branch.name'),
+
+  init() {
+    this.reset();
+    this.setProperties({
+      'branchesFetcher.repository': this.get('repository')
+    });
+    this.get('branchesFetcher.showSearchResults').perform();
```
artursmirnov: I don't think it's a good idea to pre-fetch branches. Currently it feel like there are just 10 branches (because they are pre-fetched). It would be better to show empty list with a placeholder, like "Start typing branch name to search", and then show search results in the list as they load.

app/components/add-cron-job.js
```diff
@@ -1,45 +1,68 @@
...
-      store.unloadRecord(existingCrons.get('firstObject'));
-    }
+    if (!branches) { return []; }
 
-    const cron = store.createRecord('cron', {
-      branch,
-      interval: this.get('selectedInterval') || 'monthly',
-      dont_run_if_recent_build_exists: this.get('selectedOption') || false
+    branches.forEach((branch) => {
+      branch.set('disabled', branchNames.indexOf(branch.name) > -1);
```
artursmirnov: Am I correct, that this `disabled` property is for `power-select` to disable options in the list that already have a cron? If so, then I have two issues:

1) Are you sure that there can't be multiple crons for the same branch? I don't see any reason why not. Why not build the same branch two times a day, let's say? Not sure about that, but sounds like a valid case.

2) I don't think it's a good way to do disabling. When you set a property on the Ember Model instance, it persists in the store and affect all other places where this model is used. For example, imagine that there is another dropdown with branches in other place. This will disable branches there too, because the property will reflect there as well. If (1) is not the case and the branches still need to be disabled, then the proper way to disable them would be to derive separate `branchOptions` list instead of mutating objects in this one. Something like:
```js
branchOptions: computed('currentCronJobsBranches.@each', 'branchesFetcher.searchResults.@each.name', function() {
  const { currentCronJobsBranches = [], branchesFetcher } = this;
  return branchesFetcher.searchResults.map(({ name }) => {
    const disabled = currentCronJobsBranches.includes(name);
    return { name, disabled };
  })
})
```
This will create a new list of objects with branch `name` and `disabled`; If any other branch properties are needed, they can be exposed in the same way.

You can find a pretty similar example here: https://github.com/travis-ci/travis-web/blob/ae27f51c001dfb17bd8719113c80a8dc1e693858/app/components/dialogs/migrate-beta.js

app/components/add-cron-job.js
```diff
@@ -1,45 +1,68 @@
...
-  intervals: ['monthly', 'weekly', 'daily'],
+      this.reset();
+      cron.save();
+    },
 
-  options: ['Always run', 'Do not run if there has been a build in the last 24h']
+    onQueryChange(query) {
+      this.set('branchesFetcher.searchQuery', query);
+      this.get('branchesFetcher.performSearchRequest').cancelAll();
+      this.get('branchesFetcher.showSearchResults').perform();
```
artursmirnov: As per previous comments, `query` should be passed as an argument to the task instance via `perform` and you can do it directly in template:
```
{{field.select oninput=(perform 'performSearchRequest') }}
```

app/components/add-cron-job.js
```diff
@@ -1,45 +1,68 @@
...
-  }).drop(),
+      const cron = store.createRecord('cron', {
+        branch: this.get('selectedBranch'),
+        interval: this.get('selectedInterval'),
+        dont_run_if_recent_build_exists: this.get('selectedOption')
+      });
 
-  intervals: ['monthly', 'weekly', 'daily'],
+      this.reset();
+      cron.save();
```
artursmirnov: Is there any reason to extract this logic from Ember Concurrency task? I would suggest to keep async operations inside tasks, to handle concurrency properly. And also perform it directly from template, no need in component action.

app/templates/components/add-cron-job.hbs
```diff
@@ -1,32 +1,70 @@
...
-    {{#if save.isRunning}}
-      {{loading-indicator}}
-    {{else}}
-      <input type="submit" value="Add" class="form-submit" />
-    {{/if}}
+  <div class='form-elem'>
+    <button
+      onclick={{form.submit}}
+      class='form-submit'>
+      {{#if save.isRunning}}
```
artursmirnov: You have removed the `save` task from component, because of which now it adds a cron row immediately, without waiting for server response, and does not show loading indication.

tests/pages/settings.js
```diff
@@ -119,6 +119,7 @@ export default create({
   }),
 
   cronBranches: collection('.form--cron form select:nth(0) option'),
+  conBranchesSelect: clickable('.form--cron form .select:nth(0) .travis-form__field-select'),
```
artursmirnov: It seems to be a typo

app/templates/components/add-cron-job.hbs
```diff
@@ -1,32 +1,70 @@
...
-    <span class="label">Options</span>
-    {{#x-select value=selectedOption required="false" on-change=(action (mut selectedOption)) as |xs|}}
-      {{#xs.option value=false}}{{options.[0]}}{{/xs.option}}
-      {{#xs.option value=true}}{{options.[1]}}{{/xs.option}}
-    {{/x-select}}
+  <div class='form-elem select'>
+    {{#form.field
+      label='Options'
+      value=selectedOption 
+      requiresValidation=false
```
artursmirnov: There's no need to set this property externally, it is a computed property that derives its value depending on the component settings. This may cause unexpected results, because the component is designed differently. Please avoid overriding computed properties of components as in most cases it means misuse of the component.

app/templates/components/add-cron-job.hbs
```diff
@@ -1,32 +1,70 @@
...
-  <div class="form-elem select">
-    <span class="label">Branch</span>
-    {{#x-select value=selectedBranch required="true" on-change=(action (mut selectedBranch)) as |xs|}}
-      {{#each branches as |branch|}}
-        {{#xs.option value=branch}}{{branch.name}}{{/xs.option}}
-      {{/each}}
-    {{/x-select}}
+{{#travis-form 
+  class='form--cron' 
+  onSubmit=(action 'handleSubmit') as |form| }}
```
artursmirnov: Just for consistency, please put `as` and closing brackets to new lines, for better readability:
```
{{#travis-form
  class='form--cron'
  onSubmit=(action 'handleSubmit') 
  as |form| 
}}
```

app/templates/components/add-cron-job.hbs
```diff
@@ -1,32 +1,70 @@
...
+      required=true
+      value=selectedBranch
+      onChange=(action (mut selectedBranch))
+      as |field|
+    }}
+      {{#field.select 
+        searchField='name' 
+        placeholder='Type branch name'
+        oninput=(action 'onQueryChange') 
+        options=comparedBranches as |branch|}}
```
artursmirnov: The same about `as` and closing brackets

app/templates/components/add-cron-job.hbs
weban: This is for hide error messges

artursmirnov: Ah, I see, that makes sense! Thanks for the clarification 👍

app/components/forms/form-field.js
```diff
@@ -62,10 +62,7 @@ export default Component.extend({
     let validator = this.validator;
 
     if (!validator) {
-      if (this.required)
-        validator = presense();
-      else
-        return;
+      validator = this.required ? presense() : () => true;
```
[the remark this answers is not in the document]
weban: If validator is not set and the field is not required its mean that field should be valid but now value is undefined and form is marked as invalid

app/components/add-cron-job.js
```diff
@@ -1,45 +1,61 @@
...
-    const cron = store.createRecord('cron', {
-      branch,
-      interval: this.get('selectedInterval') || 'monthly',
-      dont_run_if_recent_build_exists: this.get('selectedOption') || false
+  performSearchRequest: task(function* (query) {
+    yield timeout(config.intervals.searchDebounceRate);
+    this.set('loadingData', true);
+    this.setProperties({
+      branches: yield Branch.search(this.store, query, this.repository.id),
+      loadingData: false
```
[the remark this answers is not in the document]
weban: About `branch.set('disabled', branchNames.indexOf(branch.name) > -1);`
Like I said from UX perspective better solution is show disabled option than nothing.
About map to {name, disabled} object there is no sense because it will require additional method to find branch object by name

app/components/forms/form-field.js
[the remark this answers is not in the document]
artursmirnov: No, if validator is not set and field is not required, it mean that it does not need to be validated. Please notice, that field has three states:
- `FIELD_STATE.DEFAULT` - no validation made
- `FIELD_STATE.VALID` - validation successful
- `FIELD_STATE.ERROR` - validation failed

This change will make the field always to change state after validation - either to `FIELD_STATE.VALID` or `FIELD_STATE.ERROR`, whereas if it does not require validation, it must remain in `FIELD_STATE.DEFAULT` state. 

If you'd revert this change, most likely you won't need to set `requiresValidation` manually, it'll just work as expected and won't show any validation badges.

@weban does it make sense?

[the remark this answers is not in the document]
artursmirnov: Form or field validity does not depend on the return value of this method. It depends on the `state` value of the field, which this method sets.

app/components/add-cron-job.js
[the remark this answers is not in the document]
artursmirnov: It seems like your comment belongs to [this conversation](https://github.com/travis-ci/travis-web/pull/2024#discussion_r286672639). Let's keep the further conversation there.

app/templates/components/add-cron-job.hbs
```diff
@@ -1,32 +1,65 @@
...
-      {{loading-indicator}}
-    {{else}}
-      <input type="submit" value="Add" class="form-submit" />
-    {{/if}}
+  <div class='form-elem'>
+    <button
+      onclick={{form.submit}}
+      class='cron-form-submit form-submit'>
+      Add
+    </button>
```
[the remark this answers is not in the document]
weban: Documentation doesn't have an indicator :P

app/components/add-cron-job.js
artursmirnov: [Response by @weban](https://github.com/travis-ci/travis-web/pull/2024#discussion_r287738199):
> About branch.set('disabled', branchNames.indexOf(branch.name) > -1);
Like I said from UX perspective better solution is show disabled option than nothing.
About map to {name, disabled} object there is no sense because it will require additional method to find branch object by name

@weban I totally agree with your point that showing disabled options is more useful. Although, my initial comment has slightly different questions, please re-read it. 

Basically, my question was, do they really need to be disabled? Because I don't see any reasons for that. And if they _do_ for some reason, then it **must** be done in a different way, because current implementation is prone to bugs and very fragile (refer to the example in the initial comment).

app/templates/components/add-cron-job.hbs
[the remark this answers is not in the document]
artursmirnov: What do you mean?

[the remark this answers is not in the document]
weban: nvmd, afer consulting with our ux lead we decided to remove branches wit active crons from the list

app/components/forms/form-field.js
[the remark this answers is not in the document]
weban: requiresValidation isnt set manually

app/components/add-cron-job.js
```diff
@@ -1,45 +1,48 @@
...
-    const branch = this.get('selectedBranch') || this.get('branches.firstObject');
-
-    const existingCrons = yield store.filter('cron', { repository_id: repoId }, (c) => {
-      c.get('branch.repoId') === repoId && c.get('branch.name') === branch.get('name');
-    });
+  search: task(function* (query) {
+    yield timeout(config.intervals.searchDebounceRate);
+    let branchNames = this.get('currentCronJobsBranches');
+    let branches = yield Branch.search(this.store, query, this.repository.id);
+    return branches.reject(branch => (branchNames.indexOf(branch.name) > -1));
```
artursmirnov: Could you please use `branchNames.includes(branch.name)` instead of `indexOf`.

app/components/add-cron-job.js
```diff
@@ -1,45 +1,48 @@
...
-
-    const cron = store.createRecord('cron', {
-      branch,
-      interval: this.get('selectedInterval') || 'monthly',
-      dont_run_if_recent_build_exists: this.get('selectedOption') || false
+  save: task(function* () {
+    const cron = this.store.createRecord('cron', {
+      branch: this.selectedBranch,
+      interval: this.selectedInterval.toLowerCase(),
+      dont_run_if_recent_build_exists: this.selectedOption
```
artursmirnov: This is a boolean property, which will always be `true` in this case, because `selectedOption` is non-empty string.

app/components/add-cron-job.js
```diff
@@ -1,45 +1,48 @@
...
-
-    this.reset();
-
     yield cron.save();
-  }).drop(),
-
-  intervals: ['monthly', 'weekly', 'daily'],
-
-  options: ['Always run', 'Do not run if there has been a build in the last 24h']
+    this.reset();
```
artursmirnov: I would suggest to add some error handling to save request. Currently it will just hang in case of API or network error.

app/components/forms/form-switch.js
```diff
@@ -0,0 +1,35 @@
...
+  onChange() {},
+
+  onToggle: abstractMethod('onToggle'),
+
+  didInsertElement() {
+    this._super(...arguments);
+    this.onInit(this.elementId);
+  },
+
+  'aria-checked': computed('checked', function () {
```
artursmirnov: It would be better to give the property a valid Javascript name, like `ariaChecked`, so that it could be re-used, and just bind it in `attributeBindings`:
```js
attributeBindings: ['ariaChecked:aria-checked', 'role', 'type'],
```

app/components/travis-form.js
```diff
@@ -1,10 +1,10 @@
 import Component from '@ember/component';
-import { computed } from '@ember/object';
 import { gt, filterBy, not } from '@ember/object/computed';
 import * as validators from 'travis/utils/form-validators';
 
 export default Component.extend({
   tagName: '',
+  className: 'travis--travis-form',
```
artursmirnov: We usually strive to tie component CSS class to component name where possible. So would it be possible to rename the `className` to just `travis-form`. Also, it's not a strict rule, but we sometimes use the [BEM naming convention](http://getbem.com/naming/) for CSS styles, so this name will conflict with it (`--travis-form` does not look like a valid modifier).

And, as far as I understand, there's a typo: `className` should be [`classNames`](https://api.emberjs.com/ember/3.10/classes/Component/properties/classNames?anchor=classNames)

app/models/branch.js
```diff
@@ -20,3 +20,18 @@ export default Model.extend({
...
+
+Branch.reopenClass({
+  search(store, name, repositoryId) {
+    return store.query('branch', {
+      repository_id: repositoryId,
+      data: {
+        name: name,
+        sort_by: 'repository.name',
+        limit: 10
+      }
```
artursmirnov: 1. There's no need to nest properties under `data` - serializer will handle that.  
2. `sort_by` appear to not be working properly (especially given `repository.name`, why sort branches by repository name?) - it just doesn't affect the results, they are not sorted anyways, so it can be removed. 
3. Limit does not make sense here too, API has default limit and we should fetch as much as it can give us (why limit users in dropdown?). 
4. Although, what we do need here is a filter by `exists_on_github`, because otherwise create cron request will return an error.

As a result, payload should be like this:
```js
    return store.query('branch', {
      repository_id: repositoryId,
      name,
      exists_on_github: true
    });
```
As you may notice, it's already not a general "branch search", it's "branch search for cron jobs". That's why I suggested to move the request to the respective component - no need to expose specific request as a global method.

app/templates/components/add-cron-job.hbs
```diff
@@ -1,32 +1,64 @@
...
+    {{#form.field
+      label='Branch'
+      value=selectedBranch
+      required=true
+      class='branch-selector travis-form__field--without-badge'
+      as |field|
+    }}
+      {{#field.select
+        searchField='name'
+        placeholder='Type branch name'
```
artursmirnov: I would like suggest a bit different labels: https://github.com/travis-ci/travis-web/blob/2adf33c31687303b6770724f7d1e65a86e3c59ba/app/templates/components/add-cron-job.hbs#L7-L21

app/templates/components/forms/form-field.hbs
```diff
@@ -66,6 +66,13 @@
     onInit=(action 'setFieldElementId')
     class='travis-form__field-component'
   )
+  switch=(component 'forms/form-switch'
+    form=form
+    checked=value
+    disabled=disabled
+    onChange=(action 'handleChange')
+    class='travis-form__field-component'
+  )
```
artursmirnov: Thanks for adding this! 😄

tests/integration/components/add-env-var-test.js
```diff
@@ -44,32 +42,28 @@ module('Integration | Component | add env-var', function (hooks) {
...
-    assert.expect(3);
+    assert.expect(2);
 
     await render(hbs`{{add-env-var repo=repo}}`);
 
     assert.dom('.env-name').doesNotHaveAttribute('value', 'precond: name input should be empty');
 
-    await click('.form-submit');
-
-    assert.dom('.form-error-message').exists('the error message should be displayed');
```
artursmirnov: Why should this assertion be removed? It looks like a still valid use case.

app/components/add-cron-job.js
weban: Nope, IE9 and EDGE 12 is not supported

app/components/forms/form-switch.js
weban: 1 - 1 from travis switch

app/models/branch.js
weban: If sort is not working correctly, I'll fis it on api side. Limit make sense because we want only 10 branches per result

app/components/add-cron-job.js
weban: And about 4 - this is the way how it is handled now

app/components/travis-form.js
weban: Do we have any doc about our code / style rules?

app/components/add-cron-job.js
weban: Lets do it in next interation

app/components/add-cron-job.js
artursmirnov: `includes` will be polifilled by Ember: https://api.emberjs.com/ember/3.10/classes/EmberArray/methods/includes?anchor=includes

About 4 - It doesn't mean that currently it's the best solution. We should always try to improve our UX. And in my opinion, it fees confusing currently, when you know that the branch exists, but you can't find it in the list. For example, the most popular case would be with `master` branch. Please try to go [here](https://crons.test-deployments.travis-ci.org/travis-ci/travis-web/settings) and add a cron job for `master` and you'll see what I mean. And then try to do the same [here](https://as-crons-improvements.test-deployments.travis-ci.org/travis-ci/travis-web/settings) - doesn't it feel less confusing?
My concern is that this confusion will cause support issues.

app/components/add-cron-job.js
artursmirnov: I don't see any reason to push it to next iteration, it's just few lines of code, which you can find here: https://github.com/travis-ci/travis-web/blob/2adf33c31687303b6770724f7d1e65a86e3c59ba/app/components/add-cron-job.js#L56-L67

app/components/travis-form.js
artursmirnov: No, unfortunately, we don't have such document currently. That's why I mentioned that it's not a strict rule, but like a general recommendation, because it may cause confusion with other developers.

app/models/branch.js
artursmirnov: 1. Yes, I'm sure, serializer does that job.
3. I didn't know about the limit. It just doesn't feel right from the UX perspective. Should we rise the conversation about it in the Slack channel?

app/components/add-cron-job.js
artursmirnov: @weban please change to `includes` and let me know what you think about (4)

app/models/branch.js
weban: 1. doesnt work witthout the wrap
2. I think that we should decide about it, 10 in my opinion is ok

app/components/add-cron-job.js
```diff
@@ -1,45 +1,76 @@
...
+    yield timeout(config.intervals.searchDebounceRate);
+    let branchNames = this.get('currentCronJobsBranches');
+    let branches = yield this.store.query('branch', {
+      repository_id: this.repository.id,
+      data: {
+        name: query,
+        sort_by: 'name',
+        limit: 10,
+        exists_on_github: true
+      }
```
artursmirnov: As I mentioned before, it's not necessary to wrap params with `data`.

artursmirnov: Please take a look here: https://github.com/travis-ci/travis-web/blob/2adf33c31687303b6770724f7d1e65a86e3c59ba/app/components/add-cron-job.js#L47-L51

And it's deployed here: https://as-crons-improvements.test-deployments.travis-ci.org/travis-ci/travis-web/settings

So you can see live how it works.

app/components/add-cron-job.js
artursmirnov: @weban please comment on this

app/components/add-cron-job.js
```diff
@@ -1,45 +1,76 @@
...
+    return store.query('branch', {
+      repository_id: repositoryId,
+      data: {
+        name: name,
+        sort_by: 'repository.name',
+        limit: 10,
+        exists_on_github: existsOnGithub
+      }
+    });
+  },
```
artursmirnov: This looks like a code duplication. Could you please either remove it here or use it in component?

app/templates/components/add-cron-job.hbs
artursmirnov: @weban could you please comment on this?

app/templates/components/add-env-var.hbs
```diff
@@ -1,21 +1,48 @@
...
+}}
+  <div class='form-elem'>
+    {{#form.field
+      label='Name'
+      value=name
+      required=true
+      showValidationStatusIcons=false
+      as |field|
+    }}
+      {{field.input class='env-name' placeholder='Name' onChange=(action (mut name))}}
```
artursmirnov: `onChange` should be bound to `form.field`, not to `field.input`, in order to handle all states and validations properly.

app/templates/components/add-env-var.hbs
```diff
@@ -1,21 +1,48 @@
...
-    {{input value=value class="env-value" placeholder="Value"}}
+  <div class='form-elem'>
+   {{#form.field
+      label='Value'
+      value=value
+      required=true
+      showValidationStatusIcons=false
+      as |field|
+    }}
+      {{field.input  class='env-value' placeholder='Value' onChange=(action (mut value))}}
```
artursmirnov: The same about `onChange`

app/templates/components/add-env-var.hbs
```diff
@@ -1,21 +1,48 @@
...
-  <div class="form-elem">
-    {{travis-switch active=public description="Display value in build log" onToggle=(action (mut public))}}
+  <div class='form-elem'>
+    {{#form.field
+      label='Display value in build log'
+      class='travis-form__field--without-frame travis-form__field--switch'
+      value=public
+      as |field|
+    }}
+      {{field.switch onChange=(action (mut public))}}
```
artursmirnov: The same about `onChange`

app/templates/components/add-env-var.hbs
```diff
@@ -1,21 +1,48 @@
...
+    }}
+      {{field.input  class='env-value' placeholder='Value' onChange=(action (mut value))}}
+    {{/form.field}}
   </div>
-  <div class="form-elem">
-    {{travis-switch active=public description="Display value in build log" onToggle=(action (mut public))}}
+  <div class='form-elem'>
+    {{#form.field
+      label='Display value in build log'
+      class='travis-form__field--without-frame travis-form__field--switch'
```
artursmirnov: 1) It would be better to set the `travis-form__field--without-frame` class by setting `disableFrame=true` property (they are bound inside component), to keep it encapsulated.
2) The `travis-form__field--switch` class should be set as a root class in component definition, because all instances of this component must have this class.

app/templates/components/forms/form-switch.hbs
```diff
@@ -0,0 +1,11 @@
+  <div class="switch-inner">
+    <span class="on">
+      {{svg-jar 'icon-passed' class='icon-on' alt='on'}}
+    </span>
+    <span class="off">
+      {{svg-jar 'icon-failed' class='icon-off' alt='off'}}
+    </span>
+  </div>
```
artursmirnov: It seems like the indentation level is incorrect here, should be 1 level less.

app/models/branch.js
artursmirnov: 1. Please refer to this comment: https://github.com/travis-ci/travis-web/pull/2024#discussion_r288986604
2. I agree, 10 sounds good to me. Anyways, we can easily increase it any time later in case of customers' feedback.

PR #2036, opened 2019-05-31
Help page milestone 4

sorta:
For https://github.com/travis-ci/product/issues/200


sorta:
I ended up moving some of the show/hide logic from the help page into the zendesk form component. I think we'll want this same behavior if we put the component elsewhere. Plus the component already had logic to show the login prompt.

sorta:
Hi @artursmirnov :smile: I hope all is well! It's not urgent, but when you have time could you review this PR? Thanks!

artursmirnov:
Ugh, sorry @sorta, seems like I missed the review request notification 😕  Will review shortly, thanks for the ping!

app/components/zendesk-request-form.js
```diff
@@ -40,7 +42,23 @@ export default Component.extend({
...
+  subscriptions: reads('accounts.subscriptions'),
+  activeSubscriptions: filterBy('subscriptions', 'isSubscribed', true),
+  isSubscribed: notEmpty('activeSubscriptions'),
+  isEducation: reads('auth.currentUser.education'),
+
+  trial: reads('auth.currentUser.trial'),
+  trialBuildsRemaining: reads('trial.buildsRemaining'),
+  noTrialYet: empty('trial'),
+
+  isPremium: or('isSubscribed', 'isEducation', 'trialBuildsRemaining', 'noTrialYet'),
```
artursmirnov: I really like how you juggle with computed macros! 😄

sorta: Thanks! :smile:

sorta:
Hi @artursmirnov, thanks for the review! I've addressed #1 and #3.

With #2 however, I wasn't sure if you were referring to just the grey background, or the image/text as well. Based on the screenshot I though it was about the background, but it looks to me like the grey background does currently reach full width (unless the issue is happening in a specific browser/situation?)

BUT if you're referring to the section content, that should be easy to address.

Either way, let me know. Thanks!

artursmirnov:
Hey @sorta, yes, I was referring to the background. And you're right, it looks correct in Chrome, the issue is in Firefox. I'm sorry for the confusion. Please let me know if you need any more details to reproduce the issue.

sorta:
Cool, thanks @artursmirnov! And no worries, that's one of the challenges of frontend. So many variables and it's not always clear which ones are relevant.

Still can't seem to reproduce this unfortunately. So far I've tried firefox 67 on Windows, OSX, and Ubuntu, logged in and out. But, I've also put in a small update to explicitly set the width for `.layout-striped-section` to 100%. Hopefully that may address it, and shouldn't introduce any issues. Please let me know if that fixes it, or if you think of any circumstances that may help reproduce this issue.

artursmirnov:
Hey @sorta, that fix doesn't seem to help, unfortunately. Here's the screenshot with more detailed information about the browser (which also shows the issue):

image

If it still doesn't help you to reproduce the issue, please let me know, maybe I will just fix it on my laptop.

sorta:
Excellent, thanks for the additional info! I'll try a few more things to see if I can reproduce

sorta:
Ok yeah I'm not sure what's going on here. I did my best to recreate those conditions and still can't seem to reproduce the issue. I also took another look through the changes in this PR to see if there were any changes that might be causing this, but nothing so far has stood out to me. So yeah, when you get a chance, could you investigate a bit and let me know if you figure out what's going? I would appreciate it very much 😄 

If it helps at all, here's a similar screenshot that I took on my macbook:
Screen Shot 2019-06-13 at 11 21 02 AM

artursmirnov:
Hey @sorta, looks like I figured that out. It's caused by "Settings -> Dashboard" turned on, which makes the `.main` container fixed width. I think this issues is outside of the PR scope, so probably it would make sense to just create an issues and fix it as a next step, what do you think?

sorta:
Ahhhh, ok that makes sense! Thanks for looking into it @artursmirnov! I don't know that I would have ever tried that hah.

Yeah I think that sounds good to me. Thanks for the review!

PR #2041, opened 2019-06-05
Improve flaky polling test

sorta:
I've seen this test occasionally fail for no reason. I think this small update could possibly address that.


sorta:
Awesome, thanks for the review @artursmirnov ! :smile: I've actually had this issue happen during a build semi-recently, so yeah hopefully this will address it

PR #2043, opened 2019-06-07
update config validation messages

svenfuchs:
some details of these have changed, but also, we want to always include the full message key in order to make the messages more useful/comprehensible


app/components/build-message.js
```diff
@@ -9,129 +9,130 @@ const { escapeExpression: escape } = Ember.Handlebars.Utils;
 export default Component.extend({
   readableMessage: computed('message.code', 'message.args', function () {
```
sorta: Since we are now depending on `message.key`, we should probably add it here as well, like this:
```readableMessage: computed('message.code', 'message.key', 'message.args', function () {```

This way if message.key changes for some reason, the computed property will know to recompute on next access. Here's more info about Ember Computed Properties if it's helpful :smile: 
https://guides.emberjs.com/release/object-model/computed-properties/

File: app/components/build-message.js [hunk 2]
```diff
@@ -9,129 +9,130 @@ const { escapeExpression: escape } = Ember.Handlebars.Utils;
...
     let code = this.get('message.code');
+    let key  = this.get('message.key');
     let args = this.get('message.args');
```
sorta: This is totally optional, but I think we can reduce these three lines into one like this:

`const { code, key, args } = this.message;`

There's a couple of things going on here:
- I think we can use `const` rather than `let`, since we won't need to reassign these variables a new value
- In current versions of ember we can now generally use `this.xyz` rather than needing to use `this.get('xyz')`, just a slightly shorter way to do the same thing
- We can use destructuring to assign multiple local variables (code, key, args) from the same parent object

But yeah, this is a small optional thing, either way is fine.

app/components/build-message.js
svenfuchs: done in https://github.com/travis-ci/travis-web/commit/537c39e16409e256068999a3fe8361547cb31d7f

app/components/build-message.js
svenfuchs: done in https://github.com/travis-ci/travis-web/commit/537c39e16409e256068999a3fe8361547cb31d7f

thank you for the great review, @sorta ❤️

app/components/build-message.js
sorta: Awesome, it looks good :) thanks!

app/components/build-message.js
sorta: Awesome, it looks good to me :smile: :heart: Thanks!

PR #2044, opened 2019-06-07
pretty print raw config json

svenfuchs:
Changes the config view to pretty print JSON payloads for requests triggered via API

image

### DIFF
--- a/app/components/raw-config.js
+++ b/app/components/raw-config.js
@@ -14,6 +14,14 @@ export default Component.extend({
     return this.get('copied') ? 'Copied!' : `Copy ${fileNameWithoutSha(source)}`;
   }),
 
+  formattedConfig: computed('rawConfig.config', 'slug', function() {
+    let config = this.get('rawConfig.config');
+    if(config && config.charAt(0) == '{') {
+      config = JSON.stringify(JSON.parse(config), null, 2);
+    }
+    return config;
+  }),
+
   filePath: computed('rawConfig.source', 'slug', function () {
     let source = this.get('rawConfig.source');
     let name = fileNameWithoutSha(source);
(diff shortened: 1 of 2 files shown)


app/components/raw-config.js
```diff
@@ -14,6 +14,14 @@ export default Component.extend({
     return this.get('copied') ? 'Copied!' : `Copy ${fileNameWithoutSha(source)}`;
   }),
 
+  formattedConfig: computed('rawConfig.config', 'slug', function () {
+    let config = this.get('rawConfig.config');
+    if (config && config.charAt(0) == '{') {
+      config = JSON.stringify(JSON.parse(config), null, 2);
+    }
+    return config;
```
artursmirnov: I would suggest to use a bit different approach here, to make the code more robust. Currently, it only checks if the config starts from `{` and assumes that it's in JSON format. Although, it doesn't guarantee the validity of the config, even if it's JSON. If `JSON.parse` will fail, the entire method will brake. Thus, my recommendation would be to wrap the JSON manipulation to `try`-`catch` block. As a bonus, in this case there's no need for condition at all:
```js
const { config } = this.rawConfig || {};
try {
  return JSON.stringify(JSON.parse(config), null, 2);
} catch() {
  return config;
}
```
In this case if any of `stringify` or `parse` will fail (either because of incorrect structure or it's not JSON at all), it will just return the string representation anyways.

How does it sound to you?

svenfuchs: @artursmirnov totally makes sense. i've updated it.

except, i think `const { config } = this.rawConfig || {};` wouldn't work as expected, right? the config blob that we want to parse sits on `rawConfig.config`, so i have kept that line in place.

artursmirnov: @svenfuchs `const { config } = this.rawConfig || {};` is an [Object destructuring](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#Object_destructuring), it does basically the same, but looks a bit cleaner. Although, both ways work perfectly!

PR #2047, opened 2019-06-14
Display DRAFT badge for draft PRs

pavel-d:
travis-ci/product#181


app/models/request.js
```diff
@@ -46,6 +47,11 @@ export default Model.extend({
     return eventType === 'pull_request';
   }),
 
+  isDraft: computed('pullRequestMergeable', function () {
+    let mergeableState = this.get('pullRequestMergeable');
+    return mergeableState === 'draft';
+  }),
```
artursmirnov: `pullRequestMergeable` looks like a dictionary field, thus I would suggest to define it's possible values as a hash, so that it was clear what they are and easier to re-use:
```js
export const PULL_REQUEST_MERGEABLE = {
  DRAFT: 'draft',
  ...
}
```

In this case, the `isDraft` field would be just a [computed macro](https://api.emberjs.com/ember/3.10/functions/@ember%2Fobject%2Fcomputed/equal): 
```js
isDraft: equal('pullRequestMergeable', PULL_REQUEST_MERGEABLE.DRAFT)
```

app/models/request.js
```diff
@@ -17,6 +17,7 @@ export default Model.extend({
   headCommit: attr(),
   baseCommit: attr(),
   branchName: attr(),
+  pullRequestMergeable: attr(),
```
artursmirnov: Am I right that this field is `string`? If so, please specify the type for `attr`:
```js
pullRequestMergeable: attr('string'),
```
This helps to ensure the field type and derialize/extract data properly. More info about it here: https://guides.emberjs.com/release/models/defining-models/#toc_transforms

tests/integration/components/build-header-test.js
```diff
@@ -89,6 +89,31 @@ module('Integration | Component | build header', function (hooks) {
...
+      request: {
+        isDraft: true
+      }
+    };
+
+    this.set('build', build);
+    await render(hbs`{{build-header item=build commit=build.commit}}`);
+
+    assert.dom('.build-title .message-label').hasText('draft');
+  });
```
artursmirnov: Thanks for adding the integration tests! Although, could you please also add a check for the badge in acceptance tests? We mostly rely on them as they ensure the final user experience.

tests/acceptance/job/basic-layout-test.js
```diff
@@ -64,6 +64,60 @@ test('visiting job-view', function (assert) {
...
+
+    assert.equal(jobPage.createdBy.href, '/testuser');
+    assert.equal(jobPage.createdBy.text, 'Mr T');
+    assert.ok(jobPage.createdBy.avatarSrc.startsWith('/images/favicon-gray.png'));
+
+    assert.equal(jobPage.log, 'Hello log');
+    assert.notOk(jobPage.hasTruncatedLog);
+    assert.equal(jobPage.rawLogUrl, `${config.apiEndpoint}/v3/job/${job.id}/log.txt`);
+  });
+});
```
artursmirnov: I would suggest to just add an assertion for `draft` badge to existing test instead of creating a new one just for this small part - similar to the previous acceptance test in this PR. What do you think?

tests/pages/build-list.js
```diff
@@ -17,6 +17,7 @@ export default create({
 
   builds: collection('.build-list .pr-row', {
     name: text('.build-info a'),
+    badge: text('.badge'),
```
artursmirnov: We are aiming to stick with [test selectors](https://ember-test-selectors.com/) and avoid using CSS classes for testing purposes. Just add an attribute to the element, something like `data-test-draft-badge` and then select it in test with attribute selector: `[data-test-draft-badge]`. This helps to make tests independent from styling. Also the add-on removes all test selectors from production builds.

tests/acceptance/job/basic-layout-test.js
pavel-d: @artursmirnov, I've added this one, because there is no test for pull request build, only for branch build

tests/pages/build-list.js
pavel-d: Cool, didn't know that

PR #2048, opened 2019-06-18
Only use the package-json.lock for npm deps installation

joshk:
Based on https://docs.npmjs.com/cli/install and https://stackoverflow.com/questions/52499617/what-is-the-difference-between-npm-install-and-npm-ci

> This command is similar to npm-install, except it’s meant to be used in automated environments such as test platforms, continuous integration, and deployment – or any situation where you want to make sure you’re doing a clean install of your dependencies.


joshk:
Thanks for the merge @artursmirnov

PR #2050, opened 2019-06-18
Show wall time on build history

clekstro:
I think the current behavior on the Build History tab -- showing total time spent across all of a build's jobs when viewing a finished build item -- is a bit confusing. Instead, I think we should show the wall time a build took. We currently provide no context as to how many jobs a build has, but include them as part of the time calculation. If you have a cron scheduled with multiple jobs, this can lead to a stark time difference which gives the impression that more investigation is required.

We currently have both time concepts in the app, which we show when viewing a single build in detail (see "Ran for" vs. "Total time"). This would only impact builds with multiple jobs, as jobs with a single job have no difference between the wall/total time.

### DIFF
--- a/app/mixins/duration-calculations.js
+++ b/app/mixins/duration-calculations.js
@@ -3,6 +3,7 @@ import durationFrom from 'travis/utils/duration-from';
 import { computed } from '@ember/object';
 
 export default Mixin.create({
+  useDurationDirectly: true,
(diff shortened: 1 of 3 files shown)


clekstro:
In terms of the UI effects of this change, it's really just a change to the number we're displaying. But here's a before/after to illustrate (compare cron build # 17306).

Currently in production:
Screen Shot 2019-06-19 at 9 02 58 AM

Proposed change:
Screen Shot 2019-06-19 at 9 03 40 AM

PR #2053, opened 2019-06-19
Reduce liquid fire duration in tests

clekstro:
This is recommended in the project [README](https://github.com/ember-animation/liquid-fire#testing) and can't help but speed up the test suite.

### DIFF
--- a/app/templates/components/log-content.hbs
+++ b/app/templates/components/log-content.hbs
@@ -85,7 +85,7 @@
     targetAttachment="center"
     translucentOverlay=true
     overlayPosition='sibling'
-    animatable=(config-get 'liquidFire.enabled')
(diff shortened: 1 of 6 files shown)


clekstro:
@artursmirnov Yeah, I just realized that `animatable` will be false when in tests, so this doesn't change behavior. Would you be ok with removing the `config-get` calls within templates in favor of this change? I think it's nicer to just to say the modal is always animatable and pass `true`, while letting the duration config take care of the performance concerns.

artursmirnov:
@clekstro yeah, sure, that'd be awesome!

clekstro:
Approved the percy diff after viewing it in the PR environment, looks good. 🚢

PR #2054, opened 2019-06-19
Convert more tests to modern testing style

clekstro:


clekstro:
This does skip a few tests that were giving trouble (mainly `sync` tests that were leverage `setTimeout`), but overall should get us cut over to the new testing APIs.

tests/acceptance/feature-flags/app-boots-test.js
```diff
@@ -1,59 +1,64 @@
...
 
-  andThen(function () {
     assert.equal(currentURL(), '/');
   });
-});
 
-test('app does not request feature flags on boot if available in local storage', function (assert) {
-  assert.expect(1);
+  test('app does not request feature flags on boot if available in local storage', async function (assert) {
+    assert.expect(1);
```
artursmirnov: Not sure, but it seems like this test should expect 2 assertions, shouldn't it?

clekstro: I think this test is probably just a bit too clever. Instead of the normal `assert.ok(true)`, we assert the opposite (`assert.ok(false)`) so that the test would fail if the request were ever made. This isolates the failure from someone accidentally deleting the expectation that a certain amount of assertions are made, but it's definitely not the most intuitive thing to read...

artursmirnov: Ah, that makes sense! I didn't get the idea, it just catched my eye so I thought to mention. Thanks for the clarification!

PR #2058, opened 2019-06-21
[Plans page] add tailwind

sorta:


sorta:
So, Tailwind's preflight/base styles cause visual changes (generally spacing) on just about every page. I tried to configure the `ember-cli-tailwind` to leave out the preflight styles, but I was having a hard time getting it to actually do that. I then tried to see if the addition of a couple of simple styles in `base.scss` would resolve a lot of the issues, but it did not. So I was wondering if this addon is going to work for us.

Hence, why I am trying an alternate approach in https://github.com/travis-ci/travis-web/pull/2061, and it's described [here](https://github.com/chrism/emberjs-tailwind-purgecss). Additionally, with that setup, it would be easier to add PurgeCSS if we wanted, which would help reduce the bundle size.

sorta:
Closing in favor of https://github.com/travis-ci/travis-web/pull/2061

PR #2061, opened 2019-06-25
[Plans page] install tailwind

sorta:
[This attempt](https://github.com/travis-ci/travis-web/pull/2058) to add tailwind resulted in some issues. This is an alternate approach (based on [this](https://github.com/chrism/emberjs-tailwind-purgecss)) that hopefully may address or reduce those issues.

Additionally, with this setup, it would be easier to add PurgeCSS if we wanted, which would help reduce the bundle size.


sorta:
So I started to run into trouble when I got to this part of the guide https://github.com/chrism/emberjs-tailwind-purgecss#updating-css-importing. I guess we could skip it, but I think we will eventually want to write our own tailwind components and utilities. Plus, as this branch is right now, I've noticed that there is an EPIPE error that pops up during serve, which is probably not ideal. I'm starting to think it may be better after all to have it in a completely different repo.

sorta:
Never mind! Yesterday I THOUGHT that I had tried to just use SASS's `@import` and that it wasn't working well for some reason. But I tried again briefly this morning and it seems to work fine. Maybe there was some weird/subtle difference with how I had things arranged yesterday, but things seem to be working fine now.

sorta:
Ok I think this is in a good place to merge into `epic-plans-page` and start using it to do actual updates to the plans page

sorta:
Hi @artursmirnov :smile: When you have some time, could you review this? Thanks!

artursmirnov:
Sure, @sorta, will take a look shortly 👍

ember-cli-build.js
```diff
@@ -75,6 +75,18 @@ module.exports = function () {
...
+      compile: {
+        enabled: false
+      },
+      filter: {
+        enabled: true,
+        exclude: ['vendor/**/*'],
+        plugins: [
+          require('tailwindcss')('./app/styles/app/tailwind/config.js')
+        ]
+      }
```
artursmirnov: This is something different compared to [documentation](https://github.com/chrism/emberjs-tailwind-purgecss#including-in-the-build-pipeline), is it on purpose? I didn't dive deep into it, just noticed a difference :)

ember-cli-build.js
```diff
@@ -75,6 +75,18 @@ module.exports = function () {
...
+    },
+    postcssOptions: {
+      compile: {
+        enabled: false
+      },
+      filter: {
+        enabled: true,
+        exclude: ['vendor/**/*'],
+        plugins: [
+          require('tailwindcss')('./app/styles/app/tailwind/config.js')
```
artursmirnov: Would it make sense to put Tailwind to `./app/styles/tailwind`, to separate it from our old styles? I also thought about `./config/tailwind.js`, but this will require to restart ember serve after each change. What do you think?

ember-cli-build.js
sorta: Yeah, I tried that way first, and first ran into an issue where postcss was expecting `/app/styles/app.css` rather than `/app/styles/app.scss`. So I tried to follow this [guide](https://github.com/jeffjewiss/ember-cli-postcss#switching-from-sass) for working with both sass and postcss in ember, but then I ran into [this issue](https://github.com/jeffjewiss/ember-cli-postcss/issues/143) a bunch and decided to try this other `filter` based approach.

HOWEVER, I just realized that the point of that method given in that guide is to _replace_ your sass pipeline with like a postcss-based one. So I tried that method along with _removing_ `ember-cli-sass`, and that seems to allow that method to work. So it seems to be a viable option, and might be the way to go.

ember-cli-build.js
sorta: I like the idea of separating it more from our old styles. :+1: 

And yeah, it feels weird to have the config file outside the `config` folder, and especially in the `styles` folder. But I also don't like the idea of needing to restart ember serve after a change to it. It could go by itself in some other folder, like `/app/tailwind/config.js`, which I've seen elsewhere, but that feels weird too. Or maybe `/app/tailwind.config.js`? I'm open to different approaches on this

ember-cli-build.js
artursmirnov: Oh, that's a great research! Thanks for the explanation :)

ember-cli-build.js
artursmirnov: I think it should belong to `/app/styles` as it basically defines styles of the app. All our current styles live in `/app/styles/app/**`, so I though, it would be good to have Tailwind styles in `/app/styles/tailwind/**`. Although, this is just my theoretical suggestion and you may have some sensible objections.

ember-cli-build.js
sorta: :smile:

ember-cli-build.js
sorta: Nah, no objections here. That makes sense to me :smile:

gokatz:
Thanks, @sorta for detailed explanation and PR.
We ran into a few similar issues what you have described in the comments. Able to use tailwind with the help of this PR.

PR #2062, opened 2019-06-26
Slice commit messages

rafael-travis:
_Fixes https://github.com/travis-ci/bugs-questions-comments-ideas/issues/89_

## The problem

The problem is with RegExp matching a long string without spaces, it takes ~13 seconds on my macbook to process a String with ~2000 chars.

Build__5_-_rafael-travis_test_-_Travis_CI

The browser's CPU usage also increases a lot while processing that.

See this JSPerf with an example:

https://jsperf.com/long-string-match/1

## Solution

The commit message just shows about 50 characters on those 2 pages, my solution is to limit the string to that amount of characters, the UI looks the same and the RegExp matching is faster.

### DIFF
--- a/app/templates/components/builds-item.hbs
+++ b/app/templates/components/builds-item.hbs
@@ -27,7 +27,7 @@
       </div>
     {{else}}
       <div class="row-message">
-        {{{format-message build.commit.message short="true" repo=build.repo eventType=build.eventType}}}
+        {{{format-message build.commit.message short="true" repo=build.repo eventType=build.eventType maxLength=50}}}
       </div>
     {{/if}}
   </div>
--- a/app/helpers/format-message.js
+++ b/app/helpers/format-message.js
@@ -14,6 +14,11 @@ emojiConvertor.allow_native = false;
 
 function formatMessage(message, options) {
   message = message || '';
+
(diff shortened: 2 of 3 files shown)


rafael-travis:
Thanks for review it, @sorta !

app/templates/components/builds-item.hbs
```diff
@@ -27,7 +27,7 @@
       </div>
     {{else}}
       <div class="row-message">
-        {{{format-message build.commit.message short="true" repo=build.repo eventType=build.eventType}}}
+        {{{format-message build.commit.message short="true" repo=build.repo eventType=build.eventType maxLength=50}}}
```
cotsog: Just curious, why is the `maxLength` value different in the 2 files? i.e. one is 60 and the other is 50. Thanks!

rafael-travis: Because the containers' widths are different, thus amount of text that appears is different too.

cotsog: Ah! Makes sense. Thanks for explaining @rafael-travis!

rafael-travis: No problem, any time @cotsog!

PR #2065, opened 2019-06-28
Handle 403 responses on caches page

rafael-travis:
Complements https://github.com/travis-ci/travis-api/pull/929

### DIFF
--- a/app/templates/repo/error.hbs
+++ b/app/templates/repo/error.hbs
@@ -1,4 +1,4 @@
-{{#if isForbidden}}
+{{#if (eq model.status 403)}}
   Your permissions are insufficient to access this content.
 {{else}}
   There was an error while loading data.
--- a/app/templates/repo/error.hbs
+++ b/app/templates/repo/error.hbs
@@ -1,1 +1,5 @@
-There was an error while loading data.
+{{#if isForbidden}}
+  Your permissions are insufficient to access this content.
+{{else}}
+  There was an error while loading data.
+{{/if}}
--- a/app/adapters/beta-migration-request.js
+++ b/app/adapters/beta-migration-request.js
@@ -1,0 +1,4 @@
+import V3Adapter from 'travis/adapters/v3';
+
+export default V3Adapter.extend({
+});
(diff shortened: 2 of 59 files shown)


rafael-travis:
@artursmirnov thanks a lot for the documentation reference. Is this how you thought about implementing this?

artursmirnov:
Hey @rafael-travis, yes, almost there :) Although, I think that `error` object should be available in the error state context and thus in `error` template (but I'm not sure about this). If it's available, then you can use it directly to display different messages:
```hbs
{{#if (eq error.status 403)}}
  Forbidden error
{{else}}
  Other error
{{/if}}
```
In this case you won't need to define a controller and even catch an `error` event, it should just work.

If `error` object is not available in the `error` substate template directly, I would suggest to add a `repo/error` route and set error object manually, like this: 
```js
setupController(controller, error) {
  this._super(...arguments);
  controller.set('error' error);
}
```
And then use it in template as above.

The issue with your current solution is that controllers in Ember are singletons, so when you set `isForbidden` to it, it will persist there forever, unless you reset it manually.

rafael-travis:
@artursmirnov thanks, I couldn't find a way to do that #if on my researches, that's much better, let me know if this looks good to you.

app/templates/repo/error.hbs
```diff
@@ -1 +1,5 @@
-There was an error while loading data.
+{{#if (eq model.status 403)}}
```
rafael-travis: @artursmirnov it's accessible through this `model` variable.

artursmirnov:
Hey @rafael-travis, thanks for working on this! It looks great to me 👍 I really like such a straight forward solution. Please merge the PR as soon as you feel it's ready for release to production.

rafael-travis:
@artursmirnov thank you for educating me on Ember, I appreciate it 👍

rafael-travis:
@artursmirnov `percy/travis-web` failed but I don't have access to see what happened, do I need to be added to the team in there?

artursmirnov:
@rafael-travis, no worries, fixed ;)

rafael-travis:
@artursmirnov thanks!

PR #2069, opened 2019-06-28
Address style typo

sorta:


sorta:
@artursmirnov I present to you, my tiniest PR ever :laughing: Just noticed this and wanted to take care of it. Please review when you have a free moment. Thanks!

app/styles/app/base.scss
```diff
@@ -172,7 +172,7 @@ span.emoji.emoji-sizer {
   display: inline-block;
   border-bottom: 1px solid transparent;
   color: $asphalt-grey;
-  margin: auto 1px
+  margin: auto 1px;
```
artursmirnov: Ugh.. I'm wandering why the build didn't fail 🤔

sorta: Yeah I wondered the same thing. Weird.

sorta: It does cause the build to fail on the new setup I have going for the plans page update, which is how I noticed it.

Maybe that line was being interpreted as SASS rather than SCSS or something?

PR #2074, opened 2019-07-03
Make footer more consistent

sorta:
I noticed, while working on https://github.com/travis-ci/travis-web/pull/2070 (which this is split from), that the footer layout is inconsistent. Especially on "center" layout pages. So I came up with this update that should hopefully help with that.

### DIFF
--- a/app/components/travis-layout.js
+++ b/app/components/travis-layout.js
@@ -6,7 +6,7 @@ export default Component.extend({
 
   layoutName: '',
 
-  layoutClass: computed(function () {
+  layoutClass: computed('layoutName', function () {
     return this.layoutName.replace('layouts/', 'layout--');
   })
 });
--- a/app/components/travis-layout.js
+++ b/app/components/travis-layout.js
@@ -7,6 +7,6 @@ export default Component.extend({
   layoutName: '',
 
   layoutClass: computed(function () {
-    return this.layoutName.replace('/', '-');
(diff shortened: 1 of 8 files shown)


sorta:
Hi @artursmirnov, can you review this small update when you get a chance? Thanks :smile:

app/components/travis-layout.js
```diff
@@ -1,3 +1,12 @@
 import Component from '@ember/component';
+import { computed } from '@ember/object';
 
-export default Component.extend({});
+export default Component.extend({
+  classNameBindings: ['layoutClass'],
+
+  layoutName: '',
+
+  layoutClass: computed(function () {
```
artursmirnov: Should this computed property depend on `layoutName`? Currently it will only calculate on first request and never update afterwards. Is it expected?

artursmirnov:
And I really like the Percy diffs 😄

app/components/travis-layout.js
sorta: You know, I had a weird feeling that I was forgetting something haha. Yes, I think it should depend on `layoutName`. I've made that change. Thanks for catching it! :smile:

artursmirnov:
Wow, that was fast 🚀 Looks perfect, thanks @sorta!

sorta:
Yeah I happened to be checking my email, saw the notification, and it was an easy change :rocket: Thanks for the review!

PR #2075, opened 2019-07-03
As update deps

artursmirnov:
Updating Ember version and dependencies


artursmirnov:
This PR updated Ember version to the latest 3.11 along with all it's dependencies and all add-ons. Also I run a few codemods that remove deprecations. 

I don't think this PR is reviewable as it has almost 200 files changed, and most of the changes are made by codemods. So it's mostly for information and to make everyone aware.

artursmirnov:
@emmanuel-ci this is a good question, thanks! No, you won't need to run these codemods, as this PR does not introduce new syntax. It only update dependencies and resolves deprecations (`sendAction`, some jQuery APIs, `computed` changes, etc.), but all of them are resolvable through code reviews. Although, I'd still recommend to [get familiar with deprecations](https://deprecations.emberjs.com/v3.x/) and use modern and recommended syntax (e.g. avoid using `.get` wherever possible).

app/components/add-ssh-key.js
```diff
@@ -56,16 +58,16 @@ export default Component.extend({
...
-        value: this.get('value')
+        description: this.description,
+        value: this.value
       });
 
       try {
         yield sshKey.save();
         this.reset();
-        return this.sendAction('sshKeyAdded', sshKey);
+        return this.sshKeyAdded(sshKey);
```
rafael-travis: Out of curiosity, is this doing anything or is just a noop?

app/components/created-by.js
```diff
@@ -11,8 +11,8 @@ export default Component.extend({
   }),
 
   showUser: computed('nameOrLogin', 'eventType', function () {
-    let nameOrLogin = this.get('nameOrLogin');
-    let eventType = this.get('eventType');
+    let nameOrLogin = this.nameOrLogin;
+    let eventType = this.eventType;
```
rafael-travis: Another curiosity, I saw that we use `let` instead of `const` throughout the codebase, is there any reason for that?

sorta: I don't think we have any firm policy on use of `let` or `const`, and a lot of people have worked on `travis-web` over the years. I think in many cases it was someone's preference at the time, but there are probably also a few cases where there was a reason for it. But there's no single overall reason for it that I'm aware of.

app/components/add-ssh-key.js
artursmirnov: The `sshKeyAdded` is a method that can be assigned from the outside as a property, when the component is used, e.g.:
```hbs
{{add-ssh-key sshKeyAdded=(action 'onSshKeyAdded')}}
```

artursmirnov: The no-op is just a stub so that the action won't fail if the property is not assigned. This makes components more flexible. And also explicitly defines public interface of a component.

PR #2077, opened 2019-07-04
As moss fix feedback

artursmirnov:
Some improvements for the issues reported by support [here](https://travisci.slack.com/archives/C6CMPT8R0/p1562245599021900)


artursmirnov:
@sorta thanks for review! I have checked the test and the message is correct, as in the test migration is not enabled for user account.

sorta:
Ahh gotcha. Sounds good then!

PR #2078, opened 2019-07-05
Log view improved, so now it's possible to show 100k lines.

piotr-travisci:


tests/acceptance/builds/restart-test.js
```diff
@@ -27,6 +27,6 @@ test('restarting build', function (assert) {
 
   andThen(() => {
     assert.equal(topPage.flashMessage.text, 'The build was successfully restarted.', 'restarted notification should display proper build restarted text');
-    assert.equal(buildPage.singleJobLogText, 'Hello log', 'shows log text of single build job');
+    assert.equal(buildPage.singleJobLogText, '1Hello log', 'shows log text of single build job');
```
artursmirnov: I'm not sure about this change. Seems like something is wrong.

tests/acceptance/job/basic-layout-test.js
```diff
@@ -58,7 +58,7 @@ test('visiting job-view', function (assert) {
     assert.equal(jobPage.createdBy.text, 'Mr T');
     assert.ok(jobPage.createdBy.avatarSrc.startsWith('/images/favicon-gray.png'));
 
-    assert.equal(jobPage.log, 'Hello log');
+    assert.equal(jobPage.log, '1Hello log');
```
artursmirnov: Same here, do you know what have caused this change? It may also affect our customers, which may be unexpected.

tests/acceptance/job/basic-layout-test.js
```diff
@@ -112,7 +112,7 @@ test('visiting pull request job-view', function (assert) {
...
```
artursmirnov: Same here.

tests/acceptance/builds/restart-test.js
piotr-travisci: It's correct. In previous solution we had numbering in css, however it has a limitation of numbering lines when they are hidden (display:none), so numbering was moved into js. That's why log line always contains line number, even in tests.

app/utils/log.js
```diff
@@ -1064,7 +1066,8 @@ Log.extend(Log.Renderer.prototype, {
     // `createSpan` function and I'm not into biggger refactoring at the moment
     para = document.createElement('div');
     para.classList.add('log-line');
-    para.appendChild(document.createElement('a'));
+    let a = document.createElement('a');
+    para.appendChild(a);
```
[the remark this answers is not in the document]
piotr-travisci: Thanks

app/styles/app/layouts/log.scss
```diff
@@ -262,6 +256,16 @@
 
     &.open {
       height: auto;
+      .log-line {
```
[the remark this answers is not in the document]
piotr-travisci: Thanks. You are right.

app/styles/app/layouts/log.scss
```diff
@@ -262,6 +256,16 @@
...
+        display: block;
+      }
+    }
+
+    .log-line {
+      display: none;
+      &:first-of-type {
```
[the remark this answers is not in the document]
piotr-travisci: :+1:

tests/acceptance/builds/restart-test.js
artursmirnov: Ah, I see, that makes sense. Then I would suggest to make it more explicit and adapt test to actually assert that a line has both line number and text. Maybe even add an assertion to validate that numbering is correct (e.g. that second line has `2`). What do you think? Currently it just looks a bit weird and unclear.

piotr-travisci: You are right. Now tests are extended for multiline log.

PR #2092, opened 2019-07-22
Update CI to use Xenial and some additional cleanup

clekstro:


clekstro:
@artursmirnov I'd be down to try it. They're still warning that it's not completely stable yet though, so maybe we wait a bit longer?

artursmirnov:
@clekstro yeah, I totally agree, it doesn't make sense to use unstable tools 👍 Though, do you think it worth keeping an eye on the tool?

PR #2097, opened 2019-07-26
UI improvements for removing selected branch in env var selector

piotr-travisci:


artursmirnov:
Hey @piotr-travisci could you please add a bit of context here? It's not clear why this change is required and why existing `allowClear` behavior does not work here? To be honest, I would be in favor of more general solutions to keep our UIs more consistent. E.g if for some reason `allowClear` has poor usability, then it makes sense to improve it everywhere (make `x` more prominent, etc.). What do you think?

piotr-travisci:
Thank you @artursmirnov for review.
Look&feel of allowClear is now improved.

piotr-travisci:
Thank you @artursmirnov for this finding.
I fixed branch selector for long branch names and now it looks like this
Screenshot 2019-07-31 at 13 24 20

piotr-travisci:
Thank you for your constructive review @artursmirnov

PR #2102, opened 2019-08-02
Fix for text selection of log lines, which are in folds.

piotr-travisci:


artursmirnov:
Ok, this PR doesn't work for some reason. I just created a new branch, pushed it and created another PR (https://github.com/travis-ci/travis-web/pull/2103) and it's green. Let's go with that. @piotr-travisci could you please review it?

PR #2103, opened 2019-08-02
Fix line folding

artursmirnov:
Trying to fix https://github.com/travis-ci/travis-web/pull/2102

### DIFF
--- a/app/styles/app/layouts/log.scss
+++ b/app/styles/app/layouts/log.scss
@@ -245,6 +245,7 @@
       text-decoration: none;
       position: absolute;
       padding-right: 1em;
+      user-select: none;
     }
 
   }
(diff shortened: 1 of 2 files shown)


app/styles/app/layouts/log.scss
```diff
@@ -246,13 +246,13 @@
       position: absolute;
       padding-right: 1em;
     }
```
piotr-travisci:
```suggestion
      -webkit-user-select: none;
      -moz-user-select: none;
      -ms-user-select: none;
      user-select: none;
    }
```

artursmirnov: @piotr-travisci this is a great suggestion! Though, prefixes are added automatically by [autoprefixer](https://github.com/travis-ci/travis-web/blob/master/package.json#L34) and thus not necessary.

piotr-travisci: 👍 @artursmirnov.
Suggestion refactored.

artursmirnov: Thanks!

PR #2113, opened 2019-08-14
Fix slow log autoscroll + fix not working autoscroll/top in FireFox

piotr-travisci:


piotr-travisci:
Thank you for review @artursmirnov.
It wasn't working in FireFox before at all, so jerky is better than none, but you are right. We have to refactor/improve it.

PR #2115, opened 2019-08-20
[Plans] OSS section

sorta:


sorta:
I added spacing definitions to the tailwind config. The ones I added are the ones that tailwind has by default, but now they're explicitly defined in the config.

I also realized that the grid-item helper function was missing the sizePrefix for one of the possible return values, so I added that in and that's what that is about.

I'm also wondering whether, at this point, it might make sense to define an `Svg` and/or `Image` component for the UI kit?

sorta:
Anyway, we did a lot of the heavy lifting in the previous PR and this one was a bit more straightforward, so I think it's ready for an initial review when you get a chance for it @artursmirnov. Thanks!

app/templates/plans.hbs
```diff
@@ -123,33 +123,30 @@
...
-      <p class="text-big">
-        Our on-premises product is perfect for companies who want to keep using the same features of Travis CI with additional on-site security needs.
-      </p>
-      <a href="https://enterprise.travis-ci.com" title="learn more about Travis CI" class="plans-button--blue">Learn more about Travis CI Enterprise</a>
-    </div>
-    <div>
-      <SvgImage @name='enterprise-servers'/>
-    </div>
-  </section>
+         <UiKit::Button @onClick={{action 'gaCta' 'plans-oss'}} @color='grey' @invert={{true}}>
```
artursmirnov: Seems like there's an extra space here :)

artursmirnov:
With regards to the `Image` component - I don't have any strong preferences here, so whatever works best for you ;) It definitely makes sense to add one at some point, but if you're on a rush and/or don't need it right now, it's totally okay to keep using `SvgImage`.

app/templates/plans.hbs
sorta: Hah, nice catch. Removed :+1:

sorta:
Alright, that header spacing should be a bit more consistent now. I added a `leading` prop to the `Text` component to adjust line-height for cases like this.

And yeah, maybe for now it makes sense to continue using `SvgImage`, so I think I'll go with that.

PR #2116, opened 2019-08-21
[Plans] Enterprise area

sorta:

Skipping the Contact section for now because that one is a little more complex, and there has already been progress made on other sections (including this one).


sorta:
Another straightforward section, a lot like the OSS section. This should be ready for an initial review whenever you have time for it @artursmirnov. Thanks! :smile:

sorta:
Thanks for the review! #1 & #3 were fairly easy to address, they each needed some minor additions to fix.

#2 requires us to think about how we want to set responsive properties. What sort of interface we would want for that. One possibility is to use a convention like `@textAlign`, `@smTextAlign`, `@mdTextAlign`, etc. I came up with an update to the `prefix` util that could support that without requiring us to write all those different states out for every responsive prop we want. On the other hand, doing it like that means that the public interface is a little less explicit, so maybe it would be better to write out all the states?

Or, maybe rather than having more properties, we can allow more complexity in the existing props? so something like `@textAlign={{hash base='center' sm='left'}}`. And maybe we can do it in a way where `@textAlign='center'` is still valid so that we don't have to be writing `hash base` everywhere. This could result in some complex values possible for things like `@margin`, but I'm starting to feel like maybe it's the way to go.

In either case, I just remembered we'll need to update things like `checkDictionary` to support whatever approach we go with.

Or maybe this is something we can settle later haha

Let me know what you think when you get a chance @artursmirnov :smile:

artursmirnov:
@sorta, I didn't think that it would add so much complexity :) But you're right, there should be a general approach to this. 

Roughly, from the top of my head, I think, it'd be good to have an ability to override literally any property for different breakpoints. It could be a convention that any property accepts both plain value (e.g. `@textAlign='left'`) or a hash with breakpoints (e.g. `@textAlign={{hash base='left' sm='center'}}`. This would be very handy and flexible. To achieve that we could use some kind of helper function or custom computed macro that would always expand plain values to hashes and normalize them (fill empty values with `base` or `default`). How does it sound?

Although, I think it's too much for now, we can think about it later and just go with a simpler solution for now, or even leave it as it is.

sorta:
Hah yeah, I've found in building this system that it can sometimes be difficult to know what kinds of things will end up adding a ton of complexity.

I like that approach a lot, and I started coming to a similar conclusion this morning. I started prototyping something along those lines, but yeah I think it's probably a bit much for this PR. So I reverted those things and for now I'm just gonna leave those sections left aligned. This breakpoint expansion approach has me thinking about possible ways the Grid components can be refactored using this approach.

PR #2120, opened 2019-08-25
Includes Assembla as a provider

rafael-travis:


rafael-travis:
Hey hey @sorta @piotr-travisci @artursmirnov.

This PR adds Assembla as provider and also applies Piotr and Artur's suggetions to the code.

Notice the base branch is `rg-vcs-links` and not `master` just to facilitate code review since you've already reviewed the previous one.

Please let me know your thoughts, as soon as I have your approval I plan to merge this work into `master` directly, does that sound good?

Thanks for all your help with this!

app/services/external-links.js
```diff
@@ -1,67 +1,28 @@
...
-    github: `/${slug}/blob/${branch}/${file}`,
-    bitbucket: `/${slug}/src/${branch}/${file}`,
-  }),
+  branchUrl: (vcsType, { owner, repo, branch }) => vcsUrl('branch', vcsType, { owner, repo, branch }),
+
+  fileUrl: (vcsType, { owner, repo, branch, file }) => vcsUrl('file', vcsType, { owner, repo, branch, file }),
+
+  repoUrl: (vcsType, { owner, repo }) => vcsUrl('repo', vcsType, { owner, repo }),
+
+  tagUrl: (vcsType, { owner, repo, tag }) => vcsUrl('tag', vcsType, { owner, repo, tag }),
```
artursmirnov: Shouldn't these methods belong to `vcsLinks` dictionary as well? They look pretty similar to the ones in `vcsLinks`.

app/utils/vcs.js
```diff
@@ -0,0 +1,59 @@
...
+const { providers } = config;
+
+const vcsId = (vcsType) => {
+  const availableProvidersRegex = Object.keys(providers).join('|');
+  const match = (vcsType || 'github').match(new RegExp(availableProvidersRegex, 'i'));
+
+  if (!match[0]) {
+    throw new Error(`Invalid VCS Type "${vcsType}"`);
+  }
+  return match[0].toLowerCase();
```
artursmirnov: Is it necessary to run the RegExp search in `vcsType`? It looks to me like it should just be a provider key, shouldn't it? Wouldn't 
```js
const vcsId = (vcsType = 'github') => {
  vcsType = vcsType.toLowerCase();
  assert(`Invalid VCS Type "${vcsType}"`, Object.keys(providers).includes(vcsType));
  return vcsType;
}
```
work here?

app/utils/vcs.js
```diff
@@ -0,0 +1,59 @@
...
+    .match(/:[a-z]+/g)
+    .map((param) => param.slice(1))
+);
+
+const paramsValid = (template, params) => (
+  arrayEqual(Object.keys(params), templateParams(template))
+);
+
+const arrayEqual = (array1, array2) => (
+  array1.sort().toString() == array2.sort().toString()
```
artursmirnov: I'd suggest to use strict equality here (`===`)

app/utils/vcs.js
```diff
@@ -0,0 +1,59 @@
...
+);
+
+export const vcsUrl = (resource, vcsType, params) => {
```
artursmirnov: I'd suggest to set a default value for `params = {}`, because it's public and if it won't be set from the outside, many methods here (that use `Object.keys`) will break.

app/utils/vcs.js
```diff
@@ -0,0 +1,59 @@
...
+  const vcs = vcsConfig(vcsType);
+  const url = vcs.endpoint + vcs.paths[resource];
+
+  if (!paramsValid(url, params)) {
+    throw new Error(`Missing url params. URL: ${url}, PARAMS: ${JSON.stringify(params)}`);
```
artursmirnov: You can use [`assert`](https://api.emberjs.com/ember/3.12/functions/@ember%2Fdebug/assert) method in such cases. Ember CLI removes them from production builds.

config/providers.js
```diff
@@ -0,0 +1,62 @@
...
+      profile: '/:owner',
+      repo: '/:owner/:repo',
+      tag: '/:owner/:repo/releases/tag/:tag',
+    },
+    vocabulary: {
+      organization: 'Organization',
+      pullRequest: 'Pull Request',
+    },
+  },
+};
```
artursmirnov: Config looks great to me 👍

app/services/external-links.js
artursmirnov: Sounds like a great idea, thank you for clarification 👍

app/utils/vcs.js
artursmirnov: Would it makes sense to make a dictionary of possible values and map them to `vcsId`s explicitly? Or even put to config as, let's say, aliases. My concern here is that current approach could potentially cause unpredictable and hard-to-track bugs, like if someone eventually would add some other `vcsType` without mentioning a provider - in this case UI will just silently brake. What do you think?

artursmirnov: It looks great to me, thanks @rafael-travis 👍

PR #2124, opened 2019-08-29
[Plans] faq content update

sorta:

This is in response to feedback about the content of the FAQ section

### DIFF
--- a/app/templates/components/ui-kit/text.hbs
+++ b/app/templates/components/ui-kit/text.hbs
@@ -1,6 +1,4 @@
-{{#let (element this.tag) as |Tag|}}
-  <Tag
-    class={{this.allClasses}}
-    ...attributes
-  >{{yield}}</Tag>
-{{/let}}
+{{#let (element this.tag) as |Tag|}}<Tag
+  class={{this.allClasses}}
+  ...attributes
+>{{yield}}</Tag>{{/let}}
--- a/app/controllers/plans.js
+++ b/app/controllers/plans.js
@@ -7,10 +7,15 @@ import { inject as service } from '@ember/service';
 
 const { plans } = config;
 
+const utmSource = '?utm_source=plans_page';
+
 export default Controller.extend({
   config,
 
   auth: service(),
+
+  billingUrl: `${config.billingEndpoint}/${utmSource}`,
+  buildMatrixUrl: `${config.urls.buildMatrix}${utmSource}`,
 
   plans: computed(() => plans),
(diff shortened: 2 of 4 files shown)


sorta:
Small updates mainly to FAQ section content. Should be ready for review when you get a chance @artursmirnov. Thanks!

app/templates/components/ui-kit/text.hbs
```diff
@@ -1,6 +1,4 @@
-{{#let (element this.tag) as |Tag|}}
-  <Tag
-    class={{this.allClasses}}
-    ...attributes
-  >{{yield}}</Tag>
-{{/let}}
\ No newline at end of file
+{{#let (element this.tag) as |Tag|}}<Tag
```
artursmirnov: There should be a way to handle these whitespace issues. Not now, but eventually. Something like setting container `font-size: 0` and then specify it for `UiKit::Text` nodes directly.

sorta: Yeah totally. I mean, I guess we could put a base style of `font-size: 0` on `body`. Although then absolutely any text that we want to show would need to be in a `UiKit::Text` component. But that's kind of the plan anyway, but then maybe there would be some weird other edge cases or unintended consequences... But yeah, we can definitely think of and work on a way to deal with whitespace.

sorta:
Awesome, thanks for the review! :smile:

app/templates/components/ui-kit/text.hbs
artursmirnov: I'd set some sensible font size for body to keep things like base font size and `em` work properly and then set it to `0` on some root `div` for the app. Though, I'm not really sure about this solution, makes sense to google about it :) There should be a battle tested solution :)

sorta: Yeah honestly I'm not really sure about that solution as well haha. It's definitely worth researching before doing anything drastic

sorta: Huh, TIL about Handlebars `~`. This seems to work:
```
{{#let (element this.tag) as |Tag|~}}
  <Tag
  class={{this.allClasses}}
  ...attributes
  >
    {{~yield~}}
  </Tag>
{{~/let}}
```

sorta: Docs about it: https://handlebarsjs.com/expressions.html#whitespace-control

artursmirnov: Wow, TIL 😍 Thanks for the research! I knew it would be simple 😄

PR #2135, opened 2019-09-18
Login with Assembla and Bitbucket

rafael-travis:
#### `localStorage.enableAssemblaLogin = false`

Screen Shot 2019-09-17 at 8 31 00 PM

#### `localStorage.enableAssemblaLogin = true`

Screen Shot 2019-09-17 at 8 28 24 PM

#### `localStorage.enableAssemblaLogin = false`

Screen Shot 2019-09-17 at 8 29 29 PM 

#### `localStorage.enableAssemblaLogin = true`

Screen Shot 2019-09-17 at 8 29 46 PM


app/services/auth.js
```diff
@@ -55,20 +55,29 @@ export default Service.extend({
     this.store.unloadAll();
   },
 
-  signIn(data, options = {}) {
+  signInWith(provider, data, options = {}) {
+    if (['github', 'assembla'].includes(provider)) {
```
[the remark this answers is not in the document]
rafael-travis: Yep, will do, thanks!

Nat-ess:
Hi @rafael-travis, this is looking so nice! I just have one tiny detail, could we make the drop-shadow at the bottom of the dropdown, the same size as the drop-shadows on the sides?

Log-in

artursmirnov:
Hey @rafael-travis, thanks for your work on this! I have some feedback on visual part first:

1) Hover behavior is a bit weird, please see the recording:
image

2) When I click on the green button, nothing happens, which looks broken. I understand, it's by design, but probably it would make sense to change it's cursor to default, so that it at least doesn't look like clickable. What do you think?

3) I'd suggest to add a shadow on hover to green button as well - then it would look more consistent. Though, it's @Nat-ess's call :)

4) Is it expected that on test deployments it shows only GitHub currently?

5) Should it be released right away or wait for some kind of announcement? I think it doesn't make sense until we actually support other providers.

app/components/header-links.js
```diff
@@ -59,10 +59,12 @@ export default Component.extend({
     return classes.join(' ');
   }),
 
+  enableAssemblaLogin: localStorage['enableAssemblaLogin'] == 'true',
```
artursmirnov: I assume this is only for testing and development? In this case, this should not go to production. Otherwise, probably it'd be better to use `config/environment` or feature flags for this purpose.

app/components/landing-default-page.js
```diff
@@ -4,9 +4,11 @@ import { inject as service } from '@ember/service';
 export default Component.extend({
   auth: service(),
 
+  enableAssemblaLogin: localStorage['enableAssemblaLogin'] == 'true',
```
artursmirnov: Same here

app/components/landing-pro-page.js
```diff
@@ -6,16 +6,14 @@ import { inject as service } from '@ember/service';
 export default Component.extend({
   auth: service(),
 
+  enableAssemblaLogin: localStorage['enableAssemblaLogin'] == 'true',
```
artursmirnov: Same here

app/services/auth.js
```diff
@@ -55,20 +56,29 @@ export default Service.extend({
...
-        url = new URLPolyfill(uri);
+      const uri = options.redirectUri || window.location.href;
+      let url = new URLPolyfill(uri);
 
       if (url.pathname === '/plans') {
         url.pathname = '/';
       }
 
-      window.location = `${this.endpoint}/auth/handshake?redirect_uri=${url}`;
+      const path = provider == 'github' ? '/auth/handshake' : `/auth/handshake/${provider}`;
```
artursmirnov: Please use strict equality wherever possible:
```suggestion
      const path = provider === 'github' ? '/auth/handshake' : `/auth/handshake/${provider}`;
```

app/services/auth.js
```diff
@@ -55,20 +56,29 @@ export default Service.extend({
     this.store.unloadAll();
   },
 
-  signIn(data, options = {}) {
+  signInWith(provider, data, options = {}) {
+    if (availableProviders.includes(provider)) {
+      this.signIn(data, options, provider);
+    } else {
+      throw new Error(`Invalid provider to authenticate ${provider}`);
+    }
```
artursmirnov: You can use [`assert`](https://api.emberjs.com/ember/3.12/functions/@ember%2Fdebug/assert) here instead:
```js
assert(`Invalid provider to authenticate ${provider}`, availableProviders.includes(provider));
this.signIn(data, options, provider);
```

app/services/multi-vcs.js
```diff
@@ -0,0 +1,22 @@
+import Service, { inject as service } from '@ember/service';
+import { computed } from '@ember/object';
+
+export default Service.extend({
+  features: service(),
+
+  enabled: computed('features.{enableAssemblaLogin,enableBitbucketLogin}', function () {
+    return this.features.get('enableAssemblaLogin') || this.features.get('enableBitbucketLogin');
```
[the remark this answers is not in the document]
rafael-travis: Yep, I get the following error:

Travis_CI

piotr-travisci:
Thank you @rafael-travis for fixing menu in RWD. Could you also center buttons?
Screenshot 2019-11-15 at 13 43 52

rafael-travis:
@piotr-travisci done, thanks!

Travis_CI

PR #2139, opened 2019-09-19
add css class icon-alert and alert.svg for config messages

svenfuchs:
Build config validation messages have a new level `alert`. This will be used when a plain string is given on a node that represents a secret and a `secure` value should be given:

image

Before:

image

After:

image

### DIFF
--- a/app/styles/app/modules/build-messages.scss
+++ b/app/styles/app/modules/build-messages.scss
@@ -45,7 +45,7 @@
     }
 
     code {
-      background-color: #f4f6fa;
+      background-color: #f1f1f1;
       font-family: $font-family-monospace;
       font-size: $font-size-sm - 1px;
       padding: 2px;
--- a/app/styles/app/modules/build-messages.scss
+++ b/app/styles/app/modules/build-messages.scss
@@ -45,7 +45,7 @@
     }
 
     code {
-      background-color: #eef0f4;
+      background-color: #f4f6fa;
       font-family: $font-family-monospace;
       font-size: $font-size-sm - 1px;
       padding: 2px;
--- a/app/styles/app/modules/build-messages.scss
+++ b/app/styles/app/modules/build-messages.scss
@@ -34,6 +34,10 @@
       @include colorSVG($brick-red);
     }
 
+    .icon-alert {
+      @include colorSVG($brick-red);
+    }
+
     .message {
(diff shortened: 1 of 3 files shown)


svenfuchs:
@sorta totally makes sense! i had not even noticed those. the tooltip is now there.

also https://github.com/travis-ci/travis-api/pull/967 will fix the ordering (going to merge that one as soon as the build is 🍏 )

svenfuchs:
i have dialed down the background color for `code` in those messages just a little bit, so it's more readable (at least for my eyes)

and i have removed some "Dropping" wording from the messages we render, as this is now parametrized in travis-yml and we do not actually drop anything right now, but might in the future (or even make this a preference).

svenfuchs:
Looking much better now:

image

app/styles/app/modules/build-messages.scss
```diff
@@ -34,14 +34,18 @@
...
+
     .message {
       padding-left: 7px;
       border-left: 1px solid $dry-cement;
       margin-left: 2px;
     }
 
     code {
-      background-color: #eef0f4;
+      background-color: #f4f6fa;
```
sorta: I personally find this a little harder to read hah (or maybe more distracting?), but maybe that's just me. We've gotten some complaints in the past about low-contrast stuff, so it might be worth running this by the Creative team. What do you think?

I also checked and noticed that we use `#f1f1f1` for some `code` blocks on the repo settings page (eg https://travis-ci.com/travis-ci/travis-web/settings), so that could be another option worth considering.

svenfuchs: hmmm, maybe. i am guessing you have much better eyes than i do :)

but i'd argue that "low contrast" is not "low contrast", it depends a lot of what being contrasted and why. in this case there's way too many tiny code blocks, so parsing the entire line really hard imo. and that's a case where lower contrast is much better, we still have the different font type as well.

using `#f1f1f1` seems like the path of least conflict then, so i'd use that?

but i can also just revert this if you don't like pushing this without the Creative team being in the loop.

sorta: Yeah totally, those are good points. I'm actually good with either color, I just wanted to pause a bit, think about it, and get some more of your thoughts on it. :)

PR #2143, opened 2019-09-23
Add line numbers and line links to config view

sorta:
For https://travisci.assembla.com/spaces/Web/tickets/213/details


sorta:
Some examples I've been working with:

https://so-config-add-line-nums.test-deployments.travis-ci.com/travis-ci/perl-builder/builds/128712741/config
https://so-config-add-line-nums.test-deployments.travis-ci.com/travis-ci/kubernetes-config/jobs/238181811/config

sorta:
Hi @piotr-travisci & @emmanuel-ci :smile: 

Thanks for the review! I wanted to add some notes here about the things you've brought up:

1. Linking the whole line - I agree, I think that's the way to go, and it's also something that the Product team has requested since this initially went into review
2. Line skipping - as far as I can tell, this is intentional. It seems to be the default behavior of the addons we're using. It allows the text to wrap while having the line numbers reflect what was in the original file.
3. Line heights are off - This might be a trickier issue to resolve. The line numbers plugin tries to set the heights of the numbers, but it doesn't always work well, especially if you zoom in or out. I've tried a few things and haven't come up with a solution yet. I have a few other things I'd like to try, but it has me considering whether we want to keep using prism for this. This is the only place where it's used. But I think it does do well with the syntax highlighting, so any replacement solution would need to include similar functionality.

Either way, for now I'll be focusing more on the higher priority plans page & analytics updates.

sorta:
I've updated the linking to make the whole line clickable rather than just the icon at the start. And I think I've also found a way to have the line numbers line up consistently. So I think this is ready for more testing and another round of code review whenever you have a chance @MariadeAnton @piotr-travisci @emmanuel-ci. Thanks!

piotr-travisci:
Looks & works perfect 🎉 
Thank you @sorta 👏

sorta:
Awesome, thanks for the review @piotr-travisci & @emmanuel-ci! :tada:

PR #2166, opened 2019-10-03
add a repo setting for build config validation

svenfuchs:
This UI will not be visible until we enable the respective feature flag in Api https://github.com/travis-ci/travis-api/pull/972 https://github.com/travis-ci/travis-api/pull/972/files#diff-6e0aafdc3c01d2ed637500356503248eR27

image


app/controllers/settings.js
```diff
@@ -32,6 +32,15 @@ export default Controller.extend({
     return isPrivate && settings.hasOwnProperty('allow_config_imports');
   }),
 
+  showBetaFeatures: computed('showConfigValidationSwitches', function () {
+    return this.get('showConfigValidationSwitches');
+  }),
```
sorta: This could possibly be simplified by using `reads` rather than computed. It would be like:
```
import { alias, filterBy, reads } from '@ember/object/computed';

// Some other controller stuff here

  showBetaFeatures: reads('showConfigValidationSwitches'),
```

app/controllers/settings.js
```diff
@@ -32,6 +32,15 @@ export default Controller.extend({
...
+
+  showConfigValidationSwitches: computed('model.settings', function () {
+    let settings = this.get('model.settings');
+    return settings.hasOwnProperty('config_validation');
+  }),
```
sorta: Similarly, some of the computed property functions that ember provides could simplify this, although, it might require 2 of them. Something like:
```
  hasNoConfigValidation: none('model.settings.config_validation'),
  showConfigValidationSwitches: not('hasNoConfigValidation'),
```

We've been trying to move more towards using these when possible. Here's the docs for these functions https://api.emberjs.com/ember/3.12/modules/@ember%2Fobject#functions-computed

app/templates/settings.hbs
```diff
@@ -152,6 +152,38 @@
...
     </section>
   {{/if}}
+  {{#if this.showBetaFeatures}}
+    <section class="settings-section beta_features">
+      <h2 class="small-title display-inline">
+        Beta Features
+      </h2>
+      <p>
+        These features are in beta. If you find things to be not working as expected
+        or lacking, please let us know in the <a href="https://travis-ci.community">Travis CI community forum</a>
```
sorta: We have a bunch of URLs in `config/environment.js`, so that we can just have them in one place and easily change if we need to. Looks like one already exists for the community site, so with the`config-get` helper, this link could be `<a href={{config-get 'urls.community}}>`

app/templates/settings.hbs
```diff
@@ -152,6 +152,38 @@
...
+        or email <a href="mailto:support@travis-ci.com">support</a>.
```
sorta: Same idea here, this could be `<a href={{config-get 'urls.support'}}>`

app/controllers/settings.js
svenfuchs: done 🎉

app/templates/settings.hbs
svenfuchs: done 🎉

app/templates/settings.hbs
svenfuchs: done 🎉

app/controllers/settings.js
svenfuchs: done 🎉

svenfuchs:
thanks for the great suggestions @sorta! i've made all the changes :)

PR #2168, opened 2019-10-04
Update GTM to track all pages

sorta:
For https://travisci.assembla.com/spaces/Web/tickets/216/details


sorta:
A couple of things:

I thought that GOOGLE_TAGS_PARAMS was required, but this is not the case. And in fact, a blank value is what's needed to generate the snippet shown in the ticket. It basically associates with a specific GTM Container, but not with any particular GTM environment. But I think we'll want to leave the staging environment GOOGLE_TAGS_PARAMS as they are, so that they can be limited to the staging environment in GTM.

Also, I'm leaving the existing `gaCode` google analytics stuff in the application route as-is for now, but I think we should consider updating all things like that to use `ember-metrics`. But, this is something that we would want to coordinate with the Product team etc. It's a big enough change that it should be its own ticket.

sorta:
Hmm not sure why GTM isn't connecting on the PR deployment. But I've done a deploy to staging and that's working, so you can test there https://staging.travis-ci.org/

sorta:
I ended up deciding to convert the old manual Google Analytics calls to use the metrics service. There were only a small number of them and it was easy enough to do.

Also, I tested to confirm that the metrics service responds fine to having an empty metricsAdapters array. So I've updated the `config.metricsAdapters` array to always be present to simplify things. The try block in `application.js` is probably not necessary, but since it runs on every route transition, I think it's good to have just in case.

app/routes/application.js
```diff
@@ -24,11 +25,17 @@ export default TravisRoute.extend(BuildFaviconMixin, {
...
-      }
-    });
+    if (config.metricsAdapters.length > 0) {
+      const { metrics, raven, router } = this;
+      router.on('routeDidChange', () => {
+        try {
+          const { currentURL: page } = router;
+          metrics.trackPage({ page });
+        } catch (err) {
+          raven.logException('Metrics error');
```
artursmirnov: Hmm.. I'm wandering would it send errors to Sentry when tracking is disabled by adblockers? In that case we may have a loooot of the errors that we cannot avoid.

sorta: Ahhh, that's a good point! Maybe I'll remove this for now, and we can reintroduce it in the future if needed

PR #2174, opened 2019-10-08
add a badge to "view config" if msgs > info are present

svenfuchs:
with no `alert`, `error`, or `warn` level message: https://sf-config-msgs-badge.test-deployments.travis-ci.org/svenfuchs/test/jobs/590057267/config
image

with a `warn` level message https://sf-config-msgs-badge.test-deployments.travis-ci.org/svenfuchs/test/jobs/595157604/config
image

with an `error` and `warn` level message https://sf-config-msgs-badge.test-deployments.travis-ci.org/svenfuchs/test/jobs/595158948/config
image


app/templates/components/job-tabs.hbs
```diff
@@ -15,14 +15,22 @@
...
       >
         {{#if this.job.build.request.noYaml}}
           <EmberTooltip @text="Try a more recent job to see the config" />
         {{/if}}
+
         View config
+
+        {{#if this.messagesMaxLevel}}
+          <span class="tooltip-wrapper">
+            <SvgImage @name="circle" @class={{this.messagesBadgeIconClass}} />
```
artursmirnov: I'm not sure about using `svg` here. Just `span` or `div` with `border-radius: 50%` would make the same circle. It's really minor though, so just mentioning it here :)

svenfuchs: good point, i'll change that :)

svenfuchs:
@artursmirnov wow, that's odd. i think it might have to do with the fact that the `job_tabs` component is used on both the build and job views ... e.g. the badge shows up if i navigate to https://sf-config-msgs-badge.test-deployments.travis-ci.com/travis-ci/travis-web/jobs/243876491 ... but not when i then reload the page on that URL, so i am guessing there also might be something wrong with the `computed` property's dependency?

app/components/job-tabs.js
```diff
@@ -1,11 +1,24 @@
...
+import { computed } from '@ember/object';
 import { isEmpty } from '@ember/utils';
 import Component from '@ember/component';
 
 export default Component.extend({
-
   tagName: 'div',
   classNames: ['travistab'],
 
+  messagesMaxLevel: computed('request.messages', function () {
```
artursmirnov: This might be a culprit for badge not showing up 😅 Currently the computed property tracks only the `request.messages` property itself, but not it's content. E.g. the CP will re-calculate on `set('request.mesages', something)`, but it won't react on `get('request.messages').pushObject(message)` - because it doesn't track _content_ changes. In order to make Ember track contents, we should provide it a path:
```suggestion
  messagesMaxLevel: computed('request.messages.@each.level', function () {
```
After this, Ember will start also tracking the `level` property of each array item, the array changes (add, remove, etc.) and the property assignment.

app/templates/components/build-layout.hbs
```diff
@@ -21,7 +21,7 @@
       {{/if}}
     {{else}}
       <JobInfrastructureNotification @job={{this.job}} />
-      <JobTabs @job={{this.job}} @repo={{this.repo}} />
+      <JobTabs @job={{this.job}} @request={{this.job.build.request}} @repo={{this.repo}} />
```
artursmirnov: Is it necessary to pass `@request` separately? It feels like it could be derived from `this.job` inside of the component

app/templates/job.hbs
```diff
@@ -7,7 +7,7 @@
       @repo={{this.repo}}
     />
     <JobInfrastructureNotification @job={{this.job}} />
-    <JobTabs @job={{this.job}} @repo={{this.repo}} />
+    <JobTabs @job={{this.job}} @request={{this.job.build.request}} @repo={{this.repo}} />
```
artursmirnov: The same here about `@request`

app/components/job-tabs.js
svenfuchs: aaaah! nice. i didn't know about `@each` in those keys.

artursmirnov: Yeah, and sometimes you can also find `collection.[].prop` syntax - it's essentially the same, just a shorthand.
But there's a limitation with this key - it only works one level deep, so `collection.@each.prop.another` won't work (it'll throw a warning)

app/templates/components/build-layout.hbs
svenfuchs: moved it back 👍

app/templates/job.hbs
svenfuchs: same here.

app/components/job-tabs.js
```diff
@@ -1,11 +1,24 @@
...
 import { isEmpty } from '@ember/utils';
 import Component from '@ember/component';
 
 export default Component.extend({
-
   tagName: 'div',
   classNames: ['travistab'],
 
+  messagesMaxLevel: computed('this.job.build.request.messages.@each.level', function () {
+    let msgs = A(this.get('request.messages')).rejectBy('level', 'info');
```
artursmirnov: Looks like `this.request` does not exist anymore. So `msgs` will always be empty. 
```suggestion
    let msgs = A(this.get('job.build.request.messages')).rejectBy('level', 'info');
```

app/components/job-tabs.js
```diff
@@ -1,11 +1,24 @@
...
+import { computed } from '@ember/object';
 import { isEmpty } from '@ember/utils';
 import Component from '@ember/component';
 
 export default Component.extend({
-
   tagName: 'div',
   classNames: ['travistab'],
 
+  messagesMaxLevel: computed('this.job.build.request.messages.@each.level', function () {
```
artursmirnov: The path seems to be invalid, context should not be included into it:
```suggestion
  messagesMaxLevel: computed('job.build.request.messages.@each.level', function () {
```

PR #2175, opened 2019-10-08
[Plans] Handle UTM params in URL

sorta:
For https://travisci.assembla.com/spaces/Web/tickets/220/details


app/utils/object-collect.js
```diff
@@ -0,0 +1,16 @@
...
+      const item = this.get(key);
+      if (isPresent(item)) {
+        collection[key] = item;
+      }
+      return collection;
+    }, {});
+
+    return result;
+  });
+}
```
piotr-travisci: Nice tool 👍

sorta:
Thanks for the review @piotr-travisci & @artursmirnov! :heart:

RE the snake casing: yeah, I don't particularly like that part of it. One alternative could be to gather the queryParams together into `@utmFields` in the controller. I thought that might be weird, but that was before I abstracted things a bit more into the `lead` model and the `objectCollect` util. Idk, what do you think?

app/utils/object-collect.js
sorta: Thanks! :smile:

artursmirnov:
@sorta I really like how simple it is now, and I don't think that case in naming worth code complication. Let's just keep it as it is - it's so good 👍

sorta:
Ok yeah, that sounds good to me :+1:  Thanks!

PR #2177, opened 2019-10-09
Error handling at JSON.parse

piotr-travisci:


piotr-travisci:
Thank you @artursmirnov for your review.
As always, it's nice to read your detailed and constructive comments!

PR #2180, opened 2019-10-10
[Plans] Rename lead_source to referral_source

sorta:
Renaming at the request of the marketing team. Related API PR is https://github.com/travis-ci/travis-api/pull/976


sorta:
In retrospect, it probably wasn't necessary to change the name in every single place, but I like when this kind of stuff is consistent, and it wasn't difficult or time-consuming to do.

sorta:
Excellent, thanks for the review @piotr-travisci! :tada:

PR #2185, opened 2019-10-15
Enable sitewide utm capture

sorta:
Continuing on https://travisci.assembla.com/spaces/Web/tickets/220/details
It was requested that "the UTMs must captured regardless of how many pages the user visited in between"

The update in this PR should hopefully address this request


artursmirnov:
Hey @sorta, I would like to suggest a bit different approach. What do you think about a service that would store all UTM params globally? Something like `service('utm')`. With this service one could be able to read utm params from anywhere in the app any time. (Would be nice to make them read-only). E.g.:
```js
  utm: service(),
  utmSource: reads('utm.source'),
  save() {
    const utmParams = this.utm.collect('source', 'medium') // returns { utm_source: 'source_value', utm_medium: 'medium_value' }
  }
})
```
Due to the UTM params nature, they just need to be captured once - on application load. Thus, we can add this service to, say, application route and call `capture` in `init`:
```js
export default ApplicationRoute = Route.extend({
  utm: service(),

  init() {
    this._super(...arguments);
    this.utm.capture();
  }
})
```
Also it would be nice if `capture` method would also clear url from the utm params, because it doesn't make sense to keep them.

To simplify capturing of query params in service, there is a nice add-on: https://github.com/NullVoxPopuli/ember-query-params-service

What do you think?

sorta:
@artursmirnov it sounds like a good idea to me! Thanks for the suggestion, I'll give it a try :smile:

sorta:
Ready again for code review when you get a chance @artursmirnov. Thanks!

PR #2187, opened 2019-10-15
Swap flex-grow and flex-shrink styles for .job-arch column

jakemac53:
Currently this is set with `flex-grow: 1` which allows it to grow, but that adds a lot of extra whitespace.

Setting instead the `flex-grow: 0` and `flex-shrink: 1` gives a better experience. It doesn't grow unnecessarily and can shrink a bit at smaller screen widths still (which was already happening).

## Before

with-grow

## After

with-shrink

### DIFF
--- a/app/styles/app/layouts/jobs.scss
+++ b/app/styles/app/layouts/jobs.scss
@@ -215,7 +215,7 @@
 
   @media #{$medium-up} {
     position: relative;
-    flex: 1 0 6em;
+    flex: 0 1 6em;
     padding: 0.3em 0.3em 0.5em 0.7em;
     align-self: flex-start;
     white-space: nowrap;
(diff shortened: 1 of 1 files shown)


jakemac53:
@emmanuel-ci anything I should do regarding the ci failure? It doesn't look related to this, the error is `Error response from daemon: Get https://quay.io/v2/: unauthorized: Invalid Username or Password`.

jakemac53:
> In the meantime, just a minor suggestion, would it make sense to just have: `flex-basis: 6em`? `flex: 0 1 auto` is already the default.

Sure, done

jakemac53:
@emmanuel-ci I would be happy to but I can't push a branch to this repo directly since I am not a contributor.

jakemac53:
(feel free to make this change separately as well if that is easier, I don't need credit for the commit I just want the style to be updated :D )

PR #2192, opened 2019-10-17
Resolve current npm audit security issues

sorta:
For https://travisci.assembla.com/spaces/Web/tickets/223/details

### DIFF
--- a/app/styles/app/layouts/log.scss
+++ b/app/styles/app/layouts/log.scss
@@ -107,7 +107,7 @@
     padding-right: 16px;
     text-align: right;
     color: #999;
-    background: inline-image("ui/workers-close.svg") no-repeat right 6px;
+    background: url(/images/ui/workers-close.svg) no-repeat right 6px;
     @supports(position: sticky) {
       position: sticky;
       float: right;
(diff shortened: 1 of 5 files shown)


sorta:
I had npm audit fix all of the things that could be automatically fixed. It was pretty straightforward and only changed `package-lock.json`

sorta:
Regarding this advisory https://www.npmjs.com/advisories/785 which is for a 'Regular Expression Denial of Service' vulnerability in `clean-css`:

I'm not fully decided yet, but I may leave this as-is. The severity is low, and it seems like trying to resolve it would not be trivial. It's a build-time-only dependency, it shouldn't present a vulnerability while running, as far as I can tell.

sorta:
In regards to the Prototype Pollution vulnerabilities that were not able to be automatically resolved:

Some info at
https://npmjs.com/advisories/1065
https://npmjs.com/advisories/782
https://npmjs.com/advisories/577

The severity is high for a couple of those, so I would like to address it if possible. The issue is caused by an older lodash version listed as a dependency for `ember-cli-inline-images`. It doesn't appear to be very actively maintained, but we could try to reach out to the maintainers about updating lodash. However, I noticed that the ember addon is only used in a small handful of places, and so I experimented with removing it altogether as a short term fix. It was fairly easy to convert, and this also has the benefit of reducing our dependencies.

The one place where I think there may be a performance concern to consider is the favicons. The favicons change according to build status on certain pages, and having the favicons inlined as dataURIs maybe makes a lot of sense, since it would otherwise result in a lot of requests for the favicon images. So I'm not yet 100% sold on this approach. An alternative may be to hardcode the inlined favicon images?

sorta:
Also, I removed `ember-cli-htmlbars-inline-precompile` because I saw a message during an `ember serve` that said

> DEPRECATION: ember-cli-htmlbars-inline-precompile is no longer needed with ember-cli-htmlbars versions 4.0.0 and higher, please remove it from `/home/sorta/wrk/travis-web/package.json`

PR #2194, opened 2019-10-21
As fix auth

artursmirnov:
Ticket: https://eu-app.assembla.com/spaces/Web/tickets/realtime_cardwall?ticket=222

This PR also includes some relevant clean up of the legacy code that has been refactored in scope of authorization.

The major update here is that the app now uses only one storage (either `localStorage` or `sessionStorage`) - the one that has been updated last, so no synchronization is involved anymore. Another important improvement is that the authorization information in the app now is read-only. The app only reads it and never writes. So it's always exactly the same that has been set by authorizer.


artursmirnov:
It works for me locally and I expect it to work on test deployments, but lots of tests are failing. Nevertheless, it'd be good to get some early feedback if you have some time.

app/services/auth.js
```diff
@@ -1,276 +1,196 @@
...
+  //           return Promise.reject();
+  //         }
+  //       })
+  //       .then(({ installation = null }) => {
+  //         this.currentUser.setProperties({ installation });
+  //       });
+  //   } else {
+  //     return Promise.resolve();
+  //   }
+  // },
```
rafael-travis: Comment needed?

app/services/auth.js
```diff
@@ -313,3 +220,18 @@ export default Service.extend({
...
+  const record = store.push(store.normalize('user', user));
+  const installation = store.peekAll('installation').findBy('owner.id', user.id) || null;
+  record.setProperties({ installation });
+  return record;
+}
+
+function runAfterSignOutCallbacks() {
+  afterSignOutCallbacks.forEach((callback) => {
+    callback();
+  });
```
rafael-travis: would `afterSignOutCallbacks.forEach(callback)` work?

app/services/auth.js
artursmirnov: Yeah, this method is quite complex and I kept it for reference while fixing tests. I'll make sure to clean up before merge 👍

app/services/auth.js
artursmirnov: It should! Great suggestion, thanks 👍

artursmirnov: Ah, no, it won't. `callback` is a function that's passed to iterator, so it's undefined outside of it.

artursmirnov: Though, I did simplify it like this: `afterSignOutCallbacks.forEach(callback => callback());`

rafael-travis: Ah, right lol, I misread it. Nice!

artursmirnov:
Hey folks, this PR is green now and ready for the final review 🎉 Could you please take a look when you have time?

app/models/user.js
```diff
@@ -1,45 +1,37 @@
...
   allowMigration: attr('boolean'),
 
-
   type: 'user',
 
-  fullName: computed('name', 'login', function () {
-    let name = this.name;
-    let login = this.login;
-    return name || login;
+  fullName: or('name', 'login'),
```
piotr-travisci: Like Scouts do it:
"Leave it better than you found it."
👏

PR #2197, opened 2019-10-22
Display arch name as it is unless there is a special pretty value for it

pavel-d:


pavel-d:
@artursmirnov, could you please take a quick look when you got a moment?

PR #2203, opened 2019-10-25
Remove jQuery from travis-web

sorta:
For https://travisci.assembla.com/spaces/Web/tickets/197/details

This PR removes the last bit of jquery that is used directly in our app.


tests/acceptance/builds/cancel-test.js
```diff
@@ -30,6 +29,8 @@ module('Acceptance | builds/cancel', function (hooks) {
     await click('[data-test-repo-actions-cancel-button]');
 
     assert.dom('[data-test-flash-message-text]').hasText('Build has been successfully cancelled.', 'cancelled build notification should be displayed');
-    assert.equal($('head link[rel=icon]').attr('href'), getFaviconUri('yellow'), 'expected the favicon data URI to match the one for running');
+    // Ember-test-helpers find does not work here
+    const iconHref = window.document.querySelector('head link[rel=icon]').getAttribute('href');
+    assert.equal(iconHref, getFaviconUri('yellow'), 'expected the favicon data URI to match the one for running');
```
artursmirnov: I think you can use [qunit-dom](https://github.com/simplabs/qunit-dom/blob/master/API.md#hasattribute) here:
```js
assert.dom('head link[rel=icon]').hasAttribute('href', getFaviconUri('yellow'));
```

tests/acceptance/job/basic-layout-test.js
```diff
@@ -60,7 +59,9 @@ module('Acceptance | job/basic layout', function (hooks) {
 
     assert.equal(document.title, 'Job #1234.1 - travis-ci/travis-web - Travis CI');
 
-    assert.equal($('head link[rel=icon]').attr('href'), getFaviconUri('green'), 'expected the favicon data URI to match the one for passing');
+    // Ember-test-helpers find does not work here
+    const iconHref = window.document.querySelector('head link[rel=icon]').getAttribute('href');
+    assert.equal(iconHref, getFaviconUri('green'), 'expected the favicon data URI to match the one for passing');
```
artursmirnov: Same here with qunit-dom

tests/acceptance/job/basic-layout-test.js
```diff
@@ -118,7 +119,9 @@ module('Acceptance | job/basic layout', function (hooks) {
...
```
artursmirnov: Same here :)

tests/acceptance/builds/cancel-test.js
sorta: ahh nice, I don't think I had tried qunit-dom here :) Though when I tried it just now, it couldn't find the element

element-exist-error

It's kind of strange, I would think that qunit-dom should be able to see it?

tests/acceptance/job/basic-layout-test.js
sorta: Got the same result here. qunit-dom was unable to find the element, and it broke the test. Maybe because it's in the `head`? Well, for now I think I'm going to leave these as-is, but thanks for the suggestion! It was at least worth trying :smile:

PR #2210, opened 2019-10-30
Hint for ssh private key format.

piotr-travisci:


app/styles/app/modules/forms.scss
```diff
@@ -103,6 +103,14 @@ textarea {
       max-width: 460px;
     }
   }
+
+  .ssh-hint {
+    float: right;
```
artursmirnov: I'd suggest to avoid using floats and use flex layout wherever possible. Floats require clears and may break layout in some browsers otherwise. But if you feel that it's solid and reliable in this case, then I'm totally fine with this solution 👍

piotr-travisci: @artursmirnov you are absolutely right. Event those two works the same it's always better to use newer stack. Styles refactored and waiting for Creative Team approval.

PR #2212, opened 2019-10-31
Loading indicator when waiting for status

piotr-travisci:


app/templates/components/repo-actions.hbs
```diff
@@ -66,4 +66,8 @@
       </div>
     {{/if}}
   {{/if}}
+{{else}}
+  {{#if this.islist}}
+    <LoadingIndicator @height={{true}} />
```
artursmirnov: Doesn't it look good even without `height` and styles tweaking? Here's how it looks to me (I tested only in chrome though):

2019-11-12 at 17 10

Did you maybe notice something that I missed? 😅

piotr-travisci: You are absolutely right. Thank you @artursmirnov

PR #2214, opened 2019-11-01
tweak build config view

svenfuchs:
this does:

* allow folding config sources
* fold empty sources by default (api is empty :alot:)
* display resulting build config
* remove the line between the header div and the foldable container (removed this accidentally when trying to remove a duplicate border there when folded, and then found it looked a lot more clean)

needs `travis-api/sf-request-config`: https://github.com/travis-ci/travis-api/pull/980

image

### DIFF
--- a/app/components/request-config.js
+++ b/app/components/request-config.js
@@ -1,0 +1,35 @@
+import Component from '@ember/component';
+import { computed } from '@ember/object';
+import { later } from '@ember/runloop';
+
+export default Component.extend({
+  copied: false,
+  isExpanded: true,
+
+  toggleStatusClass: computed('isExpanded', function () {
+    return this.isExpanded ? 'expanded' : 'collapsed';
+  }),
+
+  buttonLabel: computed('copied', function () {
+    return this.copied ? 'Copied!' : 'Copy build config';
+  }),
+
+  formattedConfig: computed('config', 'slug', function () {
+    const config = this.get('config');
+    try {
+      return JSON.stringify(config, null, 2);
+    } catch (e) {
+      return config;
+    }
+  }),
+
+  actions: {
+    copied() {
+      this.set('copied', true);
+      later(() => this.set('copied', false), 3000);
+    },
+    toggle() {
+      this.toggleProperty('isExpanded');
+    }
+  }
+});
--- a/app/templates/components/request-config.hbs
+++ b/app/templates/components/request-config.hbs
@@ -1,18 +1,9 @@
 <div class="inner-yaml-container">
   <div class="header" {{on 'click' (action 'toggle')}}>
-    {{#if this.fileUrl}}
-      <ExternalLinkTo @href={{this.fileUrl}}>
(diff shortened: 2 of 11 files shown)


sorta:
Hi @svenfuchs :smile: I'm reviewing the code and so far it's looking pretty good, but I'm getting errors when I try to test it:

https://sf-request-config.test-deployments.travis-ci.com/travis-ci/travis-web

Screenshot from 2019-11-04 09-16-04

I'm guessing it's because https://github.com/travis-ci/travis-api/pull/980 is not in production yet? If so, would it be possible to merge and deploy that ahead of this? Or alternatively, deploying to staging?

app/components/request-config.js
```diff
@@ -0,0 +1,35 @@
...
+  toggleStatusClass: computed('isExpanded', function () {
+    return this.get('isExpanded') ? 'expanded' : 'collapsed';
+  }),
+
+  buttonLabel: computed('copied', function () {
+    return this.copied ? 'Copied!' : 'Copy build config';
+  }),
+
+  formattedConfig: computed('config', 'slug', function () {
+    let config = this.get('config');
```
sorta: Since `config` doesn't seem to be reassigned a value in this function, I would use `const` rather than `let` here. We don't have a set policy on when to use one or the other, so consider this optional, but I generally use `const` unless I need to reassign the variable.

app/templates/components/raw-config.hbs
```diff
@@ -1,5 +1,5 @@
 <div class="inner-yaml-container">
-  <div class="file-name">
+  <div class="header" {{action 'toggle'}}>
```
sorta: I would use the `on` modifier here to make this more explicit, like: `{{on 'click' (action 'toggle')}}`

Here is some more information about it https://github.com/emberjs/rfcs/blob/master/text/0471-on-modifier.md

app/templates/components/request-config.hbs
```diff
@@ -0,0 +1,30 @@
+<div class="inner-yaml-container">
+  <div class="header" {{action 'toggle'}}>
```
sorta: Same here about using an `on` modifier.

tests/acceptance/config/yaml-test.js
```diff
@@ -123,7 +123,7 @@ module('Acceptance | config/yaml', function (hooks) {
     test('shows all unique raw configs', async function (assert) {
       await visit(`/travis-ci/travis-web/builds/${this.build.id}`);
       await page.yamlTab.click();
-      assert.equal(page.yaml.length, 2, 'expected two yaml code block');
+      assert.equal(page.yaml.length, 3, 'expected two yaml code block');
```
sorta: Should the text here say `expected three yaml code block`? :)

svenfuchs:
> I'm guessing it's because travis-ci/travis-api#980 is not in production yet?

Oh, right, probably, yes. I've just deployed it.

PR #2216, opened 2019-11-06
Missing first_sync page restored + UI beautification

piotr-travisci:


app/services/auth.js
```diff
@@ -197,8 +201,8 @@ export default Service.extend({
 
   syncingDidChange: observer('isSyncing', 'currentUser', function () {
     const user = this.currentUser;
-    if (user && user.get('isSyncing') && !user.get('syncedAt')) {
-      return this.router.transitionTo('first_sync');
+    if (user && !user.get('syncedAt')) {
+      this.first_sync = true;
```
sorta: Hmm, I thought that `this.set` was still required for stuff like this, but I guess this seems to work?

piotr-travisci: Yes it works perfectly. Thank you @sorta

piotr-travisci:
> Hmm, I thought the tests were passing for this, but it looks like they're not?

Previous implementation worked in browser, but broke all tests regarding to broken tasks.
I reimplemented it and left "newbie" validation at API side.

app/models/owner.js
```diff
@@ -18,8 +18,8 @@ export default VcsEntity.extend({
   name: attr('string'),
   login: attr('string'),
   channels: attr(),
-  isSyncing: attr('boolean'),
-  syncedAt: attr(),
+  isSyncing: attr(''),
+  syncedAt: attr(''),
```
sorta: Does `attr('')` have a different effect than `attr()`? If not, then let's change these to `attr()` for consistency, it seems to be what we use more often

app/models/user.js
```diff
@@ -16,8 +16,9 @@ export default Owner.extend({
   token: attr('string'),
   secureUserHash: attr('string'),
   gravatarId: attr('string'),
-  firstLoggedInAt: attr('date'),
+  firstLoggedInAt: attr(),
   allowMigration: attr('boolean'),
+  recentlySignedUp: attr(''),
```
sorta: Same here about `attr('')`

app/models/owner.js
piotr-travisci: @sorta Great finding. Thank you. I already corrected it.

app/models/user.js
piotr-travisci: @sorta Great finding. Thank you. I already corrected it.

piotr-travisci:
Thank you @sorta for quick review 🌷

PR #2221, opened 2019-11-08
Pm fix first sync page

piotr-travisci:


app/models/owner.js
```diff
@@ -18,7 +18,7 @@ export default VcsEntity.extend({
   name: attr('string'),
   login: attr('string'),
   channels: attr(),
-  isSyncing: attr('boolean'),
```
artursmirnov: Is there any reason to remove the type from attribute? Usually it's better to provide types, because it's not only helps to visually understand the data, but also helps to _ensure_ the expectations. If no type is passed to the `attr`, Ember Data will use `raw` transform, which doesn't parse the input and assigns whatever comes from API. Thus, if for some reason the format will change on API, the UI will just brake.

app/models/user.js
```diff
@@ -16,8 +16,9 @@ export default Owner.extend({
   token: attr('string'),
   secureUserHash: attr('string'),
   gravatarId: attr('string'),
-  firstLoggedInAt: attr('date'),
```
artursmirnov: Same here with regards to the attribute type. The [`date` transform](https://api.emberjs.com/ember-data/release/classes/DateTransform) parses the date string to the `Date` object. Thus, before the change this attribute was a `Date` object, after the change it's just a string, e.g. `2019-11-11T00:50:23Z`

app/services/auth.js
```diff
@@ -116,9 +116,11 @@ export default Service.extend({
 
       Travis.trigger('user:signed_in', this.currentUser);
 
-      this.reloadCurrentUser().then(() =>
-        Travis.trigger('user:refreshed', data.user)
-      );
+      this.reloadCurrentUser().then(() => {
+        Travis.trigger('user:refreshed', data.user);
+        if (this.currentUser && this.currentUser.get('recentlySignedUp') && this.currentUser.get('recentlySignedUp') === true)
+          this.metrics.trackPage({ page: '/first_sync' });
```
artursmirnov: This doesn't look right to me. 

1) Wouldn't it report `/first_sync` page visit twice for new users? First one - here, second one - after actual redirect to the `first_sync` page?

2) I'd suggest to use `reportNewUser` method instead, as it's already in place (see below) and aimed for the same purpose. You could just change it as necessary and invoke wherever it work best.

artursmirnov:
Hey @piotr-travisci, thanks for catching this up! I understand that this PR is already merged, but I see potential significant issues with it. Could you please take a look at my comments when you have a chance and maybe revisit it? Thanks!

PR #2226, opened 2019-11-12
First authentication metrics improved

piotr-travisci:


app/services/auth.js
```diff
@@ -171,13 +169,17 @@ export default Service.extend({
...
     const signupUsers = this.storage.signupUsers || [];
 
     if (!syncedAt && !signupUsers.includes(login)) {
       metrics.trackPage({ page: '/virtual/signup-success' });
       this.storage.signupUsers = signupUsers.concat([login]);
     }
+    if (recentlySignedUp && recentlySignedUp === true)
+      this.metrics.trackEvent({
+        event: 'first_authentication'
+      });
```
artursmirnov: I'd like to suggest to combine these two reports together. Ines asked to remove `/virtual/signup-success` anyway, and `!signupUsers.includes(login)` may be helpful for `first_authentication` event as well. It just ensures that current user has not yet reported the signup.

piotr-travisci: Thank you @artursmirnov 
Good catch. I refactored it.

PR #2231, opened 2019-11-14
Cron jobs fetch refactored

piotr-travisci:


app/models/repo.js
```diff
@@ -141,13 +141,21 @@ const Repo = VcsEntity.extend({
...
+  cronJobs: computed('id', 'fetchCronJobs.lastSuccessful.value', function () {
+    const crons = this.fetchCronJobs.get('lastSuccessful.value');
+    if (!crons) {
+      this.get('fetchCronJobs').perform();
+    }
+    return crons || [];
   }),
 
+  fetchCronJobs: task(function* () {
+    const id = yield this.id;
```
artursmirnov: Hmm.. Are you sure the `yield` makes sense here? I don't think `this.id` returns promise.

Also it looks like `this.store.filter` should have filter, shouldn't it?

piotr-travisci: Thank you Artur!

PR #2238, opened 2019-11-18
Refactor jquery out of ajax and api services

sorta:
Another attempt at this PR https://github.com/travis-ci/travis-web/pull/2224, which ended up causing the settings page to fail.


sorta:
Ok I figured out what was going on on the settings page. I updated a test to fail when `/repos/:id/key` returns a 404 on the settings page, and then fixed the issue. I want to do some more manual testing of various spots around the app, but otherwise I think this is ready for review again.

mirage/config.js
```diff
@@ -73,7 +73,7 @@ export default function () {
   });
 
   this.get('/users/permissions', (schema, request) => {
-    let authorization = request.requestHeaders.Authorization;
+    const { authorization } = request.requestHeaders;
```
artursmirnov: Hmm.. Are you sure about this change? Before it was with capital `A` (and JS is case sensitive). Just wanted to bring it to your attention.

sorta:
Thanks for the detailed feedback @artursmirnov! You're absolutely right, and it was the inclusion of the `travis-api-version` header that seems to have been causing the 404. I've updated that request to not include that header, and the test deployment settings page should now have a succeeding `/repos/:id/key` request. Thanks for catching it!

mirage/config.js
sorta: Thanks for bringing it to my attention :) This change is due to the fact that `fetch` lowercases header names

https://github.com/ember-cli/ember-fetch/issues/100
https://github.com/github/fetch/issues/249

This resulted in an issue where some tests were failing because this function was now receiving the header as `authorization` rather than `Authorization`, so the change is intentional. Although, now that I think of it, maybe it would be safer to do like an `authorization || Authorization` approach. What do you think?

artursmirnov: Ah, I see! Thanks for clarifications, I didn't know that :) To be honest, this behavior of fetch is kinda weird - not all backends are case insensitive.

In this case, let's keep it lowercase, so that we had in tests exactly the same as in production 👍

sorta: Sounds good :) Yeah I agree that it's weird, and was quite surprised when I found out about it. A lot of people might have limited control over their backend, or updating the backend for it might take a huge amount of work. But it seems like the preferred way forward

app/services/ajax.js
```diff
@@ -1,190 +1,162 @@
...
+  handleFetchError(reject, error) {
+    reject(error);
+    this.logFetchError(error);
+  },
+
+  logFetchError(response) {
+    if (this.features.get('debugLogging')) {
+      const { status = 'UNKNOWN' } = response;
+      // eslint-disable-next-line
+      console.log(`[ERROR] Fetch error (${status}): ${response}`);
```
piotr-travisci: Is it worth to send it also to Sentry?

sorta: Thanks for the review :smile: That's a good question! I've gone back and forth about it. On the one hand, yes it would be handy to have these show up in sentry and maybe help catch some issues that we would otherwise miss. But on the other, it could result in a lot of noise in sentry. I could honestly go either way, but I'd be interested to hear @artursmirnov's opinion on it

artursmirnov: Also for intentional logging (e.g. to make other devs aware of something) it's better to use [warn](https://api.emberjs.com/ember/3.14/functions/@ember%2Fdebug/warn) as it doesn't expose logging to production builds.

sorta: Oooh nice I like that! and then we can also remove the eslint disable comment

sorta: Ok, I've updated the function to use sentry and `warn` instead of `console.log`

artursmirnov: About the sentry, I agree, it could be too noisy. And also this general case could cover well-handled errors, like, for instance, `401` or `404` on repo pages, etc. We don't need such errors in Sentry as it's already too overwhelmed :) I'd better suggest to handle API errors on a per-case basis (where the particular endpoint is used)

sorta: ahh yeah, good point :) I've removed the call to sentry, and left in the use of `warn`

artursmirnov: haha, I'm sorry for being so slow 😄

sorta: hah no worries :D I was already looking at and working on this, so I happened to respond to it very quickly

artursmirnov:
Hey @sorta, it looks great to me, thanks for your work on this! I spent some time testing the API comms - and it looks identical to production, so I think it should be good to go 🚀

PR #2241, opened 2019-11-19
Fix issue where .org test deployments act as if TRAVIS_PRO=true

sorta:
Today I noticed an interesting issue. Some .org PR test deployments appear to be behaving as if TRAVIS_PRO=true. The landing page and plans page are 2 places that I know of where you can easily spot a difference. I haven't yet been able to figure out a pattern behind which PR deployments are affected, nor am I really sure why.

For example, compare:
https://travis-ci.org/
https://greenkeeper-ember-resolver-6-0-0.test-deployments.travis-ci.org/

https://travis-ci.org/plans
https://greenkeeper-ember-resolver-6-0-0.test-deployments.travis-ci.org/plans

### DIFF
--- a/config/environment.js
+++ b/config/environment.js
@@ -165,7 +165,7 @@ module.exports = function (environment) {
 
   const { TRAVIS_PRO, TRAVIS_ENTERPRISE, SOURCE_ENDPOINT } = process.env;
 
-  if (TRAVIS_PRO && TRAVIS_PRO !== 'false') {
+  if (TRAVIS_PRO) {
     ENV.featureFlags['pro-version'] = true;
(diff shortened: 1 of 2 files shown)


sorta:
I'm actually not sure if this addresses the issue

sorta:
Going to resume some testing

sorta:
So actually, this does seem to address the issue. I'm not sure why, as far as I can tell, no value at all should be coming through for TRAVIS_PRO on any of the .org deployments

sorta:
@artursmirnov found the source of the issue, TRAVIS_PRO was set at https://travis-ci.com/travis-ci/travis-web/settings. Thanks!

PR #2246, opened 2019-11-21
Fix for duplicated jobs in Running tab

piotr-travisci:
https://travisci.assembla.com/spaces/Support/tickets/realtime_cardwall?ticket=52


app/templates/components/queued-jobs.hbs
```diff
@@ -1,37 +1,41 @@
...
-    >
-      {{#if job.repo.slug}}
-        <StatusIcon @status={{job.state}} />
-        <LinkTo @route="job" @models={{array job.repo job}}>
-          {{job.repo.slug}}
-        </LinkTo>
-      {{/if}}
-      <p class="tile-title float-right">
-        <SvgImage @name="icon-hash" @class="icon" />
+  {{#if this.queuedJobs.length}}
```
artursmirnov: This `{{#if}}{{/if}}` in not necessary as `{{#each}} list {{else}} list empty {{/each}}` can handle it.

app/templates/components/running-jobs.hbs
```diff
@@ -1,6 +1,6 @@
 {{#if this.jobs.isLoaded}}
-  {{#if this.jobs.length}}
-    {{#each this.jobs as |job|}}
+  {{#if this.runningJobs.length}}
```
artursmirnov: Same here with `{{#each}}{{else}}{{/each}}`. I understand that it was here before, but it's easy to improve ;)

PR #2247, opened 2019-11-21
Remove jQuery from build and job tab components

sorta:
Split from https://github.com/travis-ci/travis-web/pull/2203

The `<LinkTo` components handle highlighting automatically on the build.index / build.config / job.index / job.config routes. This functionality covers the case where the log tab should be highlighted on parent pages like https://so-job-tabs-remove-jquery.test-deployments.travis-ci.com/travis-ci/travis-web


sorta:
I agree, it's very exciting :)

PR #2248, opened 2019-11-21
Remove jQuery used to set body class in routes

sorta:
Split from https://github.com/travis-ci/travis-web/pull/2203. A lot of these seem unused and unnecessary. The one that does seem to be used is in the `auth` route.

### DIFF
--- a/app/routes/application.js
+++ b/app/routes/application.js
@@ -1,5 +1,4 @@
 /* global Travis */
-import $ from 'jquery';
 import URL from 'url';
 import TravisRoute from 'travis/routes/basic';
 import config from 'travis/config/environment';
@@ -40,13 +39,6 @@ export default TravisRoute.extend(BuildFaviconMixin, {
       });
     }
 
-    return this._super(...arguments);
-  },
-
-  renderTemplate: function () {
-    if (this.get('features.proVersion')) {
-      $('body').addClass('pro');
-    }
     return this._super(...arguments);
   },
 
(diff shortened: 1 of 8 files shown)


sorta:
Some thoughts about the `/auth` route:

I initially replaced the jQuery there with a `document.querySelector` version that basically did the same thing. But I started to feel that it's unnecessary, so I removed it.

The `auth` ID seemed to be used for 2 styling purposes - to hide the top bar signin button on `/auth` page, and to hide the 'Help make Open Source a better place' cta on `/auth` page.

The cta is already being hidden on the pro version of `/auth` through other means, and I think it looks fine on the `.org` version of the page.

Likewise, I think the top bar signin button looks fine on the `/auth` page. And it may even be good to have there, since the `signin` template that the `/auth` page uses just has the github button, and is not really set up with the multi-vcs logic that's in the top bar.

I didn't want to mess around with the multi-vcs stuff in this PR, but it may be good in the future to make a component to encapsulate a multi-vcs signin button.

And even if we do end up wanting to hide those buttons on `/auth`, I think there's a better way to accomplish it than setting an ID attribute on the `body` element and then hiding with css.

Let me know if that all makes sense to you, and what you think :)

artursmirnov:
Generally, this is such a great improvement 😍These attributes bothered me for so long! Thanks for cleaning them up 👍

sorta:
Thanks for the reviews! Yeah, I was thinking along similar lines about the top bar signin button.

PR #2249, opened 2019-11-21
Remove jquery from repo-list-item component

sorta:
Split from https://github.com/travis-ci/travis-web/pull/2203

I changed to using `window.pageYOffset` because `window.scrollY` is not supported by IE. Some info: 
https://developer.mozilla.org/en-US/docs/Web/API/Window/scrollY
https://developer.mozilla.org/en-US/docs/Web/API/Window/pageYOffset

### DIFF
--- a/app/components/repos-list-item.js
+++ b/app/components/repos-list-item.js
@@ -2,22 +2,13 @@ import Component from '@ember/component';
 import Polling from 'travis/mixins/polling';
 import colorForState from 'travis/utils/color-for-state';
 import { computed } from '@ember/object';
-import { inject as service } from '@ember/service';
 
 export default Component.extend(Polling, {
   tagName: 'li',
   pollModels: 'repo',
   classNames: ['repo'],
 
-  scroller: service(),
-
   color: computed('repo.currentBuild.state', function () {
     return colorForState(this.get('repo.currentBuild.state'));
   }),
-
-  scrollTop() {
-    if (window && window.pageYOffset) {
-      this.scroller.scrollTo(0, 200);
-    }
-  }
 });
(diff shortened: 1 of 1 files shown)


app/components/repos-list-item.js
```diff
@@ -1,23 +1,23 @@
...
     return colorForState(this.get('repo.currentBuild.state'));
   }),
 
   scrollTop() {
-    if (window.scrollY > 0) {
-      return $('html, body').animate({
-        scrollTop: 0
-      }, 200);
+    if (window && window.pageYOffset) {
+      this.scroller.scrollTo(0, 200);
```
artursmirnov: As far as I understand, `scrollTo` [receives 3 arguments](https://github.com/peopleconnectus/ember-concurrency-scroll#scrollerscrolltostart-end-options-function) - `start`, `end` and `options`. With this call it seems that `start = 0` and `end = 200`, which means that the service will scroll from 0 to 200 - is it correct? Because the old code seems to be scrolling to 0 from whatever scroll value was. I didn't test it, so it's mostly my assumption, but I wanted to raise it to your attention.

sorta: Hmm good point, I think I may have misinterpreted it at some point. I'll look into it

sorta: So yes, I think you're right and this should be `this.scroller.scrollTo(window.pageYOffset, 0)`

But I also noticed, that this function isn't ever used in this component or its template. It seems like it used to be done on clicking the component, but it was disconnected some time ago. So, it seems like it's actually safe to remove altogether

artursmirnov: Nice! :)

sorta:
Awesome :smile: Thanks for the review @artursmirnov! :tada:

PR #2253, opened 2019-11-25
Fix ui migration confusion

clekstro:
While we removed the "wait" from the waitlist for the migration beta some time ago, I neglected when rolling out that change to remove all references to a waitlist within the UI. As users are immediately accepted, I didn't think much of it at the time. Unfortunately, this has resulted in some confusion, so needs to be addressed.

### DIFF
--- a/app/components/dialogs/migrate-beta.js
+++ b/app/components/dialogs/migrate-beta.js
@@ -27,7 +27,7 @@ export default Component.extend({
       yield this.user.joinMigrateBeta(this.selectedAccounts.without(this.user).toArray());
       this.onClose();
       this.flashes.clear();
-      this.flashes.success('You have been successfully placed on the waitlist for beta!');
+      this.flashes.success('You have successfully joined the beta!');
     } catch (error) {
       this.flashes.clear();
       this.flashes.error('There was an error. Please try again later.', undefined, true);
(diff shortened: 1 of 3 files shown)


app/components/dialogs/migrate-beta.js
```diff
@@ -66,9 +66,9 @@ export default Component.extend({
 });
 
 function makeOptionFromAccount(account) {
-  const { id, title, isMigrationBetaAccepted, isMigrationBetaRequested, isOrganization, isUser } = account;
+  const { id, title, isMigrationBetaAccepted, isOrganization, isUser } = account;
   const isNotAdmin = isOrganization && !account.get('permissions.admin');
-  const state = isMigrationBetaAccepted ? 'subscribed' : isMigrationBetaRequested ? 'waitlisted' : isNotAdmin ? 'not admin' : '';
-  const disabled = isOrganization && (isMigrationBetaAccepted || isMigrationBetaRequested) || isUser || isNotAdmin;
+  const state = isMigrationBetaAccepted ? 'subscribed' : isNotAdmin ? 'not admin' : '';
+  const disabled = isOrganization && (isMigrationBetaAccepted) || isUser || isNotAdmin;
```
artursmirnov: Just a tiny visual tweak 😉 
```suggestion
  const disabled = isOrganization && isMigrationBetaAccepted || isUser || isNotAdmin;
```

PR #2259, opened 2019-11-27
Replace ember-modal-dialog

sorta:
Ember-modal-dialog is dependent on jQuery. We're considering whether it might make sense to replace it with our own component.

It seems that it and liquid-fire are the only remaining dependencies that rely on jquery. I tried disabling jquery integration on this branch and it broke one test (which I looked into and I think resolved), but things generally seem to work. We can re-enable the jquery integration if we want to disable it on a separate PR.

Modals to watch out for:
- Insights visibility settings
https://so-replace-modal-dialog.test-deployments.travis-ci.com/account/preferences
- Cancel Subscription
https://so-replace-modal-dialog.test-deployments.travis-ci.com/account/subscription
https://so-replace-modal-dialog.test-deployments.travis-ci.com/organizations/travis-pro/subscription
- Migrate Beta
https://so-replace-modal-dialog.test-deployments.travis-ci.com/organizations/travis-pro/migrate
- Repository migration (when you have a repo that can be migrated)
https://so-replace-modal-dialog.test-deployments.travis-ci.com/account/migrate
- Log Removal
https://so-replace-modal-dialog.test-deployments.travis-ci.com/travis-ci/travis-web/builds/140270785
- Build Status Badge, same as above
- Trigger Build, same as above
- Migrated Settings
https://so-replace-modal-dialog.test-deployments.travis-ci.org/travis-ci/travis-web/settings


mirage/config.js
```diff
@@ -64,8 +64,15 @@ export default function () {
...
-    return this.serialize(schema.users.first(), 'v3');
+  this.get('/user', function (schema, request) {
+    const { authorization } = request.requestHeaders;
+    const firstUser = schema.users.first();
+
+    if (authorization !== `token ${firstUser.token}`) {
+      return new Response(403, {}, {});
+    }
+
+    return this.serialize(firstUser, 'v3');
```
sorta: When I turned off the jquery integration, it broke the automatic signout test. In the test, the token expired flash message was no longer showing up, because the `fetchCurrentUser` task in the `auth` service was not getting an error from `currentUser.reload`. I tried on production changing my token in localStorage and reloading the page while logged in, and noticed that `/user` returns with a `403`, which seems to allow the auto signout to work.

So I made this update to mirage config, which returns 403 if token doesn't match. I'm not really sure why this test works at all when jquery integration is enabled?

artursmirnov: Nice, @sorta thanks for catching this 👍

app/templates/components/dialogs/cancel-subscription-modal.hbs
```diff
@@ -1,21 +1,11 @@
-{{#if @isOpen}}
-  <ModalDialog
-    @targetAttachment="center"
-    @translucentOverlay={{true}}
-    @overlayPosition="sibling"
-    @clickOutsideToClose={{true}}
-    @animatable={{true}}
+
```
sorta: I thought it might be nice if Modal showed and hid itself, to reduce the logic necessary to use it. It also made it a little easier to implement the animation. As a result right now a few of these templates are not properly indented, or have a superfluous div to make it more obvious what actually changed, to make at least the initial review a little easier. I would like to properly indent these before merging (or if people feel it would be better to keep outside-if logic, revert to that)

app/services/animation.js
```diff
@@ -0,0 +1,34 @@
+import Service from '@ember/service';
+import { computed } from '@ember/object';
+import fade from 'ember-animated/transitions/fade';
+import config from 'travis/config/environment';
+
+const { environment } = config;
+const isTest = environment === 'test';
```
artursmirnov: Another way of checking for testing is `if (Ember.testing) ...`, though, I'm not sure if it's recommended, just shorter 😄Current implementation works perfectly though, comment is mostly for information ;)

app/templates/components/dialogs/cancel-subscription-modal.hbs
artursmirnov: This is a great idea, @sorta, thank you for your care, it's so much appreciated ❤️

app/services/animation.js
sorta: Ahh nice! I think I like that better, so I switched to it. Thanks for letting me know!

sorta:
Thanks for the reviews @artursmirnov & @piotr-travisci !

PR #2266, opened 2019-12-03
Fix issue where migration requests request would get a 500 error

sorta:
I was doing some testing on staging and noticed a 500 error on beta_migration_requests endpoint when going to /account/subscription. It seems it was due to the `travis-api-version` header being set to 3


sorta:
Excellent @artursmirnov, thanks for the review! :smile:

PR #2267, opened 2019-12-04
Add activate all github button to getting_started page

sorta:
For https://travisci.assembla.com/spaces/Web/tickets/250/details

I started with just copying the existing logic from a similar button into a new component that can be used wherever we need this thing. And then I started thinking, maybe it would be better to make this more vcs-agnostic? Or is this so specific to github that there isn't really a point to it? Thoughts and suggestions?

Thanks!


app/templates/components/button-activate-all-github.hbs
```diff
@@ -0,0 +1,10 @@
+{{#if this.isVisible}}
+  <button
+    class="button--blue"
+    data-test-activate-button
+    {{on 'click' (action 'activateAll')}}
+  >
+    <SvgImage @name={{vcs-icon this.owner.vcsType}} @class="fill-current align-text-bottom w-5 h-5 ml-1 mr-2" />
```
artursmirnov: I think this is very GitHub specific logic and won't work with other VCSes, so the icon image could be just github.

sorta: Yeah, that makes sense to me. Change made :)

sorta:
Thanks for the review @artursmirnov !

sorta:
Excellent, thanks for confirming and for the review @piotr-travisci !

PR #2269, opened 2019-12-04
Move client release header to api service

sorta:
This should address an issue where the help page zendesk form is erroring. The X-Client-Release header, I believe should only be relevant to our API, and very few things use the ajax service directly, so I don't think this change should cause problems


artursmirnov:
I've tested it in the test deployment and it works perfectly. I'm going to merge and release this fix as it's pretty crucial for customers. @sorta thank you for fixing the issue! 🎉

PR #2271, opened 2019-12-05
Fix success image size on help page

artursmirnov:
Issue description:
When user sends a form to Zendesk the success image scales to the page width and looks giant.


artursmirnov:
Thanks for the review @piotr-travisci! 🎉

sorta:
Much better, thanks for fixing this!

PR #2272, opened 2019-12-05
Running (jobs) tab refactor

piotr-travisci:


app/services/pusher.js
```diff
@@ -28,6 +29,11 @@ export default Service.extend({
       store.push(store.normalize('commit', data.job.commit));
     }
 
+    if (name === 'job' && data.job) {
+      store.push(store.normalize('job', data.job));
+      this.jobState.reloadJobs.perform();
+    }
```
artursmirnov: Ah, I see the purpose for the `store.peekAll` now :) Though, I'd suggest the following approach:
```js
const newJob = store.push(store.normalize('job', data.job));
this.jobState.jobs.addObject(newJob);
```
It should do the trick.

Another (and probably more reliable and solid) approach would be to re-fetch jobs from server by `this.jobState.fetchJobs.perform();` here, but I'm not sure about the API load.

piotr-travisci:
@artursmirnov Thank you very much for such a constructive review + chat.
I implemented all your suggestions and I think it's ready to review again.

PR #2276, opened 2019-12-09
FilteredArrayManager refactored to prevent duplicated query

piotr-travisci:


app/utils/filtered-array-manager.js
```diff
@@ -183,8 +183,11 @@ let FilteredArrayManager = EmberObject.extend({
     const filterArray = this.filteredArrayManagerForType(modelName).getFilterArray(queryParams, filterFunction, dependencies);
 
     if (queryParams) {
-      const queryPromise = this.store.query(modelName, queryParams);
-      const promise = resolve(queryPromise).then(() => filterArray);
+      let currentRecords = this.store.peekAll(modelName);
+      if (filterFunction) {
+        currentRecords = currentRecords.filter(record => filterFunction(record));
+      }
+      const promise = resolve(currentRecords).then(() => filterArray);
```
artursmirnov: I'm not really sure about this change. I didn't try it live, but logically it does completely different things. My concerns are:
1) It doesn't use `queryParams` even though it only runs when they exist. And previous implementation did use them
2) It never runs requests to API, always picking records from the store, not sure if it's expected. I guess, at some point it should fetch data from API, shouldn't it?
3) `filterFunction` might already be applied by `filteredArrayManagerForType`, not sure if it needs to be applied manually

Also, as a side note, I think it's safe to shorten `filterFunction` call to just `currentRecords.filter(filterFunction)`

artursmirnov: It has been discussed privately, all concerns are resolved. Looks good to me 👍

piotr-travisci: For wider visibility of implemented solution:
`const filterArray = this.filteredArrayManagerForType(modelName).getFilterArray(queryParams, filterFunction, dependencies);` code already make API call via `store.query`.
In the meantime while we are waiting for data from backend we try to show what we already have in store
`let currentRecords = this.store.peekAll(modelName);`

piotr-travisci: and many thanks for your review @artursmirnov !

PR #2282, opened 2019-12-12
Call logout at API service when signing out

piotr-travisci:


app/routes/application.js
```diff
@@ -129,6 +130,7 @@ export default TravisRoute.extend(BuildFaviconMixin, {
 
     signOut() {
       this.auth.signOut();
+      this.api.get('/logout');
```
artursmirnov: What do you think about moving this to [`auth.signOut`](https://github.com/travis-ci/travis-web/blob/master/app/services/auth.js#L73) method? It handles the sign out process and has all the cleanups, so probably I'd expect this request to be there as well. How does it sound to you?

piotr-travisci: Thank you Artur for your suggestion. As always your ideas are great.
I refactored it.

PR #2285, opened 2019-12-13
Running jobs filter reinvented

piotr-travisci:


app/components/repository-sidebar.js
```diff
@@ -82,7 +82,7 @@ export default Component.extend({
       let showRunningJobs = this.get('features.showRunningJobsInSidebar');
       let runningJobs = this.get('jobState.runningJobs');
       if (!showRunningJobs) { return []; }
-      return runningJobs;
+      return runningJobs.sort((a, b) => parseFloat(a.number) - parseFloat(b.number));
```
artursmirnov: The suggestion that I mentioned on Slack was to use [`sortBy`](https://api.emberjs.com/ember/3.12/classes/EmberArray/methods/sortBy?anchor=sortBy) here:
```js
return runningJobs.sortBy('number');
```
Though, it's very minor and completely optional, more for awareness ;)

PR #2286, opened 2019-12-13
Ensure feature flags are removed on signout

clekstro:
Feature flags are cached for performance reasons in local storage. Currently, when you log out, these are not cleared. If you're logged in in another browser window, these flags will never be updated, as we assume they haven't changed, so you'll have to repeat the action.

With this PR, this data is cleared, meaning that when you make an update in one browser window and log out, you'll get the desired settings upon your next login.

### DIFF
--- a/app/services/auth.js
+++ b/app/services/auth.js
@@ -71,8 +71,10 @@ export default Service.extend({
   redirectUrl: null,
 
   signOut(runTeardown = true) {
-    this.localStorage.clearAuthData();
-    this.sessionStorage.clearAuthData();
+    [this.localStorage, this.sessionStorage].map(storage => {
+      storage.clearAuthData();
+      storage.clearPreferencesData();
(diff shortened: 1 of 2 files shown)


app/services/auth.js
```diff
@@ -71,8 +71,10 @@ export default Service.extend({
   redirectUrl: null,
 
   signOut(runTeardown = true) {
-    this.localStorage.clearAuthData();
-    this.sessionStorage.clearAuthData();
+    [this.localStorage, this.sessionStorage].map(storage => {
```
artursmirnov: I'd suggest to use `forEach` here, just to make it explicit that we're operating on the array items, not transforming them.
```suggestion
    [this.localStorage, this.sessionStorage].forEach(storage => {
```
Though it's just an opinion and completely optional, `map` works perfectly as well.

PR #2289, opened 2019-12-16
Display complete (one line / 72 chars) commit message

mvorisek:
Currently the limit is 50 chars, but the standard commit message limit is 72 chars (one line, this limit is also enforced by most git GUI clients).

image

This PR raises the limit to 72 chars to display the complete commit message for better UX as the most important details are often at the end of the commit message and the complete first line of the commit message needs to be displayed everywhere.


artursmirnov:
Hey @mvorisek, thanks a lot for your contribution! It totally makes sense. We'll check it on our test deployments to verify that the longer message fits well on narrows screens as well.

app/templates/components/builds-item.hbs
```diff
@@ -46,7 +46,7 @@
           short="true"
           repo=this.build.repo
           eventType=this.build.eventType
-          maxLength=50
+          maxLength=72
```
native-api: * See https://travis-ci.community/t/display-complete-commit-title-in-travis-builds/6481/2 for why if there is a limit, existing 50 is the right limit.

mvorisek: @native-api The article mention best practises for developers. The article also mentions that the average length is 50 chars, so should half of the commit messaged be cropped?

native-api: I may be reading a different article but for me, the "“50 / 72” rule" that it's dedicated to reads:

> 1. The first line of your commit message must be maximum 50 characters long. No more, and (ideally), no less.

And this "first line" is exactly what is printed in `row-message`.

native-api: Yes, the article discusses these limits later in order to justify them. But the resulting rule it proposes is this.

native-api: In any case, note the _"if there is a limit"_ part. In the general comment, I proposed to abolish the limit altogether.

mvorisek: @native-api

> The first line of your commit message must be maximum 50 characters long. No more, and (ideally), no less.

This is an advice for developer, but as you can see in the histogram, 50 chars are not sufficient as it crops 50% of the commit messages.

For me this PR is a quick fix and further improvement can be done in another PR or if you want to extend this directly...

native-api: Oh, so you are not actually following that rule. Okay then.

PR #2292, opened 2019-12-17
Npm audit fix

native-api:
While exploring possibilities for https://travis-ci.community/t/display-complete-commit-title-in-travis-builds/6481 without much fruit and seeing the OP propose their PR anyway instead, thought I might as well fix an obvious and straightforward thing.

See commit messages for details.


artursmirnov:
Hey @native-api, thank you for your contribution! 

Unfortunately, we cannot accept it at the moment, because the fix bumps up the `ember-cli` dependency to `3.14`. In Ember apps it's recommended to keep all framework parts at the same version, otherwise it may break some things unexpectedly. So the update of `ember-cli` would also require updating the `ember-source` and `ember-data` at least. Because of that we usually strive to update dependencies in a batch. 

Of course, we understand the danger of having known vulnerabilities in the codebase and toolset (and thanks again for willing to help here!), but since they all are build-time ones and of a low severity, I think it's pretty safe to keep it as is before our next dependencies update.

PR #2297, opened 2019-12-18
UI Kit remove mixins

sorta:
I ran into this eslint rule in a personal project, and I think the reasons for it are sound https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/no-new-mixins.md. I have felt sometimes that mixins can make things awkward. And since there are currently only 2 mixins in the UI Kit, I thought it might be a good time to remove them. It was pretty easy in this case, and the sharing of spacing and border logic is now achieved through composability with the `box` component.

The increased use of composability exposed an issue in the `getResponsiveProp` util function, which this PR also addresses. And I also ended up slightly simplifying the `link` component template.


artursmirnov:
Yeah, mixins are basically deprecated, Octane doesn't support them, as they don't work with native ES5 classes. So thanks for taking care of this 👍

PR #2298, opened 2019-12-18
Fix signout request

artursmirnov:
It seems like currently when app loads for unauthenticated user, `signOut` is getting called, which causes `403` error on API call:

image

Steps to reproduce:
- Log out from the app
- Open Network tab in browser DevTools
- Reload page
- See 403 error on /logout endpoint


artursmirnov:
Thanks for the review, @piotr-travisci 🎉

PR #2299, opened 2019-12-18
Fix intermittent failures of insights settings modal test

sorta:
I think it may be due to timing issues, when the modal shows and hides. I did see it intermittently for a bit, but I made some changes and haven't seen it since then. However, the ember data team reached out because it is failing in their partner test suite.

Example failed run https://github.com/emberjs/data/pull/6911/checks?check_run_id=354030973#step:6:2287


sorta:
Thanks for the review @artursmirnov !

PR #2310, opened 2020-01-06
New sign up/in page

sorta:
For https://travisci.assembla.com/spaces/Web/tickets/256/details

https://so-bb-sign-up-in-production.test-deployments.travis-ci.com/
https://so-bb-sign-up-in-production.test-deployments.travis-ci.com/signin
https://so-bb-sign-up-in-production.test-deployments.travis-ci.com/signup

https://so-bb-sign-up-in-production.test-deployments.travis-ci.org/
https://so-bb-sign-up-in-production.test-deployments.travis-ci.org/signin
https://so-bb-sign-up-in-production.test-deployments.travis-ci.org/signup

As demonstrated by Artur Kremens, you can enable bitbucket login with
`window.localStorage['enableBitbucketLogin'] = 'true'`
In browser devtools console.

Since multi vcs is not planned to be used on .org, I left .org landing page alone as much as possible. For the new signup page and redesigned signin page, since they are using styleguides styles, I used the UI Kit to implement them, but tried minimize updates to the UI Kit. Currently it does not remember the last VCS used, I think to do so properly would require updates to the api handshake, and I wanted to get reviews started on this as soon as possible.

The `auth` and `signin` pages now use separate templates, as otherwise changes to the signin page led to some interesting issues. I copied the previous signin template content over to the new auth template so that it could stay the same. Although, now that I think about it, maybe the auth route should be updated too? It wouldn't be hard to make it the exact same as the new signin page, using a component.

https://so-bb-sign-up-in-production.test-deployments.travis-ci.com/auth?redirectUri=http%3A%2F%2Flocalhost%3A4200%2Faccount%2Fsubscription

I also removed the auto-signin logic from the signin route, since it doesn't make sense for multi vcs. Although, maybe it would be good to reinstate it when multi vcs is disabled?


AlonaHlobina:
Thank you @sorta! Great job! It looks fantastic! I am so excited about it!!! Only a few minor things I noticed here:
- `Testing your open source projects is always 10000% free!` - I believe it is a typo in design and `100%` is enough for us :-)
- Under the `Pull request build flow` section, we have `GitHub tells Travis CI the build is mergeable.` Let's change `GitHub` to `cloud platform.`
-  The button `Start building now` at the bottom of the page directly log me into my GitHub account. I am not sure what kind of behavior is expected for it. I assumed it should lead me to the sing up/in step first? @sorta, what do you think? 
Regarding `removed the auto-signin logic from the signin route, since it doesn't make sense for multi vcs` - I think it is a good idea to remove it for now. @artursmirnov What do you think about it?

sorta:
Hi @AlonaHlobina, thanks for the feedback! :smile: I've addressed the things you've mentioned as follows:

- Hah, it did seem perhaps a little excessive. Changed to `100%`
- Changed to `Cloud platform`
- Yeah, I think it would make sense to have it go to one of those pages, maybe the signup page. Updated the button to do so.

artursmirnov:
@AlonaHlobina @sorta removing `auto-signin` totally makes sense to me 👍

artursmirnov:
@sorta I have tested it for a bit - it looks really great, thanks for your work on it 👍But there still are couple of things that caught my eye:
- Bitbucket sign in/up is available on `.org` as well (with feature flag enabled), but fails with a misconfiguration error on Bitbucket side (I guess, that's expected)
- I have profiles on both GitHub and Bitbucket, but when I press `Sign in with Bitbucket` button, it seems to redirect me correctly, but then authorizes me with my GitHub account. 
- I'm curious, why is GitHub green? :) It seems to be black - its logo, heading and so on.

sorta:
Thanks for the testing & feedback @artursmirnov. I looked into those things:

- Hmm yeah, maybe it would be better to disable altogether on `.org`. I've made an update to that effect
- Initially I had thought this was because some backend stuff is still in progress, but I looked into it a bit more. I believe it is because multi VCS login is disabled on the production api. There is this check in the handshake `method = Travis::Features.enabled_for_all?(:vcs_login) ? :vcs_handshake : :handshake`. It appears to be enabled on the staging api though. This gives us a few options for testing it without necessarily enabling on the production api.
- This is a good question, it does seem a little weird to me. I would be interested to hear what @AlonaHlobina or @Nat-ess think about it

artursmirnov:
@sorta you're right, the authorization issue was caused by test-deployments communicating with production API. I tried on staging and it works fine, though seems like your changes are not there yet (but that's okay, since your changes doesn't really require API, they are mostly visual). So thanks for explaining! 😄

sorta:
@artursmirnov absolutely, yw :smile: I thought about deploying to staging, but it seems that artur kremens is currently using it to test other vcs changes. So yeah, since this update is mostly visual, it's fine for now, though I definitely prefer to have the whole thing working together.

With more people working on frontend vcs stuff now though, relying only on staging environment may become more of a bottleneck. I wonder if we should consider enabling it on production api?

Nat-ess:
Thanks for this @sorta, below are some minor comments. It looks good!

- Change Sign-in/up green link at the bottom to `#32d282` on sign up/in pages (I think all new pages should use our new styles, regardless if the landing page still does not)
- The blue blob at in the background seems off in colour, maybe this is an svg thing, but the colour should be `#0068ff`
- On the landing page, the button on the top right says "Sign in with GitHub" should we change it to say "Sign in" or remove it all together?
- Is there any way, to have the Branch and PR build flows copy on two lines? Maybe Removing "You" on the cloud step? Could "goodness" on the PR build flow move up?

> I'm curious, why is GitHub green? :) It seems to be black - its logo, heading and so on.

I did this to match our new colour styles, but now im wondering if it would be an issue with github?

artursmirnov:
@Nat-ess thanks for the response! I thought it's the case, my concern though is that all our users use GitHub as well and this color may be confusing for some of them. 

> im wondering if it would be an issue with github?

There's a policy around logos: https://github.com/logos which recommends to avoid color changes, though I'm not sure if our case is a violation 🤔

Nat-ess:
Yeah, we've changed the colour of the logo in a few occasions and it's been ok. Maybe we could do their icon in grey (so it matches on other parts of our UI), like this:
Screen Shot 2020-01-10 at 11 23 52 AM

artursmirnov:
@Nat-ess what do you think about making the entire button (along with the logo) black? E.g. black icon, black border, black text. I understand that it may conflict with our UI color scheme, but that just could be more intuitive and habitually for our users. Everywhere I saw GitHub login buttons they were of black color, here are a few examples:
- [stackoverflow](https://stackoverflow.com/users/login?ssrc=head&returnurl=https%3a%2f%2fstackoverflow.com%2f)
- [percy](https://percy.io/login)
- [zeit](https://zeit.co/login)

AlonaHlobina:
@sorta @artursmirnov Is there any way we can make this design changes work together with full sign in Bitbucket process and the rest of UI changes on which Piotr is working now? Since that is what the Marketing team will be testing on Monday

Nat-ess:
Thanks for the examples @artursmirnov, we could do something like this:

Screen Shot 2020-01-10 at 11 48 41 AM

artursmirnov:
@Nat-ess thanks! It looks great to me 👍The green version was definitely more bright and beautiful, and aligned better with the new styles, but I think this way it would be more convenient and intuitive for users.

Nat-ess:
Yeah, no worries! It can be tricky, but also it's important to make sure our UI is intuitive, like you mentioned and users recognize what we're trying to reference/do.

sorta:
> Thanks for this @sorta, below are some minor comments. It looks good!
> 
> * Change Sign-in/up green link at the bottom to `#32d282` on sign up/in pages (I think all new pages should use our new styles, regardless if the landing page still does not)
> * The blue blob at in the background seems off in colour, maybe this is an svg thing, but the colour should be `#0068ff`
> * On the landing page, the button on the top right says "Sign in with GitHub" should we change it to say "Sign in" or remove it all together?
> * Is there any way, to have the Branch and PR build flows copy on two lines? Maybe Removing "You" on the cloud step? Could "goodness" on the PR build flow move up?

Thanks @Nat-ess for the great feedback! :smile: I've addressed as follows:

- Altered the link color
- Yeah, the color stops in the svg looked a bit purple. I've tried to adjust it, let me know what you think. Here's what the current ones are
blob-color-stops
- I removed github from the text in the button. I'd like to avoid removing this button for now, but if it proves to be a problem yeah we should consider removing it
- It's a bit tricky, the current implementation for these is very sensitive to any style changes, so yeah the easiest way would be to alter the wording. I got them all to 2 lines by setting branch step 1 to "Push your code to cloud platform" and PR step 2 to "Travis CI is notified the build is mergeable". But those both sound awkward to me, so idk. I've considered a larger refactoring of these 2 diagrams, but I don't know if we have time for it. Not sure about moving "goodness" up, that one is showing on 2 lines for me, is it showing up differently for you?

Nat-ess:
> Yeah, the color stops in the svg looked a bit purple. I've tried to adjust it, let me know what you think. Here's what the current ones are

Looks good to me!

>I removed github from the text in the button. I'd like to avoid removing this button for now, but if it proves to be a problem yeah we should consider removing it

Im more in favour of keeping it, but hope it doesn't become a problem.

>It's a bit tricky, the current implementation for these is very sensitive to any style changes, so yeah the easiest way would be to alter the wording. I got them all to 2 lines by setting branch step 1 to "Push your code to cloud platform" and PR step 2 to "Travis CI is notified the build is mergeable". But those both sound awkward to me, so idk. I've considered a larger refactoring of these 2 diagrams, but I don't know if we have time for it. Not sure about moving "goodness" up, that one is showing on 2 lines for me, is it showing up differently for you?

Here's what it looks like for me:
Screen Shot 2020-01-13 at 3 08 37 PM

sorta:
Ahh I see! Ok I made some more changes. I think I got the PR goodness one sorted out, and did my best to get the others to fit on 2 lines as well. When you get a chance, please let me know what you think @Nat-ess. Thanks!

Nat-ess:
I think we should keep the extra "your" in the first diagram copy. I think it's ok if it goes on three lines, i'd rather have the copy be clear over a visual change.

The "goodness" is still on two lines for me, if it's too tricky, let's keep it as is.
Screen Shot 2020-01-13 at 4 00 21 PM

sorta:
Huh, that's kind of bizarre. Well I made one more change, let me know if it helps. Thanks!

Nat-ess:
It works!

artursmirnov:
I've changed base branch for this PR to `epic-bb-integration`, so that we could review and merge all related PRs before release, test everything properly and then release everything at once.

sorta:
Good idea @artursmirnov, thanks!

app/routes/signup.js
```diff
@@ -0,0 +1,25 @@
...
+
+  activate: function () {
+    this.set('headData.useTailwindBase', true);
+    return this._super(...arguments);
+  },
+
+  deactivate() {
+    this.set('headData.useTailwindBase', false);
+    return this._super(...arguments);
+  },
```
artursmirnov: It feels like these methods worth a mixin, something like `withTailwind`, what do you think?

sorta: That sounds good to me! :smile: And actually, I've remembered that there already is a mixin for this, so I've updated to use it

sorta:
Thanks for the review!

PR #2312, opened 2020-01-07
UI Bitbucket related changes

piotr-travisci:


app/templates/components/github-apps-repository.hbs
```diff
@@ -8,6 +8,12 @@
   {{this.name}}
 </LinkTo>
 {{#if this.admin}}
+  {{#if this.isNotMatchGithub}}
+    <TravisSwitch
+      @active={{this.repository.active}}
+      @onToggle={{perform this.toggleRepositoryTask}}
+    />
+  {{/if}}
```
sorta: I'm not sure when this would come up, it seems like this component is only used when `isGithubRepository` is true. But maybe there's a case I'm forgetting about, or this is precautionary. Either way, I think this is fine, just wanted to note and ask about it :smile:

sorta:
Although I just noticed in the percy diffs that the icons on the dashboard look a little bit low. Maybe adjusting the `vertical-align` value on them might help

app/templates/components/github-apps-repository.hbs
piotr-travisci: Actually github-apps-repository component is used for all repositories, not only GH ones. You are right this is a little confusing name. We should refactor it (at least rename) in future.

piotr-travisci:
Nice finding about vertical align of vcs icon.
Thank you @sorta for review!

artursmirnov:
I've changed base branch for this PR to `epic-bb-integration`, so that we could review and merge all related PRs before release, test everything properly and then release everything at once.

PR #2324, opened 2020-01-21
Add dev flags

sorta:
While working on the jenkins page and vcs integration, I thought a bit about the current usage of feature flags.  I was considering to switch the jenkins page work from using an epic branch process to a feature flag process. But, right now there doesn't seem to be a firmly established, easy workflow for using feature flags to show and hide work in progress.

The `feature-flags` service seems meant for user-facing 'beta features' which a user can toggle in their settings, and it is persisted to DB. It's a little complex and doesn't help if we need to flag something visible when logged out, so I think it's not really suitable for this use.

The bitbucket and assembla flags are much closer, but use a lot of custom code for each flag. Additionally, with a localStorage based approach, an astute user could read the code and enable unfinished features for themselves just by setting something in localStorage. Maybe unlikely or not a big problem now, but if we start to do this more I could see it possibly creating an issue.

I think, ideally a dev/WIP flags process should:

- Require minimal code changes to set up a new flag
- Make it easy to flip flags on and off in development and test-deployment environments as needed
- Make it easy to transition from "dev" flag to a regular feature flag, if needed
- Avoid unauthorized use of the unfinished feature

What I have so far is an environment variable with which you can pass a comma separated list of flags that you would like to enable. This would allow you to enable flags in development using `.env` file, on test deployments using travis settings, and also on staging via heroku, if needed.

Here's what I'm thinking the process might be like:

1. Think about whether we may want to be able to disable the new feature after it is released, keeping in mind things like Enterprise.
   - If yes, you can add `'new-thing': false,` to `ENV.featureFlags` in `environment.js`, and then add `new-thing` to ENABLE_FEATURE_FLAGS var in `.env` file
   - If no, you can just add `dev-new-thing` to the ENABLE_FEATURE_FLAGS var in `.env` file
2. Add any checks needed to hide work in progress, using the flag name you set up and the `features` service
3. Set ENABLE_FEATURE_FLAGS env var as needed in test deployment environments, using travis settings. Or on staging through heroku.
4. Once feature is ready to release, either remove the checks for `dev-new-thing`, or set `'new-thing': true,` in `environment.js`, and feel free to remove the flag from ENABLE_FEATURE_FLAGS

The part that can be a little awkward with this approach is enabling on test deployments. It's easy to target all test deployments by setting ENABLE_FEATURE_FLAGS on all branches in travis settings. But if you need to target only a single PR, it's not so simple.

Let's say you have a PR to merge branch `so-new-thing` into `master`. If you set an env var `ENABLE_FEATURE_FLAGS=dev-new-thing` targeting the `so-new-thing` branch, the test deployment will not receive that env var. This is because in PR builds, the relevant branch name is the one being merged into (in this case, master).

A workaround could be to set the PR to merge into a branch other than master (perhaps something like `demo-new-thing`). It's awkward, but seems to work.

Perhaps in many cases, targeting all test deployments would be sufficient. And I guess there's always the option to deploy to staging. But yeah, it would be ideal to be able to easily target individual test deployments.

I would love to hear everyone's thoughts, ideas, and suggestions. Maybe there is another, better approach.


sorta:
A bit of extra background:

In an earlier version of this update, the flags were forced to be namespaced with a `dev-` prefix, and there was a `dev-flags` service to load flag values from localStorage, as well as some convenience function for stuff like `enable`/`disable`/`isEnabled`. I felt this had a few potential issues:

- An astute user could read the code and enable unfinished features for themselves just by setting something in localStorage. Maybe unlikely or not a big problem now, but if we start to do this more I could see it possibly creating an issue.
- The namespacing is maybe handy for development, but a bit awkward if you wanted to keep that flag after the feature was finished and widely enabled, as you would need to rename from `dev-new-thing` to `new-thing`.
- Additionally, if you used the `dev-flags` service convenience functions during development, you would need to switch to the regular `features` service. So removing the service is simpler, less confusing, and more DRY.

So the current version just has non-namespaced flags, that can be enabled via env var. It is a little bit more work to flip flags on and off, but I feel it is overall better.

sorta:
Thanks for the review and feedback @artursmirnov!

Yes, if we could easily target individual test deployments, that would be very nice and help make the process easier :) I worry that doing test deployments on `Branch` build may result in many more test deployments, possibly unnecessarily. But maybe not really?

That's a good point about env vars in Travis UI. I did have a thought that it might eventually become awkward if we end up with a lot of flags to enable this way, but at the time I didn't see a way around it that didn't negatively impact some other concern. But thinking about it again, I'm starting to have an idea about a possible way to address it.

sorta:
I added some logic to check env vars for individual flags that are set in `environment.js`. So if you have flag `new-thing`, it will check env var `FLAG_NEW_THING`, and if the value is 'true' it will enable the flag, and if 'false' it will disable the flag. This should make it easier to use Travis UI to turn on and off individual flags. And ENABLE_FEATURE_FLAGS remains for easy enabling via `.env` file during development.

PR #2293, opened 2019-12-17
Travis vs jenkins page

sorta:
For https://travisci.assembla.com/spaces/Web/tickets/243/details

Test URLs:
- https://epic-tvj-add-page-production.test-deployments.travis-ci.com/travisci-vs-jenkins
- https://epic-tvj-add-page-production.test-deployments.travis-ci.org/travisci-vs-jenkins

Staging Test URLs:
- https://epic-tvj-add-page-staging.test-deployments.travis-ci.com/travisci-vs-jenkins
- https://epic-tvj-add-page-staging.test-deployments.travis-ci.org/travisci-vs-jenkins

PR #2328, opened 2020-01-23
[TVJ] add contact section

sorta:
Part of https://github.com/travis-ci/travis-web/pull/2293
merges into epic branch

I noticed while testing this that on staging environments, the contact forms fail. This appears to be because the `travis-ci.org -- DEV` organization in Close has had its trial expire. The production test deployments work as expected, but the lead will go into the Close organization that production uses, so it's probably best to avoid that for now. I posted in slack about it, so that the relevant teams can discuss it.

### DIFF
--- a/app/components/sales-contact-thanks.js
+++ b/app/components/sales-contact-thanks.js
@@ -1,0 +1,7 @@
+import Component from '@ember/component';
+
+export default Component.extend({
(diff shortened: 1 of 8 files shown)


sorta:
Thanks for the review @emmanuel-ci !

PR #2329, opened 2020-01-24
"Head back" vulnerability fixed

piotr-travisci:


piotr-travisci:
Thank you @artursmirnov

PR #2330, opened 2020-01-27
[TVJ] Add testimonial section

sorta:
Part of https://github.com/travis-ci/travis-web/pull/2293

Merges into epic branch


app/components/ui-kit/grid-item.js
```diff
@@ -25,15 +26,15 @@ function screenClass(key, screen) {
 
 const FLEX_SIZES = {
   INITIAL: 'initial',
-  EVEN: 'even',
+  FREE: 'free',
```
sorta: I think using `one` or `1` for this could be weird or confusing, although it would be closest to the tailwind name. And I'm not sure that `even` is right either. This PR contains the first use of this utility, so now is a good time to decide what to call it.

app/components/ui-kit/box.js
```diff
@@ -8,10 +8,12 @@ import concat from 'travis/utils/ui-kit/concat';
 const COLORS = {
   WHITE: 'white',
   GREY_LIGHT: 'grey-light',
+  GREY_LIGHTER: 'grey-lighter',
 };
 
 const BG_COLORS = {
   [COLORS.WHITE]: 'white',
+  [COLORS.GREY_LIGHTER]: 'grey-100',
```
sorta: `grey-light` is `grey-150` in the border colors definition, so I thought a new value like `grey-lighter` might be appropriate here? Or maybe not, perhaps a 100 bg goes well with a 150 border, so maybe `grey light` works for both, though it seems a little weird

app/templates/travisci-vs-jenkins/index.hbs
```diff
@@ -192,8 +192,8 @@
   <UiKit::Box
     @tag='section'
     @borderWidth={{hash top='xs'}}
-    @margin={{hash bottom=16}}
-    @padding={{hash top=20}}
+    @margin={{hash bottom=24}}
+    @padding={{hash top=24}}
```
sorta: Once I saw the whole page altogether, I felt this section could use a little more spacing

app/components/ui-kit/box.js
artursmirnov: Sounds good to me too, thank you for the comment @sorta 👍

app/components/ui-kit/grid-item.js
artursmirnov: I'm not sure that I'm completely understanding the sense of the Tailwind property, but it looks like it just binds to `flex: <value>`. Is it correct? If so, then I'd like to suggest sticking with the CSS nature of the value, so that it was more intuitive. In this sense `even` sounds better to me. Or maybe `grow-single` (and eventually `grow-double`, `grow-triple`, etc.)? My concern about `free` is that most likely if I'd meet it somewhere in the code, I wouldn't understand what it's doing unless I look into implementation.

app/controllers/travisci-vs-jenkins/index.js
```diff
@@ -1,9 +1,14 @@
 import Controller from '@ember/controller';
 import { later } from '@ember/runloop';
+import config from 'travis/config/environment';
+
+const { blog: blogUrl } = config.urls;
 
 export default Controller.extend({
   referralSourceName: 'travisci-vs-jenkins',
 
+  caseStudyUrl: `${blogUrl}/2019-06-5-case-study-ibm-cloud-kubernetes-service`,
```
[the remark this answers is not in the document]
sorta: Yeah maybe so. Change made :)

app/templates/travisci-vs-jenkins/index.hbs
```diff
@@ -263,4 +263,79 @@
...
+
+
+        <UiKit::Grid @wrap='no-wrap' as |grid|>
+          <grid.item @flex='free' @padding={{hash top=2 right=6}}>
+            <SvgImage @name='kubernetes' @width={{85}} />
+          </grid.item>
+          <grid.item>
+            <UiKit::Text @margin={{hash bottom=6}} @size='2xl' @weight='light'>
+              &ldquo;
+              It has been a game changer for use because our process to deploy was hours if not
```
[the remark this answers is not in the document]
sorta: Ahh yeah totally. Thanks for catching it!

app/components/ui-kit/grid-item.js
sorta: Yeah that's pretty much correct, they are shorthands for `flex` property values. Here's the TW docs page for them https://tailwindcss.com/docs/flex . I agree, the naming is not really all that clear, so yeah maybe it would be good to replace with our own. So I've updated the tailwind config and the component to use such naming, let me know what you think.

sorta:
Thanks for the feedback @artursmirnov and @emmanuel-ci! All comments should be addressed now and this is ready for further review when you get a chance. Thanks!

artursmirnov:
It looks great to me, thank you for addressing the feedback, @sorta, amazing job 👍

sorta:
Excellent, thanks for the reviews @emmanuel-ci & @artursmirnov ! :tada:

PR #2331, opened 2020-01-27
Address issue where build os icon is not aligned with other icons/text

sorta:
It was reported that the apple logo at the bottom left of the build info box looked misaligned. For example:

https://travis-ci.com/travis-ci/dpl/jobs/280674895

Compare that with

https://so-realign-build-os-icon.test-deployments.travis-ci.com/travis-ci/dpl/jobs/280674895


sorta:
Thanks for the quick review @artursmirnov !

PR #2344, opened 2020-02-05
VCS name instead of hardcoded Github in ToolTips

piotr-travisci:


simi:
`s/instetad/instead/g`

piotr-travisci:
@simi 
👍

PR #2349, opened 2020-02-10
[BB] update vcs localstorage flags logic

sorta:
Currently, the bitbucket and assembla feature flags cannot be meaningfully set via env var, because of the way the setting of the flags from localstorage works. With the current logic, unless the correct localstorage key is set to 'true', the flag is disabled upon init of the `feature-flags` service. So even if you enable it via env var, it becomes immediately disabled on initial page load. I've updated the logic to only disable if the relevant localstorage key is set to 'false'.

Ultimately, I would like to remove this stuff entirely. It's not really necessary now that we can set the flags via env var, and it doesn't really belong in the feature flags service. But since Artur Kremens team uses these localstorage flags for testing, we should possibly wait or communicate/coordinate such a change.

I would also like to at some point rename the relevant feature flags (`enable-assembla-login` and `enable-bitbucket-login`).  None of our other flags have `enable` at the front, and I think it's a bit redundant.

But for now, I have only implemented the minimum change. When you get a chance, please let me know what you think!

### DIFF
--- a/app/services/feature-flags.js
+++ b/app/services/feature-flags.js
@@ -17,13 +17,19 @@ export default Service.extend({
   },
 
   _setEnableAssemblaLogin() {
-    const enableAssemblaLogin = window.localStorage['enableAssemblaLogin'] == 'true';
-    enableAssemblaLogin ? this.features.enable('enable-assembla-login') : this.features.disable('enable-assembla-login');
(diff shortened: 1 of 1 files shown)


sorta:
Yes, doing those other changes after release sounds good to me :+1: Thanks for the reviews!

PR #2354, opened 2020-02-11
double secure against wrong job number

piotr-travisci:


piotr-travisci:
Thank you @sorta for quick review.
Absolutely agree with you. We should refactor log and pusher part.

PR #2355, opened 2020-02-14
Pm trigger button restore

piotr-travisci:


app/components/branch-row.js
```diff
@@ -19,14 +20,20 @@ export default Component.extend({
...
   hasTriggered: false,
 
-  commitUrl: computed('branch.repository.{slug,vcsType}', 'branch.last_build.commit.sha', function () {
-    const [owner, repo] = this.get('branch.repository.slug').split('/');
-    const vcsType = this.get('branch.repository.vcs_type');
+  commitUrl: computed('branch.repository.id', 'branch.last_build.commit.sha', function () {
+    const repository = this.store.peekRecord('repo', this.get('branch.repository.id'));
+    const owner = repository.owner.login;
+    const repo = repository.name;
+    const vcsType = repository.vcsType;
```
artursmirnov: I have some concerns about this change. Generally, computed properties should operate only with data that's listed in the dependent keys. Here, `store` is another variable that's not tracked by the computed property. If `store` will change, this property won't recompute, thus getting out of sync and putting the app to invalid state. 

I can see that `branch` has a relation for [`repo`](https://github.com/travis-ci/travis-web/blob/master/app/models/branch.js#L11). Wouldn't it work without the need to `peekRecord` directly from store? Something like:
```javascript
const { owner, name, vcsType } = this.branch.repo;
```
To be honest, I'm not sure what's the `repository` that the code here is referring to 🤔

app/components/branch-row.js
```diff
@@ -19,14 +20,20 @@ export default Component.extend({
...
     const commit = this.get('branch.last_build.commit.sha');
-
     return this.externalLinks.commitUrl(vcsType, { owner, repo, commit });
   }),
 
+  vcsType: computed('branch.repository.id', function () {
+    const repository = this.store.peekRecord('repo', this.get('branch.repository.id'));
+    return repository.vcsType;
```
artursmirnov: Same here, if the store content will change, the computed property won't be notified and won't recompute.

piotr-travisci:
Thank you @artursmirnov for your constructive review.
You are absolutely right. The simpler the better!

app/components/branch-row.js
```diff
@@ -19,11 +19,10 @@ export default Component.extend({
   isTriggering: false,
   hasTriggered: false,
 
-  commitUrl: computed('branch.repository.{slug,vcsType}', 'branch.last_build.commit.sha', function () {
+  commitUrl: computed('branch.repository.{slug,vcsType}', 'branch.last_build.commit.sha', 'vcsType', function () {
```
artursmirnov: It seems like the `repository.vcsType` is not required as a dependency anymore, right?
```suggestion
  commitUrl: computed('branch.repository.slug', 'branch.last_build.commit.sha', 'vcsType', function () {
```

PR #2360, opened 2020-02-19
Copied over the PDF link and some copy from the Enterprise landing page

pauladamgordon:
@Pezi777 I've moved some of the content around to incorporate enterprise and added some of the copy from the enterprise landing page - we'll have to decide if this is okay, if we need to add anything else or if it needs a rewrite.

@Montana if you could check things over from a syntax point of view to make sure I haven't broken anything that would be great!

### DIFF
--- a/app/templates/plans/index.hbs
+++ b/app/templates/plans/index.hbs
@@ -172,6 +172,53 @@
       </grid.item>
     </UiKit::Grid>
   </UiKit::Box>
+  
+  {{!-- Enterprise section --}}
+  <UiKit::Box @tag='section' @margin={{hash bottom=20}} data-test-plans-page-enterprise-section="true">
+    <UiKit::Grid  @sm={{2}} @align='center' @gap={{4}} @justify='around' as |grid|>
+      <grid.item>
+        <UiKit::Box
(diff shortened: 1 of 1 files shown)


artursmirnov:
@ines-rx you can preview the changes using our test deployments: https://pauladamgordon-enterprise-migration.test-deployments.travis-ci.com/plans

Each pull request in `travis-web` repository is automatically deployed and is available for preview by a unique URL. This URL is submitted directly to PR as a "check" and can be found here:

image

pauladamgordon:
@artursmirnov Do you have any idea why Percy would be failing? The test deployment pages looks fine.

artursmirnov:
@pauladamgordon yes, it fails because it finds a difference compared to the `master` version. It just takes two screenshots (one from `master`, another from branch) and compares them. If there are any differences, it marks the check as failed and requires a review. It doesn't judge on its own whether the changes are valid or not, it just states there are changes. In this case the author of the changes needs to open the diff (by the Details link) and manually review the differences and "Approve" the expected ones.

In case of this PR all changes look valid to me, so I just approved them.

PR #2368, opened 2020-02-26
Fix empty pr list

sorta:
For https://travisci.assembla.com/spaces/VCS/tickets/118-pull-request-request-tab/details

In that ticket Alona posted a demonstration of the specific issue. Basically, clicking on a repo in the side bar and then on the pull requests tab would result in the empty builds message.

I tried to create a generic load more component that we could use elsewhere if we wanted to utilize a similar pattern.

Can test by signing in and clicking repos in the sidebar, then clicking on the Pull Requests tab. Compare the test deployment vs production:

https://so-fix-empty-pr-list.test-deployments.travis-ci.com/
https://travis-ci.com/

Other useful test URLs:
https://so-fix-empty-pr-list.test-deployments.travis-ci.com/travis-ci/php-src-builder.1/pull_requests
https://so-fix-empty-pr-list.test-deployments.travis-ci.com/travis-ci/travis-web-index/pull_requests
https://so-fix-empty-pr-list.test-deployments.travis-ci.com/travis-ci/travis-enterprise-worker-installers/pull_requests
https://so-fix-empty-pr-list.test-deployments.travis-ci.com/travis-ci/travis-web/pull_requests

Relevant:
https://github.com/travis-ci/travis-web/pull/1687
https://github.com/travis-ci/travis-api/pull/778


sorta:
Hi @AlonaHlobina, since this issue turned out to not really be specific to VCS work, I have this set to merge into `master` and it's not necessarily set up to have bitbucket enabled. Would you prefer that it is more associated with the bitbucket branch/integration?

app/models/repo.js
```diff
@@ -136,17 +140,27 @@ const Repo = VcsEntity.extend({
...
-      const isPullRequest = b.get('eventType') === 'pull_request';
-      return this._buildRepoMatches(b, id) && isPullRequest;
-    });
-    return this._buildObservableArray(builds);
-  }),
+  // Using dynamicQuery effectively relies on accurate pagination, which
+  // is not possible if skip_count=true, which is the default for queries in
+  // the build adapter. force_count allows overriding this behavior, but we
+  // should use it cautiously for performance reasons. I think it may
+  // be fine for pull requests though?
```
sorta: Pagination is used in the load more component to show/hide the "Show More" button, as well as the missing notice. If `skip_count=true`, then the Show More buttons is always visible, and the missing notice never is. Which is not ideal, but better than the Missing Notice showing when it should not, which is the current situation. Still, I might make some updates today to allow these things to work properly even if `skip_count=true`

AlonaHlobina:
> Hi @AlonaHlobina, since this issue turned out to not really be specific to VCS work, I have this set to merge into `master` and it's not necessarily set up to have bitbucket enabled. Would you prefer that it is more associated with the bitbucket branch/integration?

Thank you @sorta! I am fine with any of your decisions as far as it works for BB as well:-)

app/models/repo.js
sorta: Ok I made some updates to dynamicQuery to allow handling pagination when skip_count = true. If needed, we can move that logic elsewhere (like into the load more component, or possibly even into the api?).

app/models/repo.js
```diff
@@ -136,17 +140,27 @@ const Repo = VcsEntity.extend({
...
+  fetchBuilds({ page, filter, eventType, forceCount }) {
+    const { id: repoId, store } = this;
+    const offset = (page - 1) * limit;
+
+    return store.paginated('build', {
```
artursmirnov: Replacing `store.filter` by `store.paginated` may cause issues when builds are pushed to the store by Pusher. I'm not entirely sure, but `filter` seems to be something dynamic and updates its contents when store contents change, whereas `paginated` just returns whatever exists in the store at the moment and never updates unless called explicitly. But I may be wrong here, it's worth checking as a _possible_ issue 😄 Even if I'm right, maybe Pull Requests list doesn't actually need to be dynamic? 🤔

app/templates/components/load-more.hbs
```diff
@@ -0,0 +1,20 @@
...
+  {{#if this.missingNotice}}
+    {{#if this.isNoneFound}}
+      <MissingNotice @title={{this.missingNotice}} />
+    {{else}}
+      {{yield this}}
+    {{/if}}
+  {{!-- Otherwise, just yield and let them handle it if they want --}}
+  {{else}}
+    {{yield this}}
+  {{/if}}
```
artursmirnov: I'd like to suggest a little bit different approach here:
```hbs
  {{#if this.isNoneFound}}
    {{#if hasBlock}}
      {{yield this}}
    {{else}}
      <MissingNotice @title={{or (this.missingNotice 'Some default not found message')}} />
    {{/if}}
  {{/if}}
```
In this case when nothing's found it'll yield whatever is in the block if it `hasBlock` (built-in hbs helper), otherwise it'll show `missingNotice` or fallback to default message.
What do you think?

app/models/repo.js
sorta: That's a good point, I'll look into it a bit

sorta: I think you're right, this results in store changes from pusher to not be loaded. I tried switching back to `store.filter` but leaving in the other stuff, but it didn't help. And I tried a few other things as well, but everything I've tried so far either results in the original issue, won't load pusher updates, or only loads a single build until you click to show more.

I'd prefer not to lose the live updates from pusher, so we may need a different approach. I might try a few other things tomorrow, but if you have any suggestions, let me know.

Although I believe that ember 3.13+ breaks some of this same stuff, so if that's coming up soon anyway, we could hold on this until then. This issue of the missing notice showing on initial load is not specific to the vcs update, and seems to have existed for a while.

app/templates/components/load-more.hbs
sorta: Hmm this sounded good to me at first, but after thinking about it for a bit, I'm not sure. The `LoadMore` component is meant to be used in block form, so in this case would it always yield? Which actually is maybe fine, I'm now thinking maybe that whole bit should be replaced with just `{{yield this}}`. It would reduce the complexity of this component. And the component user is already responsible for looping through and displaying the items, so maybe it makes more sense anyway. WDYT?

sorta:
Artur's concern about `filter` vs `pagination` turned out to be valid. The implementation as it was at that point broke live updates of incoming pull requests coming in from pusher.

I tried a number of things that had various issues, including using `store.filter` along with dynamicQuery. But I think I've arrived at a modified solution that doesn't break live updates of pull requests. It again uses dynamicQuery to simplify pagination and querying for more items, but this time also uses computed properties to filter and sort results from `store.peekAll` in order to deal with live updates.

Ember data has deprecated `store.filter`, so usage of it relies on our custom implementation. Some more info:
https://github.com/emberjs/rfcs/blob/master/text/0326-ember-data-filter-deprecation.md
https://github.com/ember-data/ember-data-filter#recommended-refactor-guide

I'd like to rely less on those custom implementations. So I adapted one of their refactor suggestions to this situation.

Anyway, please let me know if you have any questions or suggestions!

sorta:

PR #2372, opened 2020-02-29
Fix links, proper provider and slug

piotr-travisci:


app/components/branch-row.js
```diff
@@ -26,6 +27,15 @@ export default Component.extend({
     return this.externalLinks.commitUrl(vcsType, { owner, repo, commit });
   }),
 
+  vcsType: computed('branch.repository.id', function () {
+    const repository = this.store.peekRecord('repo', this.get('branch.repository.id'));
+    return repository.vcsType;
+  }),
+
+  provider: computed('vcsType', function () {
+    return this.get('vcsType') && this.get('vcsType').toLowerCase().replace('repository', '');
```
sorta: I think in the long run it would be nice to have a standard, centrally located way to do this. Maybe something in  `app/utils/vcs.js`?

I think in one case what I did was get the provider config with one of those utils, and then read the provider's `urlPrefix`. But maybe a util like `vcsUrlPrefix` would be useful, to make it just like a single function to call. Or possibly we shouldn't rely on the urlPrefix, but rather add a `slug` property or something for this exact purpose.

Anyway, these are some optional ideas. Not really necessary for launch :+1:

app/templates/components/builds-item.hbs
```diff
@@ -81,7 +81,7 @@
   </h3>
   <div class="row-item row-commit">
     <ExternalLinkTo @href={{this.build.commit.url}}>
-      <EmberTooltip @text="{{concat "View commit on " (vcs-name this.build.repo.vcsType)}}" />
+      <EmberTooltip @text="{{concat "View commit on " (vcs-name this.vcsType)}}" />
```
sorta: It seems fine, though I'm not sure I understand the motivation for this change

piotr-travisci: We reduced repo representation in response and we don't have vcs_type in json, so this must be passed from parent element

sorta: ahh gotcha, makes sense :+1: Thanks!

app/components/created-by.js
```diff
@@ -14,5 +14,10 @@ export default Component.extend({
     let nameOrLogin = this.nameOrLogin;
     let eventType = this.eventType;
     return nameOrLogin && eventType !== 'cron';
+  }),
+
+  userProvider: computed('user.provider', 'provider', function () {
```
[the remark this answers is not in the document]
piotr-travisci: Nice catch. Thank you for suggestion

piotr-travisci:
Thank you @sorta for review and your suggestions.

PR #2377, opened 2020-03-02
[TVJ] Remove contact form and dev flag

sorta:
Marketing requested a final edit on the jenkins page: removal of the contact form and linking to the one on the plans page. This PR implements that update and enables the page to be visited by removing the dev flag.

Test: https://so-tvj-remove-contact-form.test-deployments.travis-ci.com/travisci-vs-jenkins


pauladamgordon:
Looks good to me! 🚢

app/controllers/plans/index.js
```diff
@@ -31,18 +40,18 @@ export default Controller.extend({
...
 
     toggleContactScroll() {
-      this.set('scrollToContact', true);
-      setTimeout(
-        () => this.set('scrollToContact', false),
-        500
+      this.set('section', SECTION.NONE);
+      later(
+        () => this.set('section', SECTION.CONTACT),
+        100
```
artursmirnov: I'm curious, would it work with just `next`? Theoretically, it should, because `next` ensures that all values from the previous run are set and in effect.
```js
next(() => this.set('section', SECTION.CONTACT))
```

app/controllers/travisci-vs-jenkins.js
```diff
@@ -0,0 +1,15 @@
...
+export default Controller.extend({
+  router: service(),
+
+  caseStudyUrl: config.urls.caseStudy,
+
+  goToContactForm() {
+    this.router.transitionTo('plans', {
+      queryParams: { section: 'contact' },
+    });
+  }
```
artursmirnov: It seems like this action is only used in templates, is it? If so, then probably it could be converted to `LinkTo`, what do you think @sorta? 
```hbs
<LinkTo @route='plans' @query={{hash section='contact'}}>
```

app/controllers/plans/index.js
sorta: Oh yeah, that probably would work. I'll try it out.

app/controllers/travisci-vs-jenkins.js
sorta: Yeah that makes sense, I'll make that change.

sorta:
Thanks for the review @artursmirnov! Those are good questions, and I feel similarly. I honestly dislike this change, the UX of it seems bad. But the Marketing team has been quite adamant that this is exactly what they want.

Possibly they aren't aware that the form could be (and already is) a single component that can be used in multiple places. Or maybe Pardot has some quirk that would make that awkward for some reason.

Making the UX bad for marketing purposes seems backwards to me, but it's not my specialty. Perhaps @pauladamgordon can give us some context as to how/why this decision was made?

In the meantime, I'll work on your suggested changes. Thanks!

app/controllers/plans/index.js
sorta: Change made :)

app/controllers/travisci-vs-jenkins.js
sorta: Change made :)

sorta:
The suggested tweaks have been implemented, and this is ready for a second review if desired. So questions about this strategy aside, the code changes for it are complete.

Although, I just realized a possible issue. If you're logged in and go to `/plans`, you'll be redirected to `/account/subscription`, and wouldn't be able to access the plans page contact form. I was going to try to test it on this test deployment, but I can't seem to log in on it. I get `target URI not allowed` during the handshake. Is that happening for anyone else? Login works fine when running locally and tests pass, so I think it must be something related to the test deployment, and possibly api?

artursmirnov:
@sorta, thank you for addressing the feedback! it looks great to me! 

@pauladamgordon @AlonaHlobina what do you think about the discussion above? Would it make sense to revisit the possibility of having a single form reused across the app?

pauladamgordon:
@artursmirnov this is something that will need discussing with @ines-rx - I think the original reasoning behind this was the migration to Salesforce and not wishing to have multiple forms to maintain across the site.

artursmirnov:
@pauladamgordon thank you for the details! I thought it would be the case. Do you think it would work if we re-use the same single form in multiple places across the site? It should be easy to replace with anything else that'd be required by the integration process and easy to maintain later. On the other hand it could significantly improve user experience. 

@ines-rx what do you think?

pauladamgordon:
Hi @artursmirnov ! We discussed this in the marketing meeting yesterday and decided that for logged-in users it would be better to allow the redirection to `/account/subscription`, as most people who are already logged in are not going to be accessing the plans page directly. Would that be okay?

artursmirnov:
Hi @pauladamgordon, yes, for sure, redirecting logged in users from `/plans` to `/account/subscription` totally makes sense to me 👍 But my question was for a bit different topic - whether it'd be possible to reuse the contact form on multiple pages across the app instead of redirecting a user to the single page

pauladamgordon:
Ah, sorry @artursmirnov ! That would definitely be a question for @ines-rx , unfirtunately I haven't been involved in the Salesforce migration to fully understand how the Pardot form work

ines-rx:
@pauladamgordon @artursmirnov theoretically it would work to reuse the form if the CSS is not changing. however for now to simplify things for the go-live next week please just use the link instead of the form... we can change it later on anyway :) thanks!

artursmirnov:
Should we close this PR assuming that both form and feature flag are already on production?

sorta:
Yes, that sounds good to me. The alternate PR is merged, so this should be good to go. Thanks!

PR #2378, opened 2020-03-04
Add github apps app name variable handling to waiter

artursmirnov:
This change allows to set github app name via env variable for instances running by waiter


artursmirnov:
Thanks for reviews everyone! 🎉

PR #2380, opened 2020-03-06
Remove provider icon from my builds list

artursmirnov:
When we removed the provider icon from the dashboard repositories list we forgot to remove it from my builds. So, fixed 🚀


artursmirnov:
@sorta thanks for such a swift review 🚀

PR #2384, opened 2020-03-10
Test quick fix

piotr-travisci:


piotr-travisci:
It's like you said. Thank you Artur for quick review.

PR #2391, opened 2020-03-13
Job owner checking refactored due to spaces in repo slug

piotr-travisci:


app/models/repo.js
```diff
@@ -81,6 +81,13 @@ const Repo = VcsEntity.extend({
     }
   }),
 
+  // slug built from normalized (by provider) owner and repo name
+  vcsSlug: computed('ownerName', 'vcsName', function () {
+    const ownerName = this.get('ownerName');
+    const vcsName = this.get('vcsName');
+    return `${ownerName}/${vcsName}`;
```
artursmirnov: I don't think `get` is necessary here, as both values seems to be just plain properties, aren't they?
```suggestion
    return `${this.ownerName}/${this.vcsName}`;
```

piotr-travisci:
Nice suggestion. Thank you @artursmirnov !

PR #2393, opened 2020-03-13
add request configs model, serializer, adapter

svenfuchs:


app/adapters/request-config.js
```diff
@@ -0,0 +1,26 @@
...
+    'X-Client-Release': config.release
+  },
+
+  token: reads('auth.token'),
+
+  queryRecord(store, type, { id, data }) {
+    const url = `${apiEndpoint}/repo/${id}/request/config`;
+    const headers = { ...this.headers, Authorization: `token ${this.token}` };
+    const options = { data, headers: headers};
+    return this.ajax.request(url, 'POST', options);
```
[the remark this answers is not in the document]
svenfuchs: makes sense. i'll try doing that. thank you, @emmanuel-ci!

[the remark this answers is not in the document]
svenfuchs: done!

app/serializers/request-config.js
```diff
@@ -0,0 +1,9 @@
+import V3Serializer from 'travis/serializers/v3';
+
+export default V3Serializer.extend({
+  normalize(modelClass, resourceHash) {
+    resourceHash.id = new Date().getTime();
```
sorta: Does `resourceHash.id` ever already exist coming into this function? If so, it may make sense to do something like `resourceHash.id = resourceHash.id || new Date().getTime();`. If not, feel free to ignore this

svenfuchs: no, it won't ever have an id, @sorta. this endpoint expands the posted request config _as if_ it would actually be processed by Gatekeeper/Yml, and only then _would_ be persisted.  so there's pretty much no way the client would ever get an id on this. i believe the key is present in the api payload because our v3 api abstraction makes it extremely hard to omit it, when using all the default framework tooling :)

sorta: Cool, makes sense :)

mirage/config.js
```diff
@@ -621,6 +621,49 @@ export default function () {
...
+        error_type: 'invalid_config_format',
+        error_message: 'Invalid value: invalid'
+      });
+    } else {
+      return new Response(200, {}, {
+        raw_configs: requestConfig.raw_configs,
+        request_config: requestConfig.request_config,
+        job_configs: requestConfig.job_configs,
+        messages: requestConfig.messages,
+      });
```
artursmirnov: It seems like just `requestConfig` should work as well:
```suggestion
      return new Response(200, {}, requestConfig);
```

mirage/config.js
```diff
@@ -621,6 +621,49 @@ export default function () {
...
+          level: 'info',
+          key: 'root',
+          code: 'default',
+          args: {
+            key: 'os',
+            default: 'linux'
+          }
+        }
+      ]
+    });
```
artursmirnov: I'd like to suggest creating a [mirage factory](https://www.ember-cli-mirage.com/docs/data-layer/factories) for `request-config` and make it build a record with default values. This would make it easier to adjust the records in the future. Also maybe some field can use [traits](https://www.ember-cli-mirage.com/docs/data-layer/factories#traits) to shape data.

In this case you'd just need to call `const requestConfig = server.create('request-config')` to create a record.

svenfuchs:
added to https://github.com/travis-ci/travis-web/pull/2399, so i'm closing this one here

app/adapters/request-config.js
```diff
@@ -0,0 +1,11 @@
+import V3Adapter from 'travis/adapters/v3';
+import config from 'travis/config/environment';
+
+const { apiEndpoint } = config;
+
+export default V3Adapter.extend({
+  queryRecord(store, type, { id, data }) {
+    const url = `${apiEndpoint}/repo/${id}/request/config`;
```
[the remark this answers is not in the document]
svenfuchs: oh, i had not seen `buildURL` ... i suppose we can even use `${this.buildURL('repo', id)}/request/config` then? i'm gonna try that out.

app/services/request-config.js
```diff
@@ -0,0 +1,39 @@
...
+  errorMessages: computed(() => []),
+  messages: or('loadConfigsResult.messages', 'errorMessages'),
+
+  loadConfigs: task(function* (id, data, debounce) {
+    if (debounce) {
+      const { searchDebounceRate } = config.intervals;
+      yield timeout(searchDebounceRate);
+    }
+    try {
+      return yield this.store.queryRecord('request-config', { id: id, data });
```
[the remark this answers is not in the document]
svenfuchs: oh wow, that syntax always completely trips me up. i'll change this.

app/services/request-config.js
```diff
@@ -0,0 +1,39 @@
...
+    }
+  }).restartable(),
+
+  handleLoadConfigError(e) {
+    if (e.json) {
+      e.json().then(e => {
+        this.set('errorMessages', [{ level: 'error', code: e.error_type, args: { message: e.error_message } }]);
+      });
+    }
+  },
```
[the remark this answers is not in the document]
svenfuchs: @artursmirnov this whole handler will be changed in a later PR because it actually does not work. the exception raised here does not match the expectation of this logic any more.

let's address the variable naming in that later PR.

mirage/config.js
svenfuchs: makes sense 👍

mirage/config.js
svenfuchs: makes sense.

i'll give this a try, but i'm not sure if i'm able to implement that. this was contributed by Emmanuel, so maybe i'll have to rely on him for this.

app/services/request-config.js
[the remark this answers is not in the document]
artursmirnov: Absolutely, thanks for the explanation 👍

mirage/config.js
artursmirnov: This is totally optional, just a suggestion ;) If you don't have time for it, feel free to keep it as is, it works perfectly.

PR #2395, opened 2020-03-13
Disable log interaction when it's destroyed

artursmirnov:
Ticket: https://travisci.assembla.com/spaces/VCS/tickets/154/details


artursmirnov:
@sorta the issue is kind of occasional and random. It sometimes appear when you visit a build (to create a `Log` instance for the `build`), then go to Dashboard and then restart the build from My Builds section. The restart will try to `clearLog`, but the component is already destroyed, since the log page isn't rendered, which will lead to the error in console.

PR #2402, opened 2020-03-13
add config customize form

svenfuchs:
based on https://github.com/travis-ci/travis-web/pull/2401

with this in place the triggered build's config can be customized


app/templates/components/request/config-form.hbs
```diff
@@ -0,0 +1,144 @@
...
+  <div class="request-configs-fields" data-test-build-config-form='true'>
+    <div class="form-elem">
+      {{#radio-button
+        value="sha"
+        groupValue=refType
+        changed=(action "change" "refType")
+        classNames="ref-type"
+      }}
+        Commit
+      {{/radio-button}}
```
sorta: If possible, it's preferred to use angle bracket syntax for components. In this case, it would be something like:
```
<RadioButton
  @value='sha'
  @groupValue={{refType}}
  @changed={{action 'change' refType}}
  @classNames='ref-type'
>
  Commit
</RadioButton>
```

Additionally, in this case I would lean towards making our own component rather than using `ember-radio-button`. We have similar components in `app/components/forms` that we could base a radio component on, and I have some concerns about how actively maintained the `ember-radio-button` addon is. But also, I know that making such a component can be some complex work sometimes.

app/templates/components/request/config-form.hbs
```diff
@@ -0,0 +1,144 @@
...
+        data-test-request-configs-config="true"
+        @label="Custom Build Config"
+        as |field|
+      >
+        <div class="config-type {{this.configMode}}">
+          <span class="yaml">YAML</span>
+          <span class="javascript">JSON</span>
+        </div>
+
+        {{ivy-codemirror
```
sorta: Same about angle bracket syntax here, if possible

app/styles/app/modules/request/form.scss
```diff
@@ -0,0 +1,93 @@
...
+    &.javascript .javascript,
+    &.yaml .yaml {
+      color: white;
+      background-color: $oxide-blue;
+      border-color: $oxide-blue;
+    }
+  }
+
+  .travis-form__field-component {
+    box-shadow: inset 0 0 1px 1px $dry-cement !important;
```
sorta: Same here about use of `!important`

app/styles/app/modules/request/form.scss
```diff
@@ -0,0 +1,93 @@
...
+    height: 350px !important;
+    margin: -4px;
+    padding: 8px 12px 8px 4px;
+    width: calc(100% + 8px);
+    border-radius: 10px;
+    border: 4px solid transparent;
+    box-shadow: inset 0 0 1px 1px $dry-cement;
+    transition: border-color 0.2s ease-in;
+    color: $asphalt-grey;
+    line-height: 150%;
```
sorta: This is minor/optional but for line-height a unitless value is usually preferred. In this case, it would be like `line-height: 1.5;`

https://developer.mozilla.org/en-US/docs/Web/CSS/line-height#%3Cnumber%3E

svenfuchs: interesting! i'll change this.

app/templates/components/request/config-form.hbs
svenfuchs: will try

app/templates/components/request/config-form.hbs
svenfuchs: i understand the concerns about adding a dependency that isn't maintained actively. i'll try to remove it.

out of curiosity, what's the reasoning behind preferring the angle bracket syntax? my impression was that most documentation used the curly braces (helper?) syntax.

app/styles/app/modules/request/configs.scss
```diff
@@ -3,6 +3,18 @@
     border: 1px solid $dry-cement;
     margin-bottom: 1em;
   }
+
+  h4 {
```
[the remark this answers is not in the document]
svenfuchs: ok? i'm curious why :) it's always been my habit to try and keep things simple and puristic, but i can see how maybe that is a little out of date.

i will of course change this.

app/styles/app/modules/request/configs.scss
```diff
@@ -3,6 +3,18 @@
...
+    color: $cement-grey;
+    display: block;
+    text-transform: uppercase;
+    font-weight: 600;
+    font-size: 0.9em;
+  }
+
+  h4, label {
+    margin-bottom: 0.75em !important;
```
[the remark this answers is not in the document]
svenfuchs: i agree, although it seems like a bit of an arms race in adding lots and lots of selectors just to win over some highly specific selectors that are used in some more widely used declarations (such as the forms stuff). i think that's the actual issue here (common styles shouldn't have super specific selectors imo, in order to allow more specific cases to overwrite them more easily)

but i'll try to address this, and remove the `!important` declaration.

app/components/request/configs.js
```diff
@@ -1,9 +1,16 @@
...
 import { equal, or, reads } from '@ember/object/computed';
 import { inject as service } from '@ember/service';
 import CanTriggerBuild from 'travis/mixins/components/can-trigger-build';
 import TriggerBuild from 'travis/mixins/components/trigger-build';
 
+function truncated(property, chars) {
+  return computed(property, function () {
+    return this.get(property).slice(0, 7);
+  });
+}
```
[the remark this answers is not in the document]
svenfuchs: i'm not sure, `chars`? this is meant to cut the long sha to only the first 7 characters, like we do that in the build header (and GitHub does it, too).

`String.prototype.slice` seems like a good way to do that, doesn't it? is there a better way? i couldn't find anything like `left` or so.

app/styles/app/modules/request/configs.scss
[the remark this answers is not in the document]
sorta: Eh, it's not really a big deal, you can leave this as-is if you want. It's actually a bit of a debate in the frontend world about what are the best ways to use the cascade. I'll gather some info about it and add another comment :)

app/templates/components/request/config-form.hbs
sorta: Angle bracket syntax is a somewhat recent addition to ember, so older addons, forum posts etc. will tend to have the curly brace syntax. It seems to be the way the ember community is going now, and IIRC one of the main ideas was to make it easier to differentiate components from other handlebars things (if/else, helpers, etc). It also allows clearer separation between parameters and html attributes. Here's some more info:

https://guides.emberjs.com/v3.17.0/upgrading/current-edition/templates/#toc_benefits-of-angle-brackets

app/components/request/configs.js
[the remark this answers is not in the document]
sorta: Yeah the usage here of `slice` seems fine, I was just wondering what the argument `chars` in the function `truncated` is for, it doesn't seem to be used. I saw elsewhere a call of something like `truncated('propertyName', 10)`, so I figured it was possibly a way to use the function for different lengths.

app/styles/app/modules/request/configs.scss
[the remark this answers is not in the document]
sorta: Yeah that's one of the issues that have come up with the rise of things like SCSS, it made things like nesting super easy. This is part of what things like BEM and utility classes attempt to address.

`!important` is still kind of an escalation in that arms race, but focused on a single property. If you ever need to override it, now you need a more specific selector _and_ `!important`

app/components/request/configs.js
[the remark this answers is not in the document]
svenfuchs: ooooooh, haha, totally missed that! i suppose, yes? i have copied this code from somewhere on the ember forum. i'll have another look at this, thank you!

app/styles/app/modules/request/configs.scss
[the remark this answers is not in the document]
svenfuchs: but this is MY STYLE SO IT'S REALLY IMPORTANT!!! :) just kidding ... 

i believe i have addressed this in https://github.com/travis-ci/travis-web/commit/5003d2b2e8ab86a2dbb2819bf4015c6e3255d755

app/styles/app/modules/request/configs.scss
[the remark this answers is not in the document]
svenfuchs: i see :) it's alright ...

i believe i've addressed this in https://github.com/travis-ci/travis-web/commit/5003d2b2e8ab86a2dbb2819bf4015c6e3255d755

app/styles/app/modules/request/form.scss
svenfuchs: i believe i've addressed this in https://github.com/travis-ci/travis-web/commit/5003d2b2e8ab86a2dbb2819bf4015c6e3255d755

app/styles/app/modules/request/form.scss
svenfuchs: addressed in https://github.com/travis-ci/travis-web/commit/18b6739765aeb142ff5a8cd69cfd652af9d6a8bb

app/templates/components/request/config-form.hbs
svenfuchs: i actually like this much better, too. thanks for the pointer, i'll have a look!

app/styles/app/modules/request/configs.scss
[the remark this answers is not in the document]
sorta: Haha, I mean I totally get it! I tend to think that my styles are the most important XD

Looks great, thanks for addressing! I see what you mean about the form styles, I agree that the specificity on them is too high. I have an idea for how they could be flattened, possibly without too much trouble, so I might experiment with that a bit sometime.

app/styles/app/modules/request/configs.scss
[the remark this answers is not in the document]
sorta: Looks good to me :+1: Thanks!

app/styles/app/modules/request/form.scss
sorta: Looks good :+1: Thanks!

app/styles/app/modules/request/form.scss
sorta: Nice, thanks!

app/templates/components/request/config-form.hbs
sorta: Awesome :smile: I feel like I should mention, the guide I linked is for an ember version that is a few ahead of where we're at (3.12). But I think everything on that particular page is applicable to 3.12, and it has a great summary of the advantages of angle bracket syntax

app/components/request/configs.js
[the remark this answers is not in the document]
svenfuchs: addressed in https://github.com/travis-ci/travis-web/commit/b2c75a2b9807e9bfd963abcf18cd2281d5aa3a72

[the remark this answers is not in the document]
sorta: Looks good :+1: Thanks!

app/templates/components/request/config-form.hbs
svenfuchs: addressed in https://github.com/travis-ci/travis-web/commit/4a7aa854a101dc651328fe30943edbb727ddccd4

app/templates/components/request/config-form.hbs
svenfuchs: addressed in https://github.com/travis-ci/travis-web/commit/4a7aa854a101dc651328fe30943edbb727ddccd4

svenfuchs:
i have added this to https://github.com/travis-ci/travis-web/pull/2399. gonna close this one here.

app/components/request/configs.js
```diff
@@ -44,6 +56,15 @@ export default Component.extend(CanTriggerBuild, TriggerBuild, {
     }
   },
 
+  onCustomize() {
+    if (this.status == 'customize') {
```
artursmirnov: I'd suggest using `customizing` computed property here:
```suggestion
    if (this.customizing) {
```

app/components/request/messages.js
```diff
@@ -12,18 +12,21 @@ const MSGS = {
 };
 
 export default Component.extend(WithConfigValidation, {
-  tagName: '',
-  isExpanded: false,
+  tagName: 'div',
```
artursmirnov: `div` is default, no need to specify it explicitly

app/styles/app/modules/request/configs.scss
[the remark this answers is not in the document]
artursmirnov: @svenfuchs frontend apps are getting more and more complex, and stylesheets are growing as well. And in such a complex apps it's hard to manage cascading, especially with componential approach. Using just tag selectors you define styles for all nested elements of this tag, so if markup would change and someone would add another nested element with the same tag, your style will apply to it as well. And It's sometimes very hard to overwrite. With complexity growing, it could quickly become a mess. Mainly because of this, currently there's a trend to avoid cascading at all whenever possible (there are post-css modules that add hashes to classes to make them unique), or at least scope styles to specific components (BEM for example). But using classes is already helping a lot in this sense, so it's basically like a rule of thumb.

app/templates/components/request/config-form.hbs
```diff
@@ -0,0 +1,144 @@
...
+    <div class="form-elem">
+      <h4>
+        Merge mode
+      </h4>
+      {{#radio-button
+        value="deep_merge_append"
+        groupValue=mergeMode
+        changed=(action "change" "mergeMode")
+      }}
+        Deep Merge & Append
```
artursmirnov: The `&` is a reserved symbol in HTML. It works in most cases but may be unreliable. It's better to use `&amp;` html entity instead.

app/templates/components/request/config-form.hbs
```diff
@@ -0,0 +1,144 @@
...
+      {{/radio-button}}
+      {{#radio-button
+        value="deep_merge_prepend"
+        groupValue=mergeMode
+        changed=(action "change" "mergeMode")
+      }}
+        Deep Merge & Prepend
```
artursmirnov: Same here

app/templates/components/request/config-form.hbs
```diff
@@ -0,0 +1,144 @@
...
+      }}
+        Replace
+      {{/radio-button}}
+
+      <ExternalLinkTo class="hint" href="{{config-get 'urls.triggerBuildMergeModes'}}">
+        <SvgImage @name='icon-help' @class='icon-help' />
+      </ExternalLinkTo>
+    </div>
+  </div>
+</TravisForm>
```
artursmirnov: Is it by design that this form doesn't have a submit button? Or did I miss it?

app/templates/components/request/config-form.hbs
artursmirnov: @svenfuchs how do you feel about trying to use our existing [form-checkbox](https://github.com/travis-ci/travis-web/blob/master/app/components/forms/form-checkbox.js) component instead of that outdated add-on? It looks very much like a radio (except it has a checkmark, but that shouldn't be an issue) and it's easy to make it behave like a radio (uncheck others on check).

app/templates/components/request/config-form.hbs
svenfuchs: the button is included in the parent component `request/configs`, and was added in the previous branch. eventually there will be 4 different buttons, with one of them being the main submit `Trigger Build` button. this is how it looks like in the final branch: https://github.com/travis-ci/travis-web/blob/sf-request-config-tweaks/app/templates/components/request/configs-buttons.hbs#L26-L38

app/templates/components/request/config-form.hbs
svenfuchs: done: https://github.com/travis-ci/travis-web/commit/f3eaa3c8c8ad2e32edbd6240e6d66aeb53fe1bdd

app/templates/components/request/config-form.hbs
svenfuchs: done: https://github.com/travis-ci/travis-web/commit/f3eaa3c8c8ad2e32edbd6240e6d66aeb53fe1bdd

app/components/request/configs.js
svenfuchs: yep! this has already happened at some later point: https://github.com/travis-ci/travis-web/blob/sf-request-config-tweaks/app/components/request/configs.js#L64

app/components/request/configs.js
```diff
@@ -1,9 +1,16 @@
 import Component from '@ember/component';
+import { computed } from '@ember/object';
 import { equal, or, reads } from '@ember/object/computed';
 import { inject as service } from '@ember/service';
 import CanTriggerBuild from 'travis/mixins/components/can-trigger-build';
 import TriggerBuild from 'travis/mixins/components/trigger-build';
 
+function truncated(property, chars) {
```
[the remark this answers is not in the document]
svenfuchs: happy to move it: https://github.com/travis-ci/travis-web/commit/8c51cc836c420bd0f63072b096bd8e0232b375c0

please let me know if this is not the right way to do it

app/components/request/config-form.js
```diff
@@ -0,0 +1,72 @@
...
+  }),
+
+  actions: {
+    change(field, value) {
+      this.onChange(field, value);
+    },
+
+    submit() {
+      this.onSubmit();
+    }
```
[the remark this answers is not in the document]
svenfuchs: sounds good. added: https://github.com/travis-ci/travis-web/commit/58dca855f2b80b1dedcf7309600fcf314eca5619

i'm not sure if this is the right place to put them in the file ... please advise if they should sit elsewhere.

app/components/request/config-form.js
```diff
@@ -0,0 +1,72 @@
...
+  configMode: computed('config', function () {
+    const { config } = this;
+    const { JAVASCRIPT, YAML } = CONFIG;
+    return config && config.startsWith('{') ? JAVASCRIPT : YAML;
+  }),
+
+  configType: computed('configMode', function () {
+    const { JAVASCRIPT, JSON, YAML } = CONFIG;
+    return this.configMode === JAVASCRIPT ? JSON.toUpperCase() : YAML.toUpperCase();
+  }),
```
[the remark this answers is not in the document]
svenfuchs: yeah, i believe you are right. i have removed it for now: https://github.com/travis-ci/travis-web/commit/e3e5de35a945c26a4f405a94b5247c6ff4a2d7ca

app/components/request/config-form.js
```diff
@@ -0,0 +1,72 @@
...
+  SOURCE: {
+    API: 'api'
+  },
+  JAVASCRIPT: 'javascript',
+  YAML: 'yaml',
+  JSON: 'json',
+};
+
+export default Component.extend(BranchSearching, {
+  tagName: 'div',
```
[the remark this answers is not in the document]
svenfuchs: that makes sense :) i've removed it from all the places i could find https://github.com/travis-ci/travis-web/commit/301dc629b9e9b05defaf0bc0b2583243b1e5091d (i am pushing this to the final `tweaks` branch, so i am not creating too many conflicts on these intermediate branches)

app/components/request/configs.js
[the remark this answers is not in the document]
artursmirnov: Looks perfect, thank you! 👍

app/templates/components/request/config-form.hbs
artursmirnov: I'm not sure, but does it mean that 
```hbs
<TravisForm
  @onSubmit={{perform this.triggerBuild}}
  as |form|
>
```
doesn't make sense as `onSubmit` will never be triggered?

artursmirnov:
@svenfuchs thank you for addressing the feedback 👍

PR #2403, opened 2020-03-14
add request configs preview

svenfuchs:
based on https://github.com/travis-ci/travis-web/pull/2402

with this in place the triggered build config can be previewed


app/components/request/configs.js
```diff
@@ -16,63 +16,106 @@ export default Component.extend(CanTriggerBuild, TriggerBuild, {
...
+      },
+      ref: this.ref,
+      mode: this.mergeMode,
+      config: this.config,
+      data: {
+        branch: this.branch,
+        commit_message: this.message
+      },
+      type: 'api'
+    };
```
artursmirnov: Some Javascript-fu to make it a little bit cleaner:
```javascript
    const { repo: repoRecord, ref, config, branch, mergeMode: mode, message } = this;
    const { slug, private, defaultBranch } = repoRecord;
    const repo = { slug, private, default_branch: defaultBranch.name };
    const data = { branch, commit_message: message };
    const type = 'api';
    
    return { repo, ref, mode, config, data, type };
```
Completely optional, of course ;)

app/components/request/configs.js
```diff
@@ -83,7 +126,7 @@ export default Component.extend(CanTriggerBuild, TriggerBuild, {
   },
 
   submit() {
-    this.set('processing', true);
+    this.set('submitting', true);
```
artursmirnov: Would it make sense to bind `submitting` to `this. submitBuildRequest.isRunning` instead of manual setting?

app/styles/app/modules/request/preview.scss
```diff
@@ -0,0 +1,27 @@
...
+
+  pre,
+  div > div {
+    height: calc(100% - 45px);
+    border: 1px solid $dry-cement;
+  }
+
+  div > div {
+    padding: 10px 20px;
+  }
```
artursmirnov: I'd like to suggest changing these selectors to use classes instead of tags, to make styles less dependent on the template markup structure, and thus, less fragile.

app/templates/components/request/configs-buttons.hbs
```diff
@@ -0,0 +1,37 @@
...
+  >
+    Customize
+  </a>
+  <a
+    data-test-request-configs-button-preview="true"
+    class="request-configs-button request-configs-button-preview"
+    {{on 'click' (action this.onPreview)}}
+  >
+    Preview
+  </a>
```
artursmirnov: I'd like to suggest transforming these buttons to be `<button>`s, not links. Links suppose to navigate somewhere. We have multiple classes [here](https://github.com/travis-ci/travis-web/blob/180f93652338b17fbcea4352eef48faf2c2eed6f/app/styles/app/modules/buttons.scss) that can help to reset default button styles. And if you need a cursor to be `pointer`, there's a class `.pointer` that will do the trick.

svenfuchs: sure thing! done: https://github.com/travis-ci/travis-web/commit/0c3332ab3a57adcec56b6724d389b184aaf89776

app/styles/app/modules/request/preview.scss
svenfuchs: makes sense. Sophie had a similar request on the previous branch. 

done: https://github.com/travis-ci/travis-web/commit/ee375ae73174a2030c5bd731ea2165ddb222e474

app/components/request/configs.js
svenfuchs: yes, absolutely. in fact Emmanuel asked that same question on an earlier PR, and we've postponed it. but now that i'm making all changes to the final branch, there's no point in delaying things any further.

done: https://github.com/travis-ci/travis-web/commit/8d1e6dfbc237a6afe6d052ba44b5b3b109ccdabf

app/components/request/configs.js
svenfuchs: woha! i'd love to use that. unfortunately there seems to be a syntax error that i just cannot make sense of:

```
   97 |   data: computed('repo', 'ref', 'mergeMode', 'config', function () {
   98 |     const { repo: repoRecord, ref, config, branch, mergeMode: mode, message } = this;
>  99 |     const { slug, 'private', defaultBranch } = repoRecord;
      |                            ^
  100 |     const repo = { slug, private, default_branch: defaultBranch.name };
  101 |     const data = { branch, commit_message: message };
  102 |     const type = 'api';
```

app/components/request/configs.js
```diff
@@ -16,63 +16,106 @@ export default Component.extend(CanTriggerBuild, TriggerBuild, {
...
+
   onCancel() {
     this.set('status', 'closed');
     this.reset();
   },
 
+  load(debounce) {
+    if (this.customizing || this.previewing) {
+      this.preview.loadConfigs.perform(this.repoId, this.data, debounce);
+      this.set('loaded', true);
```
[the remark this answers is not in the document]
svenfuchs: oh, most probably, yes. i'll try that.

done: https://github.com/travis-ci/travis-web/commit/9881f3578ef33be42209b6de6700ea663adf227d

i had to move it to the service because otherwise it would raise when restarted. but this looks better to me anyway.

app/components/request/configs.js
```diff
@@ -16,63 +16,106 @@ export default Component.extend(CanTriggerBuild, TriggerBuild, {
...
     }
   },
 
+  onPreview() {
+    if (!this.previewing) {
+      this.set('status', 'preview');
+      if (!this.loaded) {
+        this.load();
+      }
+    } else if (this.customized) {
```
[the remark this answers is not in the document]
svenfuchs: no, it is going to be set if the user has changed any of the form values (i.e. "customized" some of the values on the "customize" form). 

this is one of the bigger UX questions of how the buttons should behave. i think it will be best to have another discussion about this once we have the final branch done, and can look at a branch deployment together.

let's keep this for now.

app/components/request/configs-trigger-build-notice.js
```diff
@@ -1,15 +1,21 @@
...
+import { equal } from '@ember/object/computed';
 import { pluralize } from 'ember-inflector';
 
 export default Component.extend({
   tagName: 'p',
   classNames: ['request-configs-trigger-build-notice'],
 
   configs: 0,
-  customize: false,
+  status: undefined,
```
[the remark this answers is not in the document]
svenfuchs: ok, of course. i never know which one to use where in Javascript :)

done: https://github.com/travis-ci/travis-web/commit/a4fc52e1ccc343f2b92ea078f24a3794a38c7113

app/components/request/configs-preview.js
```diff
@@ -0,0 +1,21 @@
+import Component from '@ember/component';
+import { computed } from '@ember/object';
+import { reads, mapBy } from '@ember/object/computed';
+
+export default Component.extend({
+  tagName: 'div',
```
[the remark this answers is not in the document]
svenfuchs: yep, it is, on the final brach https://github.com/travis-ci/travis-web/pull/2407/commits/301dc629b9e9b05defaf0bc0b2583243b1e5091d

app/components/request/configs.js
sorta: I think on the `const { slug, 'private', defaultBranch } = repoRecord;` line, removing the quotes around `private` may do it

app/components/request/configs-trigger-build-notice.js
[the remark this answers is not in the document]
artursmirnov: It's easy - always use `null`, the only case for `undefined` is to check if a variable has not been initialized (argument not passed, input is empty, etc.), e.g. `value === undefined`.

app/components/request/configs.js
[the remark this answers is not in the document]
artursmirnov: Sounds reasonable, thank you for explaining 👍

app/components/request/configs.js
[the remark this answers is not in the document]
artursmirnov: @svenfuchs thank you for the fix! Although, I'd like to suggest a little bit different approach: https://github.com/travis-ci/travis-web/commit/9881f3578ef33be42209b6de6700ea663adf227d#r37998445 Please take a look when you got a chance and let me know what you think.

app/templates/components/request/configs-buttons.hbs
artursmirnov: Awesome, thanks! 😄

app/components/request/configs-trigger-build-notice.js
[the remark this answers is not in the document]
svenfuchs: hah, ok, thanks!

app/components/request/configs.js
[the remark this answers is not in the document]
svenfuchs: thanks! i've commented over there with a question ...

app/components/request/configs.js
svenfuchs: ah, sorry ... no, unfortunately not. copied the wrong part of the output. i had tried that one (Artur's original suggestion) first, and then thought maybe `private` is keyword that needs quoting.

```   98 |   data: computed('repo', 'ref', 'mergeMode', 'config', 'message', function () {
   99 |     const { repo: repoRecord, ref, config, branch, mergeMode: mode, message } = this;
> 100 |     const { slug, private, defaultBranch } = repoRecord;
      |                   ^
  101 |     const repo = { slug, private, default_branch: defaultBranch.name };
  102 |     const data = { branch, commit_message: message };
  103 |     const type = 'api';
```

but also, this whole thing is going to be subject to change. for instance, `config` now also needs to get an empty string as a default value ... i wouldn't know how to change this on my own.

sorta: oh yeah, I forgot that `private` is reserved. For which I know a possible solution, but with the other things you mention, and since this was an optional suggestion in the first place, maybe it's worth revisiting later

app/components/request/configs.js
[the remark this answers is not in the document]
svenfuchs: i have added that question to a new section "to be discussed/resolved" on https://github.com/travis-ci/travis-web/pull/2399 in order to keep track of this

svenfuchs:
i have cherrypicked all commits from this branch onto https://github.com/travis-ci/travis-web/pull/2399.

i am going to close this one here as i think all reviews have been either addressed, or been noted on the "to be discussed/resolved" list on https://github.com/travis-ci/travis-web/pull/2399.

please reopen if needed.

PR #2405, opened 2020-03-14
add repo/config

svenfuchs:
based on https://github.com/travis-ci/travis-web/pull/2403

with this in place the same trigger build/customize/preview functionality is available on the route `/repo/:slug/config`, except it won't default on the current build request's data (sha, branch, etc)

### DIFF
--- a/app/components/request/configs.js
+++ b/app/components/request/configs.js
@@ -111,7 +111,7 @@ export default Component.extend(CanTriggerBuild, TriggerBuild, {
       mode: this.mergeMode,
       config: this.config,
       data: {
-        branch: this.branch,
+        branch: this.refType === 'branch' ? this.branch : null,
         commit_message: this.message
       },
       type: 'api'
--- a/app/components/request/configs.js
+++ b/app/components/request/configs.js
@@ -33,7 +33,7 @@ export default Component.extend(CanTriggerBuild, TriggerBuild, {
   loading: reads('preview.loading'),
 
   refType: 'sha',
-  sha: truncated('originalSha', 10),
+  sha: reads('originalSha'),
   branch: reads('originalBranch'),
   message: reads('request.commit.message'),
   config: reads('request.apiConfig.config'),
--- a/app/components/request/configs.js
+++ b/app/components/request/configs.js
@@ -18,6 +18,12 @@ export default Component.extend(CanTriggerBuild, TriggerBuild, {
 
   preview: service('request-config'),
   features: service(),
+
+  status: 'closed',
+  closed: equal('status', 'closed'),
+  customizing: equal('status', 'customize'),
+  previewing: equal('status', 'preview'),
+  replacing: equal('mergeMode', 'replace'),
   showNewConfigView: reads('features.showNewConfigView'),
 
   repo: reads('request.repo'),
@@ -27,7 +33,7 @@ export default Component.extend(CanTriggerBuild, TriggerBuild, {
   loading: reads('preview.loading'),
 
   refType: 'sha',
-  sha: reads('originalSha'),
+  sha: truncated('originalSha', 10),
   branch: reads('originalBranch'),
   message: reads('request.commit.message'),
   config: reads('request.apiConfig.config'),
--- a/app/controllers/repo/config.js
+++ b/app/controllers/repo/config.js
@@ -1,0 +1,6 @@
+import Controller from '@ember/controller';
+
+export default Controller.extend({
+  queryParams: ['status'],
+  status: 'customize'
+});
--- a/app/router.js
+++ b/app/router.js
@@ -61,6 +61,7 @@ Router.map(function () {
     this.route('build', { path: '/builds/:build_id', resetNamespace: true }, function () {
       this.route('config');
     });
+    this.route('config');
     this.route('job', { path: '/jobs/:job_id', resetNamespace: true }, function () {
       this.route('config');
     });
--- a/app/styles/app/modules/request/buttons.scss
+++ b/app/styles/app/modules/request/buttons.scss
@@ -1,3 +1,11 @@
+.repo-config .request-configs .request-configs-buttons {
+  top: -45px;
+
+  .request-configs-button-cancel {
+    display: none;
+  }
+}
+
 .request-configs .request-configs-buttons {
   position: absolute;
   top: 1px;
--- a/app/templates/build/config.hbs
+++ b/app/templates/build/config.hbs
@@ -1,4 +1,5 @@
 <Request::Configs
+  @repo={{this.model.repo}}
   @request={{this.model}}
   @status={{this.status}}
 />
--- a/app/templates/job/config.hbs
+++ b/app/templates/job/config.hbs
@@ -5,6 +5,7 @@
   </PaperBlock>
 {{/if}}
 <Request::Configs
+  @repo={{this.model.repo}}
   @request={{this.model}}
   @status={{this.status}}
 />
--- a/app/routes/repo/config.js
+++ b/app/routes/repo/config.js
@@ -1,0 +1,5 @@
+import TravisRoute from 'travis/routes/basic';
+
+export default TravisRoute.extend({
+  titleToken: 'Trigger Build',
+});
--- a/app/styles/app/modules/request/configs.scss
+++ b/app/styles/app/modules/request/configs.scss
@@ -1,3 +1,8 @@
+.repo-config {
+  position: relative;
+  margin-top: 3em;
+}
+
 .request-configs {
   .request-configs-matrix-notice {
     border: 1px solid $dry-cement;
--- a/app/templates/components/request/configs.hbs
+++ b/app/templates/components/request/configs.hbs
@@ -66,6 +66,26 @@
     />
   {{/unless}}
 
+  {{#unless this.replacing}}
+    <h4>
+      Build {{pluralize this.rawConfigs.length 'Config' without-count=true}}
+    </h4>
+
+    {{#each this.rawConfigs as |config|}}
+      <Request::RawConfig
+        @source={{config.source}}
+        @config={{get config 'config'}}
+        @loading={{this.loading}}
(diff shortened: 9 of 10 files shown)


app/styles/app/modules/request/buttons.scss
```diff
@@ -1,3 +1,11 @@
+.repo-config .request-configs .request-configs-buttons {
+  top: -45px;
```
sorta: The way that I would normally approach this situation would be to do something like 
```
.repo-config .request-configs .request-configs-buttons {
  position: relative;
  text-align: right;
```

This way, the changes in this PR to `app/styles/app/modules/request/configs.scss` become unnecessary. The way that it's currently set up, it essentially emulates relative positioning anyway, so I think it's more straightforward to just use relative position.

app/components/request/configs.js
```diff
@@ -18,6 +18,12 @@ export default Component.extend(CanTriggerBuild, TriggerBuild, {
 
   preview: service('request-config'),
   features: service(),
+
+  status: 'closed',
+  closed: equal('status', 'closed'),
+  customizing: equal('status', 'customize'),
+  previewing: equal('status', 'preview'),
```
artursmirnov: We usually strive to extract dictionary values to constants and re-use them whenever possible, to avoid hardcoding and simplify potential refactoring, e.g.:
```js
export const STATUSES = {
  CLOSED: 'closed',
  CUSTOMIZE: 'customize',
  PREVIEW: 'preview'
};

  status: STATUSES.CLOSED,
  closed: equal('status', STATUSES.CLOSED),
  customizing: equal('status', STATUSES.CUSTOMIZE),
  previewing: equal('status', STATUSES.PREVIEW),
})
```
How does it sound to you?

app/controllers/repo/config.js
```diff
@@ -0,0 +1,6 @@
+import Controller from '@ember/controller';
+
+export default Controller.extend({
+  queryParams: ['status'],
+  status: 'customize'
```
artursmirnov: If it's the same `status` as in component, then it'd be better to:

```js
import { STATUSES } from 'travis/components/request/configs'

  status: STATUSES.CUSTOMIZE
})
```
In this case it'll always keep in sync and it won't be needed to find/replace hardcoded strings.

app/templates/build/config.hbs
```diff
@@ -1,4 +1,5 @@
 <Request::Configs
+  @repo={{this.model.repo}}
```
artursmirnov: It seems like this value can be just derived from `@request`, isn't it?

app/templates/job/config.hbs
```diff
@@ -5,6 +5,7 @@
   </PaperBlock>
 {{/if}}
 <Request::Configs
+  @repo={{this.model.repo}}
```
artursmirnov: Same here, the value can be derived from `@request`

app/router.js
```diff
@@ -61,6 +61,7 @@ Router.map(function () {
     this.route('build', { path: '/builds/:build_id', resetNamespace: true }, function () {
       this.route('config');
     });
+    this.route('config');
```
artursmirnov: I'd call this route a bit more semantically, something like `trigger-build`. Currently it reads as "Repository Config", which, as far as I understand, it isn't.

app/templates/build/config.hbs
svenfuchs: i believe i thought i'd need to add it because now that same view is also shared on the `repo` route (see below), where there is no request.

app/router.js
svenfuchs: i see your concern.

"config" certainly is an overloaded term, but it's more than "just" triggering builds. it's also about trying out and previewing request configs (potentially adding more functionality in the future).

how about we make it a nested route `repo/:slug/request/config`? i'm not sure if `repo/:slug/build/config` is taken by the build id.

app/controllers/repo/config.js
svenfuchs: nice! i didn't know i could do that ... seems obvious now :)

curiously, on the final branch this status is entirely gone. i believe this may have been a left over from previous refactoring, and i may have deleted it in the meantime.

app/components/request/configs.js
svenfuchs: ok, done. i've also then reused the constants in the code further down (in various `on*` handlers)

however, all this super-all-caps code now is a little hard on the eyes for my taste. i'll gladly keep it if this is the right JS or Ember convention.

as a bit of an alleviation ... would it be ok to use a singular `STATUS` constant and keep the keys lowercase? for example `STATUS.closed`? i usually keep such constants singular in my code because i only ever refer to one single status after all.

app/styles/app/modules/request/buttons.scss
svenfuchs: i'll try that out. usually i first try things without absolute positioning. i have forgotten what the issue was that made me think i'd need to use it here.

svenfuchs: right ... i think the issue is that the `nav` element that holds the two tabs on the left side has `display: block`, so with `position: relative` on the next element that element is going to be pushed down to the next "line". 

image

even if i then move it back up using `top: -45px` that will then leave a gap:

image

unless we want to change the style of the tabs nav element (which i'm not sure would be preferable) i think the only way to fix this is to position the buttons absolutely, isn't it?

app/components/request/configs.js
artursmirnov: @svenfuchs thank you, it looks great to me 👍 This is not a particular JS or Ember convention, just a common good practice that we use across the app. Using singular form for dictionary name is totally fine with me 👍 The names are usually totally up to developer. Although, for consistency I'd prefer keeping the keys uppercase as well. I think it's just a matter of preference and habit - with all-caps you can easily find/see constant and dictionary values even briefly scanning a file. Does it make sense to you?

app/templates/build/config.hbs
artursmirnov: Yeah, I've noticed that as well. Internally in component it could be an alias:

```js
  request: null,
  repo: reads('request.repo')
});
```
And then you can overwrite it on invocation:
```hbs
<Request::Configs
  @repo={{this.repo}}
>
```
With this invocation, external attribute will take precedence over internal property and overwrite it.

app/router.js
artursmirnov: Do you know what kind of functionality is going to be added there in the future? Probably we can find a name based on that? The nested route sounds good to me, but I'd rather suggest something like `repo/:provider/:owner/:name/requests/new/config` - as we have already `requests` route in plural form and adding one in singular form looks a bit inconsistent. What do you think?

app/styles/app/modules/request/buttons.scss
sorta: Sorry, I think I didn't explain what I was proposing well enough :) Yes that's absolutely an appropriate use of absolute position, and I didn't intend to suggest getting rid of it entirely (though now I totally see how that came across). What I meant was to override it in the `.repo-config` context, which doesn't have those tabs. You're already using the selector with that class at the front to override the `top` declaration, and what I'm proposing is to instead use it to override the `position` declaration. So the top of this file would look like this:

```
.repo-config .request-configs .request-configs-buttons {
  position: relative;
  text-align: right;

  .request-configs-button-cancel {
    display: none;
  }
}

.request-configs .request-configs-buttons {
  position: absolute;
  top: 1px;
  right: 0;
}
```
With the `.request-configs .request-configs-buttons` styles being left alone. And then in `app/styles/app/modules/request/configs.scss` remove
```
.repo-config {
  position: relative;
  margin-top: 3em;
}
```
Does it make sense?

app/router.js
svenfuchs: i think essentially this could be a "home" place for everything around configuring your build. trying out how matrix expansion works (using the preview feature), trying out how configs merge into each other (for either api requests or config imports), actually submitting those as build requests (i.e. "trigger build").

one bigger idea is to add a "build config wizard" which asks you for your language, and then from there takes you through a bunch of questions, helping newcomers build up their `.travis.yml` file.

obviously this is all pretty unknown today, but i think it would be good to keep the route as generic as possible. even the `new` in `requests/new/config` seems rather specific ...

having said that i am still undecided and going back and forth on what i think the best route naming would be ... 🤔

app/styles/app/modules/request/buttons.scss
svenfuchs: oh, ok, i see now! yes, i'll give that a show.

svenfuchs: done. see here: https://github.com/travis-ci/travis-web/pull/2407/commits/ae78866c739f8786ee0b0eb357abe4bbef1122cb

app/templates/build/config.hbs
svenfuchs: oh, i see ... yes, that makes sense.

svenfuchs: lol, somehow that's already been taken care of in the final branch. see https://github.com/travis-ci/travis-web/blob/sf-request-config-tweaks/app/templates/build/config.hbs and https://github.com/travis-ci/travis-web/blob/sf-request-config-tweaks/app/templates/job/config.hbs

app/styles/app/modules/request/buttons.scss
sorta: Looks good, thanks! :rocket:

app/router.js
svenfuchs: i have added that question to a new section "to be discussed/resolved" on https://github.com/travis-ci/travis-web/pull/2399 in order to keep track of this

svenfuchs:
i have cherrypicked all commits from this branch onto https://github.com/travis-ci/travis-web/pull/2399.

i am going to close this one here as i think all reviews have been either addressed, or been noted on the "to be discussed/resolved" list on https://github.com/travis-ci/travis-web/pull/2399.

please reopen if needed.

PR #2409, opened 2020-03-16
bb enable

piotr-travisci:


config/environment.js
```diff
@@ -385,6 +385,7 @@ module.exports = function (environment) {
   }
 
   if (environment === 'production') {
+    ENV.featureFlags['enableBitbucketLogin'] = true;
```
sorta: I think it would be better to enable on the intial introduction of the flag, around line 193. We'll want to enable it for staging as well, and if needed it can be easily overridden by env vars using the logic at about line 223

artursmirnov: This is just a quick hack-ish fix to get the BB login working on production asap. It appeared that env feature flags aren't working on production, because waiter does not pass them to config. I'm currently working on the fix, but it may take some time 😕

PR #2410, opened 2020-03-16
Fix feature flags env var for waiter

artursmirnov:
This change makes `ENABLE_FEATURE_FLAGS` variable work in `waiter` environment.


artursmirnov:
Thanks everyone for reviews 🎉

PR #2411, opened 2020-03-16
BB Landing page - Scroll to top on page load

sorta:
Addresses an issue noted by Petra. To Reproduce:

1. Make sure you are signed out
2. Go to `/`
3. Scroll down to the Learn More button in the bitbucket section. The issue is more noticeable the further you scroll
4. Click the Learn More button
5. It will correctly lead to `/integration/bitbucket`, but you will be partially scrolled down on the page

Compare:
https://travis-ci.com/
https://so-fix-bb-landing-scroll.test-deployments.travis-ci.com/

Although, I'm not sure this is much better. Let me know what you think

### DIFF
--- a/app/components/scroll-here.js
+++ b/app/components/scroll-here.js
@@ -4,8 +4,11 @@ import { inject as service } from '@ember/service';
 export default Component.extend({
   scroller: service(),
 
+  immediate: false,
+
   didInsertElement() {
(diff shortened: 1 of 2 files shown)


sorta:
That's a good idea! The animation is a little distracting and unnecessary.  I'll add that in :)

PR #2414, opened 2020-03-19
Fix build status image update issue

sorta:
For https://travisci.assembla.com/spaces/VCS/tickets/145/details

I've been testing like this:
1. Create a new repo on provider with simple tests. A basic ember project works well, and to make this faster I've just been forking one that I already have.
2. Activate repo on travis if needed
3. Go to the repo page and trigger a build. Watch as the build completes, observing the status button.
4. Flip through tabs on completion

This regularly reproduces the issue on prod for me. I think it may be a caching thing. Compare:
https://travis-ci.com/
https://so-fix-build-status-img.test-deployments.travis-ci.com/

Just realized an issue this causes though. Looking more into it.

PR #2416, opened 2020-03-20
Fix build status image update issue

sorta:
For https://travisci.assembla.com/spaces/VCS/tickets/145/details

I've been testing like this:
1. Create a new repo on provider with simple tests. A basic ember project works well, and to make this faster I've just been forking one that I already have.
2. Activate repo on travis if needed
3. Go to the repo page and trigger a build. Watch as the build completes, observing the status button.
4. Flip through tabs on completion

This regularly reproduces the issue on prod for me. I suspect it may be a caching thing, so I added a status-based query param to the image displayed on `travis-web`. Compare:
https://travis-ci.com/
https://so-fix-build-status-img-2.test-deployments.travis-ci.com/

I've seen this for both github and bitbucket, but if you want to test with bb, I would test on staging, which i did deploy to recently
https://staging.travis-ci.com/
For some reason, bitbucket status images seem generally broken on test deployments, the API is returning 403 for them for some reason. This doesn't seem to happen on prod and stag though.

This is an alternative to this PR https://github.com/travis-ci/travis-web/pull/2414 which I will probably close in favor of this one. I realized that my solution on 2414 would change the URL being given to users, not just the URL being used to display the badge on our site, which I don't think we necessarily want.

### DIFF
--- a/app/components/repo-status-badge.js
+++ b/app/components/repo-status-badge.js
@@ -1,0 +1,31 @@
+import Component from '@ember/component';
+import { inject as service } from '@ember/service';
+import { computed } from '@ember/object';
+import { or, reads } from '@ember/object/computed';
+
+export default Component.extend({
+  tagName: '',
+  statusImages: service(),
+
+  repo: null,
+  defaultBuildStateLabel: 'unknown',
+  onClick() {},
+
+  lastBuildState: or('repo.defaultBranch.lastBuild.state', 'defaultBuildStateLabel'),
+  defaultBranch: reads('repo.defaultBranch.name'),
+
+  statusImageUrl: computed(
+    'repo.slug',
+    'repo.private',
+    'defaultBranch',
+    'lastBuildState',
+    function () {
+      const { defaultBranch, lastBuildState = 'unknown', repo } = this;
+
+      const url = this.statusImages.imageUrl(repo, defaultBranch);
+      const divider = (url.indexOf('?') >= 0) ? '&' : '?';
+
+      return `${url}${divider}status=${lastBuildState}`;
+    }
+  ),
+});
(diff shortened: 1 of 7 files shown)


app/components/repo-status-badge.js
```diff
@@ -0,0 +1,30 @@
...
+  statusImageUrl: computed(
+    'repo.slug',
+    'repo.private',
+    'defaultBranch',
+    'lastBuildState',
+    function () {
+      const { defaultBranch, lastBuildState = 'unknown', repo } = this;
+
+      const url = this.statusImages.imageUrl(repo, defaultBranch);
+      const divider = (url.indexOf('?') >= 0) ? '&' : '?';
```
artursmirnov: I think it's pretty safe to use
```suggestion
      const divider = url.includes('?') ? '&' : '?';
```
What do you think?

app/templates/components/repo-status-badge.hbs
```diff
@@ -0,0 +1,11 @@
+<div class="repo-badge inline-block vertical-align">
+  <a
+    title="Latest push build on default branch: {{this.lastBuildState}}"
+    name="status-images"
+    class="pointer open-popup"
+    data-test-status-image-popup="true"
+    {{on 'click' this.onClick}}
```
artursmirnov: As long as it's an external attribute, what do you think about
```suggestion
    {{on 'click' @onClick}}
```
?

app/components/repo-status-badge.js
sorta: Ahh yeah, that sounds good to me :)

app/templates/components/repo-status-badge.hbs
sorta: Totally, thanks for catching it!

sorta:

PR #2419, opened 2020-03-24
fade out at too long branch name

piotr-travisci:


artursmirnov:
Hey @piotr-travisci, thanks for the improvement 👍 It looks good code-wise, but I didn't find any repositories to check it visually. Do you happen to know one?

PR #2422, opened 2020-03-26
GitHub accessSettings from env var

piotr-travisci:
https://travisci.assembla.com/spaces/Support/tickets/68--quot-review-and-add-quot--link-on--com-account-settings-points-to-travis-ci-org

### DIFF
--- a/config/environment.js
+++ b/config/environment.js
@@ -24,8 +24,7 @@ const {
   DISABLE_SENTRY,
   TRAVIS_COMMIT,
   SOURCE_VERSION,
-  DEPLOY_TARGET,
-  GITHUB_ORGS_OAUTH_ACCESS_SETTINGS_URL
+  DEPLOY_TARGET
 } = process.env;
 
 module.exports = function (environment) {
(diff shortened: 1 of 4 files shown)


config/environment.js
```diff
@@ -297,6 +298,10 @@ module.exports = function (environment) {
       ENV.billingEndpoint = BILLING_ENDPOINT;
     }
 
+    if (GITHUB_ORGS_OAUTH_ACCESS_SETTINGS_URL) {
+      ENV.providers.github.paths.accessSettings = GITHUB_ORGS_OAUTH_ACCESS_SETTINGS_URL;
+    }
```
artursmirnov: Wouldn't it be better to set the value in place, where it's defined, instead of suddenly updating it somewhere else? Just thinking that it might be unexpected when debugging. All env vars are available in providers' config as well. Also this would allow to keep the `deepFreeze`. What do you think?

config/providers.js
```diff
@@ -1,8 +1,6 @@
 /* eslint-env node */
 'use strict';
 
-const deepFreeze = require('deep-freeze');
```
artursmirnov: This is the only place where `deep-freeze` is used. So if you're removing it from here, probably it'd make sense to remove the dependency at all. Although, to be honest, I'd prefer to keep it here (and for other configs eventually).

artursmirnov:
@piotr-travisci would it make sense to also add this env variable handling to Waiter?

piotr-travisci:
You are right Artur. Thank you for suggestions!

PR #2423, opened 2020-03-26
Add new skip_[type] messages

svenfuchs:
This adds 6 new message types that Yml now produces when things are being skipped via `if` conditions.

It also moves the `<code>` tag to the `format` function in order to reduce the work needed to add such new messages.

### DIFF
--- a/app/components/build-message.js
+++ b/app/components/build-message.js
@@ -18,163 +18,163 @@ export default Component.extend({
     if (this[code]) {
       return htmlSafe(this[code](key, args));
     } else {
-      return htmlSafe(`unrecognised message code <code>${escape(code)}</code>`);
+      return htmlSafe(`unrecognised message code ${format(code)}`);
     }
   }),
 
   alias_key(key, args) {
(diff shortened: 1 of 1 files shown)


app/components/build-message.js
```diff
@@ -18,139 +18,163 @@ export default Component.extend({
...
       return htmlSafe(this[code](key, args));
     } else {
-      return htmlSafe(`unrecognised message code <code>${escape(code)}</code>`);
+      return htmlSafe(`unrecognised message code ${format(code)}`);
     }
   }),
 
   alias_key(key, args) {
-    return `<code>${escape(key)}</code>: key <code>${escape(args.alias)}</code> is an alias for <code>${escape(args.key)}</code>, using <code>${escape(args.key)}</code>`;
+    return `${format(key)}: key ${format(args.alias)} is an alias for ${format(args.key)}, using ${format(args.key)}`;
```
sorta: I love the idea of centralizing this kind of thing :+1: There's one thing I'm wondering about though. Many of these variables were previously only going through `escape`, now also go through `dump` and `truncate`. Is it intentional? I'm guessing so, and I checked the output on a few repos and I think it looks good to me, but I felt it was worth asking about.

PR #2425, opened 2020-03-27
Speed up tests requiring signin

sorta:
I noticed that tests have been taking longer to complete lately. This PR is intended to address that.


sorta:
Thanks for the review @artursmirnov ! Yeah it took a while to figure out and at the end I was surprised at how simple the solution ended up being.

PR #2426, opened 2020-03-27
As save utm params to server

artursmirnov:


app/routes/application.js
```diff
@@ -28,13 +31,23 @@ export default TravisRoute.extend(BuildFaviconMixin, {
...
+          const hitCallback = () => {
+            needsReset = false;
+            controller && controller.resetUTMs();
+          };
+
+          metrics.trackPage({ page, hitCallback });
+
+          later(() => {
+            if (needsReset) this.controller.resetUTMs();
+          }, RESET_UTM_PARAMS_DELAY);
```
sorta: I'm not sure if it makes a difference here, but `later` was a part of the test slowdown that I recently looked into. In that case, it seemed that awaiting `settled` or `visit` resulted in trying to wait for the `later` call to clear. But I also noticed that you put an explicit timeout after settled in one of your tests here, so maybe it's not happening in this case.

If it does seem like it may be an issue in this case, we could define the delay as an `interval` in `config/environment.js`, and then set a lower number (or 0) in the test environment, as has been done for a few other delays/intervals.

artursmirnov: That's a great point! Thank you @sorta, I'll add an interval to the config and set it to `0` for tests 👍

app/adapters/user.js
```diff
@@ -1,14 +1,14 @@
...
 export default V3Adapter.extend({
   buildURL: function (modelName, id, snapshot, requestType, query) {
-    if (requestType === 'queryRecord' && query.current === true) {
-      let prefix = this.urlPrefix();
-      return `${prefix}/user`;
-    } else {
-      throw new Error(`The user adapter only supports a
-        queryRecord request type with a query of current: true`);
-    }
+    const isQueryingCurrentUser = requestType === 'queryRecord' && query.current === true;
```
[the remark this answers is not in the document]
artursmirnov: @emmanuel-ci thank you for the comment! I have double-checked the logic and it seems correct. The adapter only supports fetching a current user or (the new feature) update current user. For fetching current user, the request must be of `queryRecord` type and have `current: true` in query. Thus, when both conditions are `true`, the `isQueryingCurrentUser` will also be `true` and `assert` will pass. Does it make sense?

artursmirnov:
Thank you for reviews, everyone! 🎉 This PR requires https://github.com/travis-ci/travis-api/pull/1080 to be released first, so putting it on hold so far.

PR #2430, opened 2020-03-31
Added Travis Vs Jenkins to list of URLs

pauladamgordon:


config/environment.js
```diff
@@ -101,6 +101,7 @@ module.exports = function (environment) {
       status: 'https://www.traviscistatus.com/',
       support: 'mailto:support@travis-ci.com',
       terms: 'https://docs.travis-ci.com/legal/terms-of-service/',
+      travisvsjenkins: 'https://travis-ci.com/travisci-vs-jenkins',
```
sorta:
```suggestion
      travisVsJenkins: 'https://travis-ci.com/travisci-vs-jenkins',
```
I would suggest camel casing the key here. What do you think?

sorta:
Actually, after thinking about this a bit more, I would like to suggest a different approach that would involve closing this PR. I'll discuss it more in https://github.com/travis-ci/travis-web/pull/2431

pauladamgordon:
Thanks @sorta, makes perfect sense to me!

PR #2431, opened 2020-03-31
Added Travis Ci vs Jenkins to footer

pauladamgordon:
Relies on https://github.com/travis-ci/travis-web/pull/2430

### DIFF
--- a/app/templates/components/page-footer.hbs
+++ b/app/templates/components/page-footer.hbs
@@ -99,8 +99,8 @@
         {{/if}}
         <li>
           <ExternalLinkTo
-            data-test-footer-travisvsjenkins-link="true"
-            @href={{this.config.urls.travisvsjenkins}}
+            data-test-footer-travis-vs-jenkins-link="true"
+            @href={{this.config.urls.travisVsJenkins}}
             @title="Travis CI vs Jenkins"
           >
             Travis CI vs Jenkins
--- a/app/templates/components/page-footer.hbs
+++ b/app/templates/components/page-footer.hbs
@@ -97,6 +97,15 @@
             </LinkTo>
           </li>
         {{/if}}
+        <li>
+          <ExternalLinkTo
+            data-test-footer-travisvsjenkins-link="true"
+            @href={{this.config.urls.travisvsjenkins}}
+            @title="Travis CI vs Jenkins"
+          >
+            Travis CI vs Jenkins
+          </ExternalLinkTo>
+        </li>
       {{/unless}}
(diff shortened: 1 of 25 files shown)


app/templates/components/page-footer.hbs
```diff
@@ -97,6 +97,15 @@
             </LinkTo>
           </li>
         {{/if}}
+        <li>
+          <ExternalLinkTo
```
sorta: Is it intentional to use `ExternalLinkTo` rather than `LinkTo`? The jenkins page is an internal route, so normally it would make more sense to just use `LinkTo`. And then we wouldn't need to define the URL in config. BUT since the jenkins page doesn't show on `org`, perhaps you're using the external link to allow the `org` environment to link to the page. In which case, that makes a lot of sense.

But now I'm thinking if that's the case, a better approach might be to allow the jenkins page to be seen on `org` and use `LinkTo`. Although IIRC, one concern about that was duplicate content causing SEO issues, so maybe not. WDYT?

artursmirnov:
@pauladamgordon just a tip - when you'd like to edit multiple files in GitHub, there's a wonderful tool: https://gitpod.io/ - it's basically an in-browser IDE that integrates seamlessly with GitHub. Using this tool it's possible to edit any files in a repository and push them back to GitHub.

app/templates/components/page-footer.hbs
pauladamgordon: @sorta, happy to go with whatever works best for you - just an old dev in a marketing body trying to make things work!

app/templates/components/page-footer.hbs
```diff
@@ -97,6 +97,15 @@
...
+            data-test-footer-travisvsjenkins-link="true"
+            @href={{this.config.urls.travisvsjenkins}}
```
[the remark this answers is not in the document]
Montana: @sorta I prefer the camelcase, but also understand the extra tedious work that would go into renaming the classes.

app/templates/components/page-footer.hbs
sorta: Gotcha :smile: I think we can stick with the external link for now. :+1:

app/templates/components/page-footer.hbs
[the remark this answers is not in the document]
sorta: Yeah totally. I think I have a slight preference for snake case, but mostly I try to use whatever is most prevalent in the codebase I'm working on.

PR #2441, opened 2020-04-07
Improve consistency of running jobs tab data

sorta:
For https://travisci.assembla.com/spaces/Web/tickets/295/details

The issue seems more likely to show up with builds that have many jobs. I have seen it a number of times with `travis-ci/cpython-builder`, `travis-ci/php-src-builder`, and `travis-ci/travis-rubies`. An easy way to test is to compare the test deployment and production while such builds are going on. But I have also been running other builds on other repos to see how they progress.

A  telltale sign of the issue is a green or red job in the list. Finished jobs should not show, or be counted.


sorta:
Actually, maybe hold off on reviews for now. I want to do some additional testing

sorta:
ok I did some additional testing and I think this is actually ready for review now when you get a chance @artursmirnov @piotr-travisci @emmanuel-ci 

Thanks!

app/components/repository-sidebar.js
```diff
@@ -116,24 +82,19 @@ export default Component.extend({
...
+  tab: reads('tabStates.sidebarTab'),
+  isTabRunning: equal('tab', SIDEBAR_TAB_STATES.RUNNING),
+  isTabSearch: equal('tab', SIDEBAR_TAB_STATES.SEARCH),
 
-  repositoryResults: computed('tab', 'repositories.{searchResults.[],accessible.[]}', function () {
-    let tab = this.tab;
-    let searchResults = this.get('repositories.searchResults.[]');
-    let accessible = this.get('repositories.accessible.[]');
-    let results = accessible;
+  repositoryResults: computed('isTabSearch', 'repositories.{searchResults.[],accessible.[]}', function () {
```
artursmirnov, line 89: I'm not sure if this construction `repositories.{searchResults.[],accessible.[]}` works properly in all cases. Seems a bit unreliable to me. But maybe that's just me 😄

app/services/job-state.js
```diff
@@ -1,37 +1,37 @@
...
+export const UNFINISHED_STATES = QUEUED_STATES.concat(RUNNING_STATES);
 
 export default Service.extend({
   store: service(),
 
-  jobs: reads('fetchJobs.lastSuccessful.value'),
-  jobsLoaded: gt('fetchJobs.performCount', 0),
+  jobs: reads('peekJobs.lastSuccessful.value'),
+  jobsLoaded: gt('fetchUnfinishedJobs.performCount', 0),
+  sortedJobs: computed('jobs', function () {
```
artursmirnov, line 16: It should probably have `jobs.@each.number` for proper tracking, shouldn't it? Or you may also use [sort](https://api.emberjs.com/ember/3.12/functions/@ember%2Fobject%2Fcomputed/sort) computed macro, although it's a bit awkward in use :)

app/services/job-state.js
```diff
@@ -1,37 +1,37 @@
...
+    const { jobs } = this;
+    return jobs && jobs.sortBy('number');
+  }),
+
+  runningJobs: filter('sortedJobs.@each.state', (job) => RUNNING_STATES.includes(job.state)),
+  queuedJobs: filter('sortedJobs.@each.state', (job) => QUEUED_STATES.includes(job.state)),
+  unfinishedJobs: computed('queuedJobs.[]', 'runningJobs.[]', function () {
+    const { queuedJobs = [], runningJobs = [] } = this;
```
artursmirnov, line 24: This may be a bit fragile, because if a property value is `null` and not `undefined` (which is an often case for computed props), it won't be assigned default value. The `value || defaultValue` approach works more reliable in this sense 😄

app/components/repository-sidebar.js
sorta: Yeah, it did seem weird to me! I wasn't sure how much I wanted to refactor stuff that isn't directly related to the issue, but I already did a little, so maybe I'll take it a bit further

app/services/job-state.js
```diff
@@ -1,37 +1,37 @@
...
-  runningJobs: filter('jobs', (job) => {
-    let stillToShow = true;
-    if (job.finishedAt) {
-      const jobTime = new Date(job.finishedAt);
-      const nowTime = new Date();
-      stillToShow = !job.build.get('isFinished') || (nowTime.getTime() - jobTime.getTime()) <= DISPLAY_TIMEOUT_SEC;
-    }
-    return RUNNING_AND_FINISHED_STATES.includes(job.state) && stillToShow;
+  peekJobs: task(function* () {
+    return yield this.store.peekAll('job');
```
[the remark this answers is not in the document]
sorta: Yeah that's a good question, that would be very nice!

app/components/repository-sidebar.js
```diff
@@ -50,62 +52,26 @@ export default Component.extend({
...
-      this.router.transitionTo('index');
-    },
-
-    onQueryChange(query) {
-      if (query === '' || query === this.get('repositories.searchQuery')) { return; }
-      this.set('repositories.searchQuery', query);
-      this.get('repositories.showSearchResults').perform();
-    }
+  showRunningJobs: function () {
+    this.tabStates.set('sidebarTab', SIDEBAR_TAB_STATES.RUNNING);
```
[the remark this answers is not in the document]
sorta: I think it's a great idea! Implemented :+1:

app/components/repository-sidebar.js
```diff
@@ -116,24 +82,19 @@ export default Component.extend({
     }
   }),
 
-  tab: alias('tabStates.sidebarTab'),
+  tab: reads('tabStates.sidebarTab'),
+  isTabRunning: equal('tab', SIDEBAR_TAB_STATES.RUNNING),
```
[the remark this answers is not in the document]
sorta: Change made :)

app/components/repository-sidebar.js
sorta: Made some changes

app/services/job-state.js
sorta: Oh yeah, good point. I put in `jobs.@each.number`, but yeah `sort` computed macro is another good option for sure

app/services/job-state.js
sorta: Good point! Change made

sorta:
Thanks much for the review @artursmirnov ! I think I've addressed all your feedback

app/services/job-state.js
```diff
@@ -1,37 +1,38 @@
...
+  sortedJobs: computed('jobs.@each.number', function () {
+    const { jobs } = this;
+    return jobs && jobs.sortBy('number');
+  }),
+
+  runningJobs: filter('sortedJobs.@each.state', (job) => RUNNING_STATES.includes(job.state)),
+  queuedJobs: filter('sortedJobs.@each.state', (job) => QUEUED_STATES.includes(job.state)),
+  unfinishedJobs: computed('queuedJobs.[]', 'runningJobs.[]', function () {
+    const queuedJobs = this.queuedJobs || [];
+    const runningJobs = this.runningJobs || [];
```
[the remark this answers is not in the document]
sorta: Oh yeah that's a good point, I think you're right. Change made :)

sorta:

PR #2443, opened 2020-04-09
Settings tab available for all who has create_env_var, create_cron, c…

piotr-travisci:
…reate_key_pair permissions


tests/acceptance/repo/settings-test.js
```diff
@@ -35,6 +36,14 @@ module('Acceptance | repo settings', function (hooks) {
         admin: true
       }
     });
+
+    const mockPermissionsService = Service.extend({
+      hasPushPermission() {
+        return true;
+      }
+    });
+    stubService('permissions', mockPermissionsService);
```
artursmirnov, line 45: I think it's a bad idea to stub the service here. The purpose of acceptance tests is to verify that the app works properly in real environment. If tests are failing, then something is not working properly. Stubbing the service only hides the error by excluding the problematic area.

piotr-travisci: Many many thanks for your support @artursmirnov !!!

PR #2448, opened 2020-04-10
As fix deprecations

artursmirnov:
This is just a preliminary cleanup of existing deprecations


artursmirnov:
Thank you for review, @sorta 🎉

PR #2450, opened 2020-04-16
Add the dist value in our web UI

sorta:
For https://travisci.assembla.com/spaces/Web/tickets/157/details

I started with just adding os version data into the tooltips. And then I realized that the `arch` column approach where it collapses to '...' might work well in this instance. So then I tried that out and did a little refactoring. What do you think of this approach?

osx version data was gathered from https://docs.travis-ci.com/user/reference/osx/#macos-version

Useful repos to look at:
- https://so-job-list-dist-info.test-deployments.travis-ci.com/github/travis-ci/php-src-builder
- https://so-job-list-dist-info.test-deployments.travis-ci.com/github/travis-ci/cpython-builder
- https://so-job-list-dist-info.test-deployments.travis-ci.com/github/travis-ci/travis-rubies
- https://so-job-list-dist-info.test-deployments.travis-ci.com/github/travis-pro/nmap-check
- https://so-job-list-dist-info.test-deployments.travis-ci.com/github/sorta/headcanon-generator

Some things to check for regressions:
- https://so-job-list-dist-info.test-deployments.travis-ci.com/github/travis-ci/build-stages-demo
- https://so-job-list-dist-info.test-deployments.travis-ci.com/github/travis-ci/travis.rb
- https://so-job-list-dist-info.test-deployments.travis-ci.com/github/ezyang/pytorch-unattached/builds/81412631

test logged in and logged out


sorta:
ahh yeah that is weird behavior @artursmirnov . It looks like it's an existing thing that happens when you're logged out, but yeah it's strange and unnecessary I think. I've made some changes that I think should address it.

sorta:
Excellent, thanks for the review!

PR #2451, opened 2020-04-16
Clean up build config

sorta:
Moves a duplicate `env` key, and adds an explicit `os` key, as suggested by build config validation


sorta:
Thanks for the review!

PR #2452, opened 2020-04-16
Adjust flaky percy snapshots

sorta:
A couple of percy snapshots are incorrectly flagged as changed fairly often. I thought about trying 
 to await settled or something like that for them, but then realized that I'm not sure we want or need snapshots inside of component integration tests. I think it makes more sense to utilize them in acceptance tests. So I removed them.


sorta:
Awesome :smile: Thanks for the review!

PR #2464, opened 2020-04-22
Add gitlab to providers config and sign in page

artursmirnov:
Ticket: https://travisci.assembla.com/spaces/Web/tickets/302/details

This change introduces provider config changes and buttons on Sign In / Sign Up pages, but I wasn't able to check the rest, because API doesn't allow to log in to the app using GitLab yet.


artursmirnov:
Thanks everyone for reviews! 🎉

PR #2471, opened 2020-05-01
Fix community SVG display

sorta:
For https://travisci.assembla.com/spaces/Web/tickets/304/details

Some background info: ID collisions between elements within different SVGs has caused various issues with the display of gradients, colors, etc within SVGs. In the past this was handled with ID changes and other tricks within individual SVGs. More recently, a more universal approach was implemented by changing SVGO settings in `ember-cli-build.js` (namely,  auto-generated ID prefixes was enabled). This now automatically handles these issues for the most part, but seems like it didn't work well with an old fix within the help page "community" svg. Removing the old fix seems to resolve all issues.

I looked to see if there were any other SVGs having a similar issue, but it appears not.

To test, deploy to staging and compare with https://travis-ci.com/help


sorta:
Thanks for the review!

PR #2475, opened 2020-05-05
Gitlab behind feature flag

piotr-travisci:


app/templates/components/landing-default-page.hbs
```diff
@@ -48,6 +48,14 @@
                   </span>
                 </button>
               {{/if}}
+              {{#if this.multiVcs.enableGitlabLogin}}
+                <button class="hero-button org multi-vcs" {{action "signIn" "gitlab"}}>
```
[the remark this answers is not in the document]
piotr-travisci: Thank you @emmanuel-ci for your suggestion.
Code refactored.

config/environment.js
```diff
@@ -195,6 +195,7 @@ module.exports = function (environment) {
     'github-apps': false,
     'enable-assembla-login': false,
     'enable-bitbucket-login': false,
+    'enable-gitlab-login': false,
```
sorta, line 198: I think we were hoping to start eliminating `enable-` from the beginning of flag names. All feature flags are about enabling or disabling something, so it is somewhat redundant. And the `gitlab-login` flag already existed below.

app/templates/components/landing-pro-page.hbs
```diff
@@ -52,6 +52,28 @@
       </section>
       <div class="spacer"></div>
     {{/if}}
+    {{#if this.multiVcs.enableGitlabLogin}}
+      <section class="section-gitlab section--white section--center">
+        <div class="inner">
+          <h2 class="text-5xl text-grey-800 mb-8">
+            Now test your
+            <SvgImage @name='gitlab-full' @width={{220}} />
```
sorta, line 60: Just noticed that there is no `gitlab-full` svg, so no image displays here and it looks odd:
image

Additionally, we may want to consult with Creative and/or Product teams about content for this part. As it is right now it feels a little repetitive. It might be nice to combine the two sections into one somehow, although it would be a little more complex to implement.

config/environment.js
piotr-travisci: You are right Sophie. As it's now only for testing purposes I will leave it as it is and we will clean it before GL release.

app/templates/components/landing-pro-page.hbs
piotr-travisci: Nice catch Sophie. We should discuss it with Creative team.

piotr-travisci: Currently, we are going to hide this section due to Marketing decisions.

config/environment.js
sorta: Cool, that works for me :)

app/templates/components/landing-pro-page.hbs
Nat-ess: If we do decide to bring it back, I would simplify the copy to this:

### Now test your projects with Travis CI
Check-out our feature - now you can sign up to Travis CI using your Gitlab or Bitbucket account.

AlonaHlobina: Hi @sorta , @emmanuel-ci, and @piotr-travisci! Thank you for bringing up the topic! I have a few updates on it:
We are skipping changing the travis-ci.com page in terms of adding GitLab on it. 
It was agreed with the Marketing team that we are going for a silent release for GitLab. It means we need to have signup/in buttons as we have for BB and GB, but we are not mentioning GitLab on the main page.
We are also not doing the separate GitLab landing page as we did for Bitbucket.

PR #2483, opened 2020-05-14
Prevent inactive repos showing in dashboard starred list

sorta:
For https://travisci.assembla.com/spaces/Web/tickets/189/details

I have already resolved one issue mentioned: the one where a deactivated repo shows up in the starred list on one page and not the other

The issue with the builds of deactivated repos showing up in the my builds list is a little bit more involved though. I think it would require an update to API for filtering current user builds (the `/builds` endpoint) by whether or not their repo is active. _Unless_ what we want is for stuff to show up on the dashboard even if the repo is inactive. I think that wouldn't be to difficult, if it's what we want.

So yeah, take a look when you get a chance and let me know what you think would be the best approach.

To Reproduce:
1. Log in and go to the dashboard
2. Star a repo
3. Go to settings
4. Deactivate the repo you starred
5. Go back to dashboard and note differences between the tabs

I believe that this issue doesn't seem to affect repos that come through the Github Apps integration.


sorta:
Good point! That totally makes sense to me. Thanks for the reviews and feedback :smile:

PR #2489, opened 2020-05-21
Replace active model adapter

sorta:
Here's a question: Do we actually need `active-model-adapter`? Most adapters extend the `v3` adapter, which itself extends `RESTAdapter`. It doesn't seem super actively maintained, and I'm not sure that we utilize it all that much. Thoughts?


artursmirnov:
@sorta I think, it's a great idea! Just one other place that uses `ApplicationAdapter` is https://github.com/travis-ci/travis-web/blob/0068883a0f259a7c1a9c1cd8890fc5d6fb9afb41/app/adapters/ssh-key.js#L3 
Not sure if it needs any changes though :)

sorta:
Thanks! I did some testing and the ssh keys did need a few changes. That should be fixed now and it should ready for review :)

sorta:
Excellent, thanks for the review!

PR #2491, opened 2020-05-28
Redirect->signin in case of signout in another tab

piotr-travisci:


app/services/auth.js
```diff
@@ -83,6 +83,29 @@ export default Service.extend({
 
   redirectUrl: null,
 
+  init() {
+    this._super(...arguments);
+    window.addEventListener('focus', this.checkAuth.bind(this));
```
artursmirnov, line 88: This `bind` signature looks a little bit awkward to me :) I'd suggest an arrow function instead:
```suggestion
    window.addEventListener('focus', () => this.checkAuth());
```
Just an optional suggestion though.

app/services/auth.js
```diff
@@ -83,6 +83,29 @@ export default Service.extend({
...
+      return;
+    const { accounts } = this.storage;
+    const { vcsId } = this.currentUser;
+    let stillLoggedIn = false;
+
+    accounts.map((user) => {
+      if (user.vcsId === vcsId) {
+        stillLoggedIn = true;
+      }
+    });
```
artursmirnov, line 102: There's a simpler way to do this with Ember:
```suggestion
    const stillLoggedIn = accounts.isAny('vcsId', vcsId);
```
Here's more info: https://api.emberjs.com/ember/3.12/classes/EmberArray/methods/isAny?anchor=isAny

app/services/auth.js
```diff
@@ -83,6 +83,29 @@ export default Service.extend({
...
+
+    if (!stillLoggedIn) {
+      this.router.transitionTo('signin');
```
artursmirnov, line 105: Are you sure, we don't want to run the full `signOut` with all teardowns and cleanups?

app/services/auth.js
piotr-travisci: You are right. Arrow functions look a way more trendy :-)

app/services/auth.js
piotr-travisci: Awesome, thank you @artursmirnov !

app/services/auth.js
piotr-travisci: This is already performed in another tab - this tab, which signed out before.

PR #2493, opened 2020-05-29
Remove heroku redis urls

r-arek:
Travis web PR deployments are using hobby-dev instance of redis on heroku. URL of these instances aren't constant which causes errors in travis-web-next app. 
I've prepared redis instance with static IP and would like to set redis urls via travis job env vars.

### DIFF
--- a/config/deployment/deploy-local.sh
+++ b/config/deployment/deploy-local.sh
@@ -2,8 +2,6 @@
 
 export CLEANED_BRANCH_SUBDOMAIN=`git name-rev --name-only HEAD | tr '.' '-' | tr '/' '-' | tr '[:upper:]' '[:lower:]'`
 
-eval $(./config/deployment/store-redis-urls.sh)
-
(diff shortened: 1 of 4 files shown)


artursmirnov:
@r-arek thank you for taking care of the code update! Could you please add new Redis urls as secret env variables here: https://travis-ci.com/github/travis-ci/travis-web/settings
- `ORG_PRODUCTION_REDIS_URL` - for .org redis
- `COM_PRODUCTION_REDIS_URL` - for .com redis

This should make deployments work properly.

r-arek:
Done ✅

artursmirnov:
@r-arek thank you! Although, something is still wrong. The build (`Travis CI - Pull Request` is the one that deploys) has passed, but the deployed apps throw `500`: https://gcp-redis.test-deployments.travis-ci.com/ and https://gcp-redis.test-deployments.travis-ci.org/. Could you please take a look?

r-arek:
applied new redis url in app envs, everything should work correctly now

PR #2500, opened 2020-06-16
Increase limit for branches search request to include more results

artursmirnov:


artursmirnov:
Thank you for the review @piotrmilcarz!

PR #2507, opened 2020-06-19
ITC-35 |  Job prioritize issue

Rahulgarg30591:
In this ticket, the issue raised for the missing prioritize button on a job view is fixed.

Kindly Review
Rahul Garg


app/models/job.js
```diff
@@ -39,6 +39,7 @@ export default Model.extend(DurationCalculations, DurationAttributes, {
   queue: attr(),
   state: attr(),
   number: attr(),
+  permissions: alias('build.permissions'),
```
artursmirnov, line 42: No, this approach is incorrect. Job response has it's own `@permissions` hash, independent from the `build`'s one:

image

It isn't used at the moment, that's why there isn't a property on the `job` model, but most likely, it will be at some point. 

Also `prioritize` permission is set on `build` level, not on `job`.

I'd recommend tweaking this property to cover both cases: https://github.com/travis-ci/travis-web/blob/epic-fast-track/app/components/repo-actions.js#L67

For example, you could extract `item.permissions.prioritize` to a separate property (e.g. `hasPrioritizePermission` and make it check for permission on both `build` and `job`:
```
hasPrioritizePermission: or('item.permissions.prioritize', 'item.build.permissions.prioritize')
```
Does it make sense to you?

Rahulgarg30591: Sure @artursmirnov sure, implemented the same way now, kindly review the same thank you.

app/components/repo-actions.js
```diff
@@ -1,7 +1,7 @@
 import Component from '@ember/component';
 import { inject as service } from '@ember/service';
 import { computed } from '@ember/object';
-import { alias, and, not, or } from '@ember/object/computed';
+import { alias, and, not, or, equal } from '@ember/object/computed';
```
artursmirnov, line 4: This change seems unnecessary

Rahulgarg30591: Done, Rest @artursmirnov a small fix is also provided in this PR, which is to take care of priority status after the success of prioritizing build.

app/components/dialogs/prioritize-build-modal.js
```diff
@@ -31,6 +31,9 @@ export default Component.extend({
         yield record.increasePriority(this.shouldCancelRunningJobs);
         this.flashes.success('The build was successfully prioritized.');
         this.set('isOpen', false);
+
+        const targetBuild = this.build || this.job.build;
+        targetBuild.set('priority', true);
```
artursmirnov, line 36: That may put the app into inconsistent state. For example, if the request have silently failed to update the priority on backend, we'll have it `false` on backend and `true` on frontend. I'd recommend `targetBuild.reload()` to update the real state from backend.

Rahulgarg30591: Done.

Rahulgarg30591: Verified as well, as it is working as expected.

artursmirnov: Did you forget to push your changes?

PR #2515, opened 2020-06-25
ITC-35 IE 11 Bug Fix

Rahulgarg30591:
This PR is a design fix with respect to the open ticket: [Ticket 10](https://travisci.assembla.com/spaces/Build-Priorities_PRD/tickets/10-ie--ui-issue-----quot-prioritize-build-quot--icon-not-aligned-properly-in-the/details) as the complete UI is working as expected on other browsers but not on the IE 11 browser as it is unable to accommodate the extra prioritize button.

Kindly Review
Rahul Garg

### DIFF
--- a/app/components/repo-actions.js
+++ b/app/components/repo-actions.js
@@ -68,8 +68,8 @@ export default Component.extend({
   hasPrioritizePermission: or('item.permissions.prioritize', 'item.build.permissions.prioritize'),
   insufficientPermissions: not('userHasPushPermissionForRepo'),
   tooltips: or('labelless', 'mobilelabels'),
-  showSmallIcons: not('tooltips'),
-  canPrioritize: and('item.notStarted', 'isNotAlreadyHighPriority', 'hasPrioritizePermission', 'showSmallIcons'),
+  noSmallIcons: not('tooltips'),
+  canPrioritize: and('item.notStarted', 'isNotAlreadyHighPriority', 'hasPrioritizePermission', 'noSmallIcons'),
 
   cancel: task(function* () {
     let type = this.type;
--- a/app/components/repo-actions.js
+++ b/app/components/repo-actions.js
@@ -67,9 +67,9 @@ export default Component.extend({
   isHighPriority: or('item.priority', 'item.build.priority'),
   isNotAlreadyHighPriority: not('isHighPriority'),
   hasPrioritizePermission: or('item.permissions.prioritize', 'item.build.permissions.prioritize'),
+  canPrioritize: and('item.notStarted', 'isNotAlreadyHighPriority', 'hasPrioritizePermission'),
   insufficientPermissions: not('userHasPushPermissionForRepo'),
   tooltips: or('labelless', 'mobilelabels'),
-  canPrioritize: and('item.notStarted', 'isNotAlreadyHighPriority', 'hasPrioritizePermission'),
 
   cancel: task(function* () {
     let type = this.type;
--- a/app/components/repo-actions.js
+++ b/app/components/repo-actions.js
@@ -60,15 +60,6 @@ export default Component.extend({
   }),
 
   showPriority: true,
-
-  // showPriorityBtn: computed('showPriority', function () {
-  //   if (this.showPriority === false) {
-  //     return false;
-  //   }
-
-  //   return true;
-  // }),
-
   showPrioritizeBuildModal: false,
   canCancel: and('userHasPullPermissionForRepo', 'item.canCancel'),
   canRestart: and('userHasPullPermissionForRepo', 'item.canRestart'),
--- a/app/components/repo-actions.js
+++ b/app/components/repo-actions.js
@@ -59,13 +59,15 @@ export default Component.extend({
     }
   }),
 
-  showPriorityBtn: computed('showPriority', function () {
-    if (this.showPriority === false) {
-      return false;
-    }
+  showPriority: true,
 
-    return true;
-  }),
+  // showPriorityBtn: computed('showPriority', function () {
+  //   if (this.showPriority === false) {
+  //     return false;
+  //   }
+
+  //   return true;
+  // }),
 
   showPrioritizeBuildModal: false,
   canCancel: and('userHasPullPermissionForRepo', 'item.canCancel'),
@@ -76,7 +78,7 @@ export default Component.extend({
   hasPrioritizePermission: or('item.permissions.prioritize', 'item.build.permissions.prioritize'),
   insufficientPermissions: not('userHasPushPermissionForRepo'),
   tooltips: or('labelless', 'mobilelabels'),
-  canPrioritize: and('item.notStarted', 'isNotAlreadyHighPriority', 'hasPrioritizePermission', 'showPriorityBtn'),
+  canPrioritize: and('item.notStarted', 'isNotAlreadyHighPriority', 'hasPrioritizePermission', 'showPriority'),
 
   cancel: task(function* () {
     let type = this.type;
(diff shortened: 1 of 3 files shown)


artursmirnov:
@Rahulgarg30591 thank you for the fix, but could you please give a little bit more context wrt what it aims to fix? Thank you!

Rahulgarg30591:
> @Rahulgarg30591 thank you for the fix, but could you please give a little bit more context wrt what it aims to fix? Thank you!

Hello, @artursmirnov I am sorry for missing the context in the description, now I am adding the same here as well as on the description part. This change is done on behalf of the fix of the ticket: [Ticket 10](https://travisci.assembla.com/spaces/Build-Priorities_PRD/tickets/10-ie--ui-issue-----quot-prioritize-build-quot--icon-not-aligned-properly-in-the/details) as the complete UI was working as expected on other browsers but not on the IE 11 browser as it is unable to accommodate the extra prioritize button.

Rahulgarg30591:
Rest @artursmirnov these are the suggested changes by our designer Ashish Negi(I am unable to tag him here). He will take the points mentioned under consideration.

artursmirnov:
@Rahulgarg30591 @shairyar @AlonaHlobina why do we need prioritize buttons on list of jobs at all? We can't prioritize a job, and for build the button is in the header. Here's what I mean:

image

On this screenshot you can see that we have a "Prioritize build" button in build's actions section (which is expected), but also "Prioritize build" buttons on each job of the build, which doesn't make sense at all. We can't prioritize a job, so why would we make it _look_ as if we can? It will be confusing for users.

As far as I remember from our discussions and the PRD, the "Prioritize build" button should be on the Build header, Job header (for convenience, to be able to prioritize build from Job view), and _build_ lists. Is it correct?

Rahulgarg30591:
> @Rahulgarg30591 @shairyar @AlonaHlobina why do we need prioritize buttons on list of jobs at all? We can't prioritize a job, and for build the button is in the header. Here's what I mean:
> 
> image
> 
> On this screenshot you can see that we have a "Prioritize build" button in build's actions section (which is expected), but also "Prioritize build" buttons on each job of the build, which doesn't make sense at all. We can't prioritize a job, so why would we make it _look_ as if we can? It will be confusing for users.
> 
> As far as I remember from our discussions and the PRD, the "Prioritize build" button should be on the Build header, Job header (for convenience, to be able to prioritize build from Job view), and _build_ lists. Is it correct?

Hello @artursmirnov this is coming by default as the same component is reused, if it is not required then we can remove these extra buttons, rest I also feel like they are not solving much of a purpose and adding confusion.

AlonaHlobina:
When talking about `job_view` we are referring to this screen:
job-view


You are right, we can't prioritize a job. thus when the button is pressed the whole build to which the job associated must be prioritized.

Rahulgarg30591:
> When talking about `job_view` we are referring to this screen:
> job-view
> 
> You are right; we can't prioritize a job. Thus, when the button is pressed, the whole build to which the job associated must be prioritized.

Hello @AlonaHlobina same functionality is implemented, but the question is do we also need to show these buttons on the "Build jobs" tab, which are also prioritizing the build only?

AlonaHlobina:
Right, we can definitely remove these extra buttons.

shairyar:
@artursmirnov I agree that the following is correct

> As far as I remember from our discussions and the PRD, the "Prioritize build" button should be on the Build header, Job header (for convenience, to be able to prioritize build from Job view), and build lists. Is it correct?

Rahulgarg30591:
> Right, we can definitely remove these extra buttons.

Thank you @AlonaHlobina in the same PR I will remove the buttons @artursmirnov may you kindly re-verify it thanks.
cc: @shairyar

artursmirnov:
Thanks everyone for your inputs! @Rahulgarg30591 let's then just remove "Prioritize build" buttons from jobs lists, which would naturally fix the issue, right?

Rahulgarg30591:
> Thanks everyone for your inputs! @Rahulgarg30591 let's then just remove "Prioritize build" buttons from jobs lists, which would naturally fix the issue, right?

Right, @artursmirnov, I have tried to do the same, please check the changes I made now, I have undone the CSS changes.

app/components/repo-actions.js
```diff
@@ -66,9 +66,10 @@ export default Component.extend({
   isHighPriority: or('item.priority', 'item.build.priority'),
   isNotAlreadyHighPriority: not('isHighPriority'),
   hasPrioritizePermission: or('item.permissions.prioritize', 'item.build.permissions.prioritize'),
-  canPrioritize: and('item.notStarted', 'isNotAlreadyHighPriority', 'hasPrioritizePermission'),
   insufficientPermissions: not('userHasPushPermissionForRepo'),
   tooltips: or('labelless', 'mobilelabels'),
+  noSmallIcons: not('tooltips'),
```
artursmirnov, line 71: This is not quite it. As far as I understand, we still wan't to have this button on build lists, so it should not appear only for job. If you'd like to distinguish between full and icon display, please use only `labelless` property. Although, I'd suggest slightly different approach here. As there's a specific place where we'd like to hide these icons, independently of the state, it would be much more reliable to just add another attribute to the component, call it something like `showPriorityButton: true` (true by default, because by default we want the buttons to show), and then just set this property to false only where we want them to be hidden - particularly, in jobs list, e.g.
```hbs
<RepoActions
  @job=this.job
  @labelles=true
  @showPriority=false
/>
```
And then in template you can show the button when `{{#if and(this.canPrioritize this.showPriority)}}` - and that's it :)

How does it sound to you?

Rahulgarg30591: Hello, @artursmirnov I get it, I have made the changes accordingly and verified the working as well. May you kindly check the same.

app/components/repo-actions.js
```diff
@@ -59,16 +59,24 @@ export default Component.extend({
     }
   }),
 
+  showPriorityBtn: computed('showPriority', function () {
+    if (this.showPriority === false) {
+      return false;
+    }
+
+    return true;
+  }),
```
artursmirnov, line 68: I don't see any need in this property. It could just initialize with default:
```js
  showPriority: true,
```

Then external assignment will just overwrite this property with a new value.

Does it make sense to you?

app/components/repo-actions.js
```diff
@@ -59,16 +59,24 @@ export default Component.extend({
...
   canCancel: and('userHasPullPermissionForRepo', 'item.canCancel'),
   canRestart: and('userHasPullPermissionForRepo', 'item.canRestart'),
   canDebug: and('userHasPushPermissionForRepo', 'item.canDebug'),
   isHighPriority: or('item.priority', 'item.build.priority'),
   isNotAlreadyHighPriority: not('isHighPriority'),
   hasPrioritizePermission: or('item.permissions.prioritize', 'item.build.permissions.prioritize'),
-  canPrioritize: and('item.notStarted', 'isNotAlreadyHighPriority', 'hasPrioritizePermission'),
   insufficientPermissions: not('userHasPushPermissionForRepo'),
   tooltips: or('labelless', 'mobilelabels'),
+  canPrioritize: and('item.notStarted', 'isNotAlreadyHighPriority', 'hasPrioritizePermission', 'showPriorityBtn'),
```
artursmirnov, line 79: If you add a visibility property here, the property label becomes irrelevant. It doesn't indicate if user `canPrioritize` the build anymore, it does more. Please either update the property name or just combine two properties in template by `and()` helper - whatever you prefer.

artursmirnov: The reason why property names are important is that it may significantly confuse developers during debugging. Names must be explicit and descriptive.

Rahulgarg30591: Got it. May you please verify now?

PR #2520, opened 2020-07-09
Change default provider handling

artursmirnov:
I have ended up with a slightly different solution rather than just disabling Github via env variable. With the new solution you should be able to re-define default provider via `DEFAULT_PROVIDER` environment variable. Default provider will always be enabled. Other providers are optional and if they aren't explicitly enabled via feature flag, they won't be available for login.

### DIFF
--- a/config/providers.js
+++ b/config/providers.js
@@ -3,6 +3,11 @@
 
 const deepFreeze = require('deep-freeze');
 
+const {
+  GITHUB_ORGS_OAUTH_ACCESS_SETTINGS_URL,
+  DEFAULT_PROVIDER
+} = process && process.env || {};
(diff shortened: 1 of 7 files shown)


config/providers.js
```diff
@@ -3,6 +3,11 @@
 
 const deepFreeze = require('deep-freeze');
 
+const {
+  GITHUB_ORGS_OAUTH_ACCESS_SETTINGS_URL,
```
eugene-travis, line 7: is it github specific?

artursmirnov: Yes, this is how we pass the URL to Github access settings URL for a Github App.

PR #2523, opened 2020-07-14
Add default provider config to waiter

artursmirnov:
This change should allow to set default provider via waiter in compiled app. @eugene-travis could you please check if it works in your setup? Thanks!


artursmirnov:
Thanks for checking it @eugene-travis 👍

PR #2527, opened 2020-08-14
Warning banner in case of negative v2 subscription balance

piotrmilcarz:


app/components/profile-nav.js
```diff
@@ -61,4 +61,21 @@ export default Component.extend({
...
+    const plansPath = isUser ? `https://travis-ci.com/account/${config.planSuffix}` : `https://travis-ci.com/organizations/${login}/${config.planSuffix}`;
+    const settingsPath = isUser ? `https://travis-ci.com/account/${config.settingsSuffix}` : `https://travis-ci.com/organizations/${login}/${config.settingsSuffix}`;
+
+    if (allowance && !allowance.public_repos) {
+      this.flashes.warning(`Builds have been temporarily disabled for public repositories due to a negative credit balance. \
+                            Please go to the <a href="${plansPath}">Plan page</a> to replenish your credit balance or alter your \
+                            <a href="${settingsPath}">OSS Credits consumption setting</a>`);
+    } else if (allowance && !allowance.private_repos) {
+      this.flashes.warning(`Builds have been temporarily disabled for private repositories due to a negative credit balance. \
+                            Please go to the <a href="${plansPath}">Plan page</a> to replenish your credit balance`);
```
artursmirnov, line 77: These are huge HTML blobs being passed to a service method, which doesn't look right to me. I'd recommend you to implement these messages as templates and then display by the service. For example, just an idea - create a new type of messages: `custom` (similar to `warning` that you've created), which would receive a template name as a parameter along with the context object that provides the data required to properly render the template.

app/components/repository-layout.js
```diff
@@ -36,5 +38,19 @@ export default Component.extend({
...
+    const isUser = repo.ownerType === 'user';
+    const login = repo.ownerName;
+    const plansPath = isUser ? `https://travis-ci.com/account/${config.planSuffix}` : `https://travis-ci.com/organizations/${login}/${config.planSuffix}`;
+    const settingsPath = isUser ? `https://travis-ci.com/account/${config.settingsSuffix}` : `https://travis-ci.com/organizations/${login}/${config.settingsSuffix}`;
+
+    if (!repo.canOwnerBuild) {
+      this.flashes.warning(`Builds have been temporarily disabled for this repository due to a negative credit balance. \
+                            Please go to the <a href="${plansPath}">Plan page</a> to replenish your credit balance or alter your \
+                            <a href="${settingsPath}">OSS Credits consumption setting</a>`);
+    }
```
artursmirnov, line 54: Same here, please implement this notification as a normal template.

artursmirnov:
Also Percy reports about this regression: https://storage.googleapis.com/percy-images/8705a90ab8f9b6e724b63ce954eb9ae5b0fb913dc3d90da4450441835e87e5aa

app/components/flash-display.js
```diff
@@ -1,15 +1,21 @@
...
 export default Component.extend({
   flashes: service(),
 
-  classNames: ['flash'],
+  classNameBindings: ['type'],
   tagName: 'ul',
 
   messages: alias('flashes.messages'),
 
+  type: computed('messages', function () {
```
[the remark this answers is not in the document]
artursmirnov: This suggestion is crucial as it may cause bugs

app/components/flash-display.js
```diff
@@ -1,15 +1,21 @@
...
+    const classes = Array.from(new Set(this.messages.map(message => message.type))).join(' ');
```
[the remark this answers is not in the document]
artursmirnov: This request is optional, I'll leave it up to you.

app/models/owner.js
```diff
@@ -32,6 +32,8 @@ export default VcsEntity.extend({
   isUser: equal('type', 'user'),
   isOrganization: equal('type', 'organization'),
 
+  allowance: attr(),
```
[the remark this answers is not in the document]
artursmirnov: This is optional, you can keep it as is for now

app/components/profile-nav.js
```diff
@@ -61,4 +61,21 @@ export default Component.extend({
     return !this.features.get('enterpriseVersion') && !!billingEndpoint;
   }),
 
+  didRender() {
+    const isUser = this.model.isUser;
+    const login = this.model.login;
+    const allowance = this.model.allowance;
+    const plansPath = isUser ? `https://travis-ci.com/account/${config.planSuffix}` : `https://travis-ci.com/organizations/${login}/${config.planSuffix}`;
+    const settingsPath = isUser ? `https://travis-ci.com/account/${config.settingsSuffix}` : `https://travis-ci.com/organizations/${login}/${config.settingsSuffix}`;
```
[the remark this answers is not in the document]
artursmirnov: Page reloads will significantly affect user experience. Please consider improving this part if possible.

artursmirnov:
@piotrmilcarz I have deleted all the comments except the crucial ones. Please consider fixing the rest.

piotrmilcarz:
Thank you @artursmirnov for your review. As always, your comments are very helpful.
I improved the code, but (unfortunately) there are still links in the 'warning' Flash message, as the simple string does not allow to put the action. It's possible to execute JS code in onClick handler, but it's pure JS, without Ember (and router) context.

artursmirnov:
Hey @piotrmilcarz, thank you for working on feedback! 

I'd like to suggest you trying the following approach to implement links in flashes properly:

1) There's a [`flash-display`](https://github.com/travis-ci/travis-web/blob/master/app/templates/components/flash-display.hbs) component that basically just [renders the first item from `flashes` service](https://github.com/travis-ci/travis-web/blob/master/app/components/flash-display.js#L11). It uses [`flash-item`](https://github.com/travis-ci/travis-web/blob/master/app/templates/components/flash-item.hbs) component to render a single message and passes `flash` data into that.

2) We can tweak the `flash-display` component to handle `custom` flashes. The `custom` flash is the one that accepts a custom component and a `data` object for that component. For example:

**/app/templates/components/flash-display.hbs**
```hbs
{{#let this.messages.firstObject as |flash|}}
  {{#if (eq flash.type 'custom')}}
    {{component flash.component @data=flash.data}}
  {{else if flash}}
    <FlashItem
      data-test-components-flash-item="true"
      @flash={{flash}}
      @close={{action "closeMessage"}}
    />
  {{/if}}
{{/let}}
```
In this case, when `flash.type === 'custom'` it'll render an arbitrary component, name of which is contained in the `flash.component` and will bypass the `flash.data` object to that component as `@data` attribute.

3) Now we can create an arbitrary component that will render our custom flash. For example:

**/app/templates/components/flashes/negative-ballance.hbs**
```hbs
<div class="flash-styles">
  <span>Please visit</span>
  <LinkTo @route={{@data.route}} @models={{@data.routeModels}}>this link</LinkTo>
  <span>to resolve the issue</span>
</div>
```

4) Now we also need to add the logic for handling this new type of messages to `flashes` service. For example:

**/app/services/flashes.js**
```js
export default Service.extend({
  ...
  custom(component, data = {}) {
    assert("Component name is mandatory for custom flashes", !!component);
    const flash = { component, data, type: 'custom' };
    this.flashes.unshiftObject(flash);
  }
  ...
});
```

5) And now we can use our new `custom` flash like this:

**anywhere in the app**
```js
this.flashes.custom("flashes/negative-ballance", { route: 'account/subscription' });
```

There will likely be some more things to add in order to wire everything up to make it work, but the idea should be clear.

What do you think about it?

app/components/flash-display.js
```diff
@@ -1,15 +1,23 @@
...
-  classNames: ['flash'],
+  classNameBindings: ['type'],
   tagName: 'ul',
 
   messages: alias('flashes.messages'),
 
+  type: computed('messages.@each.type', function () {
+    const classes = this.messages.map(message => message.type).reduce(
+      (unique, item) => (unique.includes(item) ? unique : [...unique, item]),
+      []).join(' ');
```
artursmirnov, line 17: You can use Ember's [`uniqBy`](https://api.emberjs.com/ember/3.12/classes/EmberArray/methods/uniqBy?anchor=uniqBy) and [`mapBy`](https://api.emberjs.com/ember/3.12/classes/EmberArray/methods/mapBy?anchor=mapBy) array extensions to simplify this logic:
```suggestion
    const classes = this.messages.uniqBy('type').mapBy('type').join(' ');
```

piotrmilcarz:
Many many thanks @artursmirnov for your constructive review 👏

app/services/flashes.js
```diff
@@ -61,11 +68,12 @@ export default Service.extend({
       const { message, preamble = messageTypeToPreamble[type], aboveOverlay } = flash[type];
       const icon = messageTypeToIcon[type];
       const closeButton = messageTypeToCloseButton[type];
+      const autoClose = messageTypeToAutoclose[type];
       const item = { type, message, icon, preamble, closeButton, aboveOverlay };
 
       this.flashes.unshiftObject(item);
 
-      if (!closeButton) this.removeFlash(item);
+      if (autoClose) this.removeFlash(item);
```
artursmirnov, line 76: This change seems to be irrelevant. Could you please revert it (and all related ones)?

app/components/flash-display.js
```diff
@@ -1,15 +1,21 @@
...
-  classNames: ['flash'],
+  classNameBindings: ['type'],
   tagName: 'ul',
 
   messages: alias('flashes.messages'),
 
+  type: computed('messages.@each.type', function () {
+    const classes = this.messages.uniqBy('type').mapBy('type').join(' ');
+    return `flash ${classes}`;
+  }),
```
artursmirnov, line 17: It seems like this change is not relevant anymore, because your custom flash will declare it's styles and appearance separately, in its own template. And all other old types shall work as previously. Could you please revert it (and all related changes)?

app/components/profile-nav.js
```diff
@@ -61,4 +63,17 @@ export default Component.extend({
...
+  didRender() {
+    const allowance = this.model.allowance;
+
+    if (!allowance || allowance.subscription_type !== 2)
+      return;
+
+    if (!allowance.private_repos) {
+      this.flashes.custom('flashes/negative-balance-private', { owner: this.model, isUser: this.model.isUser });
+    } else if (!allowance.public_repos) {
+      this.flashes.custom('flashes/negative-balance-public', { owner: this.model, isUser: this.model.isUser });
```
artursmirnov, line 75: It feels like this action also needs an opposite one. When user has a negative balance, this flash appears. User sees the message, goes to the Plan page, replenishes the balance, but still sees the message, because it never gets removed before page reload. 

Also, how would it work when user visits the profile section multiple times? Will it add multiple flashes? I mean, I know, that it will _render_ only the first one, but will they stack up there? Probably it would make sense to implement some kind of a simple management for these notifications (f.e. clear them when user leaves the page)?

app/components/repository-layout.js
```diff
@@ -36,5 +38,15 @@ export default Component.extend({
...
     }
+  },
+
+  didRender() {
+    const repo = this.get('repo');
+
+    if (!repo.canOwnerBuild && repo.isPrivate) {
+      this.flashes.custom('flashes/negative-balance-private', { owner: repo.owner, isUser: repo.ownerType === 'user' });
+    } else if (!repo.canOwnerBuild && !repo.isPrivate) {
+      this.flashes.custom('flashes/negative-balance-public', { owner: repo.owner, isUser: repo.ownerType === 'user' });
```
artursmirnov, line 49: The same goes here with regards to the notifications management. Except that here it may even lead to bugs. Like, when user visits a repository that belongs to the org with negative balance, the message appears, then user navigates to another repository that belongs to another org with a positive balance, but the message is still there, so it confuses the user. Does it make sense?

tests/acceptance/repo/allowance-test.js
```diff
@@ -0,0 +1,66 @@
...
+
+    this.server.create('job', { number: '1234.2', repository: repo, state: 'passed', config: { env: 'JANTS', os: 'osx', language: 'ruby', rvm: 2.2 }, commit, build });
+    this.server.create('job', { allow_failure: true, number: '1234.999', repository: repo, state: 'failed', config: { language: 'ruby', os: 'jorts' }, commit, build });
+  });
+
+  test('warning is displayed in case owner cannot build', async function (assert) {
+    await page.visit({ organization: 'user-login', repo: 'repository-name' });
+
+    assert.equal(page.flash, 'Builds have been temporarily disabled for public repositories due to a negative credit balance. Please go to the Plan page to replenish your credit balance or alter your OSS Credits consumption setting.');
+  });
```
artursmirnov, line 65: Could you please add the opposite case - that message is **not** displayed when balance is positive? 

Optionally, it would be really nice to have a test, which ensures that message disappears correctly when user navigates to another page (that doesn't suppose to have the message) or when balance is getting replenished. I think that these cases are also important, but I'll leave it up to you to decide if there's a capacity to implement them.

artursmirnov (changes requested):
Hey @piotrmilcarz, thank you for implementing the feedback, I appreciate it! I still have some comments on the code though, please take a look when you have a moment.

Also, could you please provide a short instruction how to test these changes visually? I'd like to try it in action and test how it looks and works. Thank you!

app/components/repository-layout.js
piotrmilcarz: Thank you for noticing this potential behavior. I added cleanup for Flashes.

app/components/flash-display.js
piotrmilcarz: We still need this class for an additional margin for `warning` Flash (as it's not closable) at the highest level of Flash DOM (outer ul element), but I refactored it and added an optional className for the Flash object.

app/styles/app/modules/flash.scss
```diff
@@ -86,6 +86,21 @@
...
+      color: inherit;
+      border-bottom: 1px solid;
+    }
+  }
+}
+
+ul.flash.custom+.wrapper-main,
+ul.flash.custom+#main,
+ul.flash.custom+.wrapper-main+#left {
+  margin-top: 55px;
```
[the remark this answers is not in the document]
piotrmilcarz: Corrected.

app/services/flashes.js
piotrmilcarz: 👍

app/components/repository-layout.js
```diff
@@ -5,6 +5,8 @@ import { inject as service } from '@ember/service';
 export default Component.extend({
   externalLinks: service(),
   features: service(),
+  flashes: service(),
+  router: service(),
```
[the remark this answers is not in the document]
piotrmilcarz: Nice catch! 👍

piotrmilcarz:
@artursmirnov can you do me a favor and verify Percy screenshots? Unfortunately I don't have access :(

artursmirnov:
Hey @piotrmilcarz, sure! Percy looks good now, thank you for fixing the regression 👍

artursmirnov:
@piotrmilcarz also what would be the workaround to test these changes on `test-deployments`?

artursmirnov (commented):
Code-wise it looks good to me. I'd like to try it in action before approving the PR though. @piotrmilcarz thank you for working on feedback!

shairyar:
Lets deploy it so we can test it out.

piotrmilcarz:
@artursmirnov 
API and Billing are deployed now.
'piotrmilcarz/travis-web' - is an example repo with an exhausted v2subscription - mine :-)

artursmirnov:
@shairyar it is deployed: https://negative_balance_notification.test-deployments.travis-ci.com/ Although, I thought, there might be some more setup required.

With regards to the current deployment, I have a couple of questions/issues:

1) Currently the message is always visible for any repository. As far as I know, there will certain amount of time, during which we would have to maintain both billing models. And for this period, the system should handle old plans properly, not showing the message for them. @piotrmilcarz will this case be handled separately?

2) The code for both repository and profile views looks similar to me, but the message shows up only on repository page, nothing appears on the account section. Is it expected in this revision?

3) Would it make sense to add just a stub route for `/account/plan` page, so that the link didn't break the app?

4) I'd like to suggest adding an anchor for the settings page, so that it scroll to exact place (see how Insights settings work). I understand that settings are not in place yet, but we can just add an anchor now and settings later. My concern here is that we'll forget about it when we'll get to settings work. What do you think?

5) The message looks bad on narrow screens:

image

It probably needs some more spacing.

6) There's a tiny issue with the tooltip that's getting overlapped by the notice. Could you please fix that?

Otherwise it looks very good to me, thank you @piotrmilcarz for your hard work on this feature!

artursmirnov:
@piotrmilcarz thanks for deploying the backend! I'll push this branch to staging as well so that we could check everything at once 👍 And probably it'll answer some of the questions from my previous comment 😄

artursmirnov:
The branch is deployed to staging: https://staging.travis-ci.com/github/piotrmilcarz/travis-web 

Expectedly, the question N1 is resolved, works well, thank you @piotrmilcarz 👍 

WRT question N2 - I wasn't able to test is on staging, as I don't have neither push nor admin access to the test account (obviously 😄), but that's okay, I'll keep it up to you, just wanted to bring your attention to this topic. 

All other questions are still effective.

piotrmilcarz:
@artursmirnov 
1. Notification is displayed if repo owner has subscription type === 2 && is not allowed to build (private or public). If you switch repository then the message is removed.
Screenshot 2020-09-07 at 13 48 44
Screenshot 2020-09-07 at 13 49 46

2. Account page (pages) shows allowance in your (personal/user) context or organization's context. If your user does not have v2 subscription then no message will be there ;-)

3. `plan` route is already defined in the router, however, the Plan page will be delivered in separate PR.

4. Great idea, anchor added.

5 + 6. Thank you for catching this. Fixed.

artursmirnov (approved):
@piotrmilcarz thank you for the explanation, it totally makes sense! Also, your fixes look good to me, thank you 👍 Is the PR ready for merge?

piotrmilcarz:
Thank you @artursmirnov for the detailed review.
The PR is ready for the merge to epic-pricing.

piotrmilcarz:
🎉

PR #2532, opened 2020-08-25
Update VAT field display according to new requirements

artursmirnov:
Ticket: https://github.com/travis-pro/billing-v2/issues/588


shairyar (approved):
Thank you

app/templates/components/billing/information.hbs
```diff
@@ -112,29 +112,62 @@
...
-          @onChange={{action (mut newSubscription.billingInfo.vatId)}} 
-          as |Field|
-        >
-          <Field.input @name='vat'/>
-        </form.field>
-      </div>
+      {{#if this.showNonZeroVatConfirmation}}
+        <div class='flex justify-between items-center mb-16'>
+          <div class='w-4/6 pr-2'>
+            Is your company registered locally for VAT/GST?
```
shairyar, line 130: @artur can we please sending the option selected by the customer for this question in the payload as well because in the billing service we need to place some validations based on what customer selected.

artursmirnov: @shairyar yes, that's exactly what I was going to ask you about :) Could you please give me a little bit of context with regards to how/where this flag should be sent? Thanks!

shairyar: You can send this as part of a subscription payload for example when someone submits the form when creating a subscription.

artursmirnov: Probably we can add a new attribute to existing [billing-info](https://github.com/travis-ci/travis-web/blob/master/app/models/billing-info.js) model and call it something like `has-local-registration: boolean`. How does it sound to you?

artursmirnov: Hey @shairyar, I have just added the discussed attribute to the billing info (it should be sent over with subscription data). Please take a look if it works.

shairyar: Thank you, I will have a look at this today and get back to you.

shairyar (changes requested):
After making this change, the following case is not working anymore on web

> We need to have a mandatory question: “Is your company registered locally for VAT/GST”
> If reply is “Yes” – customer needs to complete mandatory field “Tax ID”

Earlier a vat field used to appear when customer would select Yes, now it is not appearing any more.

Another thing I noticed was when you are updating the billing details, this question does not appear, it is required on a new subscription form and on updating billing details form as well.

artursmirnov:
Hey @shairyar, thank you for the feedback! It should be implemented on the latest deployment. Could you please take another look when you have a chance and let me know what you think? Thank you!

shairyar (changes requested):
Hey @artursmirnov this time the form validations are not working properly, even though the data has been entered for example First Name and Last Name, the form still says the field is required and I am not able to proceed :)

artursmirnov:
Hey @shairyar, thank you for checking! Yeah, that's why tests have failed as well, fixing it now. I'm sorry for pinging you until everything works, but does it _look_ right now?

shairyar (approved):

artursmirnov:
@shairyar are API changes already released? Shall I merge and release this PR as well?

shairyar:
Let’s wait till Monday, I will deploy the backend changes and let you know

PR #2535, opened 2020-09-07
Changes for V2 Plan summary and change pages

AndriiMysko:


AndriiMysko:
Hello, @artursmirnov please take a look

artursmirnov (changes requested):
Hey @AndriiMysko, thank you for your contribution! It's a big PR 😄 

The code generally looks very good to me. Great job! Although, I still have some comments. They are mostly minor and should be easy to address. Please take a look when you have time. And please don't hesitate to ask if you have any questions or problems. I'd be happy to help!

artursmirnov:
Hey @AndriiMysko, what would be the best way to try the changes in action? I've tried to run the branch locally (with production API) and there are (expectedly) some failing API requests to plans endpoints. I can deploy your API changes to staging and use that as a backend for local environment, but would that be sufficient? Or are there any other apps that need to be deployed as well in order to test the UI? Maybe any other setup? Thanks!

AndriiMysko:
Hello @artursmirnov 
Both API and billing-v2 changes are needed to run this. Other than that, nothing else. I will commit the fixes for most of your comments in a bit as well.

artursmirnov:
I have deployed the branch to http://staging.travis-ci.com/ for testing. Everything seems to be working there, but when I merge `epic-pricing` branch into it, it stops working (probably, because something was broken at `epic-pricing` itself).

Anyway, I have noticed a few small visual issues:

1) On the plans selector view the tooltip and its icon are not aligned properly:
image

2) Clicking on "Select plan" button does nothing until the box is selected by an outside click:
image

3) Clicking on "Learn more" link leads to `/account/test` route, which doesn't exist.
image

4) For an account with old subscription (no subscription) the page appears incorrectly - it shows the trial offer (which is expected) and the new plan selector below. Doesn't seem expected. Although, after selecting a new plan, there's no way back to the old subscription view and the issue never appears after that. It's only for the old subscription.
image

5) Not sure if that's just a configuration issue, but this looks a bit awkward:
image

6) Is it expected that invoice history is displayed on the select plan view?
image

Everything else looks very good to me, thank you!

AndriiMysko:
Hello, @artursmirnov 
Looks like it hasn't been decided yet what the link for the "Learn more" will be, so for now I have left it to point back to the `plan` route. As for other comments, I've tried to fix all of them at this point.

app/models/owner.js
```diff
@@ -109,6 +109,29 @@ export default VcsEntity.extend({
...
+          id: plan.id,
+          name: plan.name,
+          startingPrice: plan.starting_price,
+          startingUsers: plan.starting_users,
+          privateCredits: plan.private_credits,
+          publicCredits: plan.public_credits,
+        });
+        return planConfig;
+      });
+    }
```
artursmirnov, line 128: Would it make sense to delegate fetching and extraction to Ember Data? It should be easy to achieve. Just overwrite the `query` hook in `v2-plan-config` adapter, make it accept the type and id parameters and build the request URL based on them. Something like this:
```js
  buildURL: function (modelName, id, snapshot, requestType, query) {
    const { type, id } = query;
    const isOrganization = type === 'organization';
    const isUser = type === 'user';

    assert("The plan config adapter only supports a query request type", requestType === 'query')
    assert("Invalid request parameters for plan config adapter", isUser || isOrganization && !!id);

    const path = isOrganization ? `v2_plans_for/organization/${id}` : 'v2_plans_for/user';
    return `${this.urlPrefix()}/${path}`
  },

  // This overrides the parent implementation to ignore the query parameters
  query(store, type, query) {
    let url = this.buildURL(type.modelName, null, null, 'query', query);
    return this.ajax(url, 'GET');
  }
});
```

Then Ember Data should extract the data properly using the built-in serializer and store it to the store.

After that, fetching the plan configs would look like this:

```js
const { id, type } = this; // owner properties
const plans = yield this.store.query('v2-plan-config', { type, id })
```

How does it sound to you?

AndriiMysko: I see, this is indeed much cleaner. Thank you for the comment!

app/templates/components/billing/select-plan.hbs
```diff
@@ -1,79 +1,97 @@
...
+                    Select plan
+                  </button>
+                {{else}}
+                  <button data-test-subscribe-button
+                    onclick={{action form.submit}}
+                    class='button--blue billing-plans__box-v2--button'
+                  >
+                    Select plan
+                  </button>
+                {{/if}}
```
artursmirnov, line 73: Now, when clicking on the button, it selects the plan on the first click and then advances to the ordering process on the second click to the same button. This two-clicks behavior feels excessive and unnecessary to me. Would it be better to both select the plan and advance to ordering in a single click? Otherwise we can change the button title for selected plan - something like "Proceed to ordering". In this case it will feel more like a confirmation, rather than something didn't work. Although, my vote is still for the single click approach, as that's what I would expect by clicking on the button. What do you think?

AndriiMysko: I'm not entirely sure how to do these two actions at once in this case.
Would this be acceptable?
```JS
<button data-test-subscribe-button
  {{action (mut this.selectedPlan) plan on='mouseDown'}}
  onclick={{action form.submit}}
  class='button--white-and-teal billing-plans__box-v2--button'
>
  Select plan
</button>
```

artursmirnov: I think the most straight forward way would be to define an action on component:

```js
  ...
  actions: {
    selectAndSubmit(plan, form) {
      this.set('selectedPlan', plan');
      form.submit();
    }
  }
  ...
})
```
And then bind it on click in template as usual.

You can also add a short delay with `later(() => form.submit(), delay)` for a visual effect if you'd like :)

How does it sound to you?

artursmirnov (changes requested):
Hey @AndriiMysko, thank you for addressing all the feedback! Code-wise it looks great to me, good job! 👍 Just a few small things left.

I have updated https://staging.travis-ci.com with the latest version of this branch. Everything looks and works good to me as well, with the exception of the "Select plan" button issue, described above.

AndriiMysko:
Hey @artursmirnov, thank you for the very detailed and helpful comments! Hopefully I have fixed everything with the last commit.

app/adapters/invoice.js
```diff
@@ -2,9 +2,11 @@ import V3Adapter from 'travis/adapters/v3';
 
 export default V3Adapter.extend({
   buildURL: function (modelName, id, snapshot, requestType, query) {
-    if (requestType === 'query' && query.subscription_id) {
+    const { type, subscriptionId } = query;
+    const route = type === 2 ? 'v2_subscription' : 'subscription';
```
artursmirnov, line 6: Wow, I really like this change! 👍

artursmirnov (dismissed):
@AndriiMysko thanks for your hard work on the feedback, I appreciate it! I have nothing more to add code-wise. It looks great to me ✨ 

Visually, I didn't find any issues, looks very good and solid to me. 

I have updated https://staging.travis-ci.com with latest changes. @shairyar please take a look as well and verify if everything is working as intended by the PRD.

artursmirnov:
I also have this case caught by Percy:
image

Previously there was a plan selector for expired GH marketplace subscription. Although, I'm not sure if it applies to the new system. @shairyar WDYT?

shairyar:
Thanks for point this out @artursmirnov this applies to the new plans as well. If I am not wrong, in case GH trial/plan expires, the plan users see subscribes them directly on Travis CI rather than GH, can you please confirm?

AndriiMysko:
@artursmirnov I have adjusted the plan selector for expired GH marketplace trial, please take a look

artursmirnov:
@shairyar according to the screenshot, users with expired GH marketplace trial/subscription have a bold button that suggests them to "Resubscribe or change plan" on GH marketplace, not directly at Travis. Although, they also have Travis plans listed below, so they have this option. The button leads to [GH marketplace](https://github.com/travis-ci/travis-web/blob/master/config/environment.js#L276)

shairyar:
Right, thanks. The new plans need to work the same way.

artursmirnov:
@AndriiMysko thanks for the quick fix, I have just checked at Percy and everything seems to be working as expected:

image

/cc @shairyar

shairyar (approved):

artursmirnov:
Alright, it seems that this PR is good to be merged to the epic branch. @AndriiMysko is there anything from your side or shall I go ahead and merge it?

AndriiMysko:
@artursmirnov as of right now there's nothing, please go ahead and merge it. Thank you for all the comments!

artursmirnov:
Awesome! @AndriiMysko and thank **you** for your hard work on this PR 👍

PR #2542, opened 2020-09-16
Separate addon purchase

pavel-d:


pavel-d:
@artursmirnov, could you please take a look? I will push tests a bit later

app/styles/app/layouts/billing.scss
```diff
@@ -155,6 +155,108 @@
...
+      height: 114px;
+      flex-basis: auto;
+      margin-right: 5px;
+
+      &--desc {
+        color: $asphalt-grey;
+        font-weight: 600;
+        font-size: 0.8em;
+        margin: 5px 0 12px 0;
+        color: #9d9d9d;
```
artursmirnov, line 174: This color property overwrites the one above. Is it on purpose? Also it'd be better to reference colors via variables ([cement-grey](https://github.com/travis-ci/travis-web/blob/master/app/styles/app/vars.scss#L22) in this case), because they are more meaningful and this allows to better understand what color stands behind it.

app/styles/app/layouts/billing.scss
```diff
@@ -155,6 +155,108 @@
...
+
+        span {
+          color: $cement-grey;
+          font-size: .5em;
+        }
+      }
+
+      &--name {
+        margin: 0px 0 10px 0;
+        color: #666666;
```
artursmirnov, line 206: This is [asphalt-grey](https://github.com/travis-ci/travis-web/blob/master/app/styles/app/vars.scss#L24) color.

app/templates/components/billing/select-addon.hbs
```diff
@@ -0,0 +1,53 @@
...
+      </div>
+    </div>
+    <div>
+      <span class='billing-plans__addons_box--desc'>Total:</span>
+      <span class='billing-plans__addons_box--price'>
+        {{format-currency this.totalPrice floor="true"}}
+      </span>
+    </div>
+    <div class="flex flex--wrap {{if this.account.hasSubscriptionPermissions 'billing-subscription__buttons'}}">
+      {{#if this.isLoading}}
```
artursmirnov, line 37: I didn't find any declaration of `isLoading` on the component class. Is it expected? According to the logic I can assume that it could likely be `this.save.isRunning`.

tests/integration/components/billing/selected-addon-test.js
```diff
@@ -0,0 +1,35 @@
...
+    const addonConfig = {
+      id: 'credits_500k',
+      name: '500 000 credits (50k Linux build minutes)',
+      price: 30000,
+      quantity: 500000,
+      type: 'credit_private'
+    };
+
+    this.selectedAddon = addonConfig;
+    this.set('selectedAddon', addonConfig);
```
artursmirnov, line 25: I don't think this duplication is needed:
```suggestion
    this.set('selectedAddon', addonConfig);
```

artursmirnov (changes requested):
Hey @pavel-d, thank you for your work on this! Code-wise it looks very good to me, I have just a few very minor comments. Could you please take a look when you have a moment?

Also it would be nice to have a very basic acceptance test that would check that addon boxes appear when they should, have required content and also that they _don't_ appear when they shouldn't. It doesn't have to be a standalone test, you can just add cases to existing ones if you'd like.

app/styles/app/layouts/billing.scss
pavel-d: Oops, right. I've just copied values from zeplin as is, will change to variables.

artursmirnov (changes requested):
Hey @pavel-d, thank you for implementing all of the feedback! The code looks perfect to me 👍 

I have a few comments related to the visual part:

1) Buttons are a bit misaligned:
image

image

2) The wording on "Buy Addons" view is a little bit confusing:
image

It looks like it's possible to buy multiple addons at once, but selector allows to select only one.

I assume that it's implemented according to the PRD, just wanted to bring your attention to this topic to make sure it's by design and not any kind of misunderstanding.

Everything else looks perfect!

artursmirnov (approved):
Thank you @pavel-d, looks great to me!

PR #2544, opened 2020-09-17
Owner allowance for repository context - refactored

piotrmilcarz:


app/models/repo.js
```diff
@@ -68,7 +69,23 @@ const Repo = VcsEntity.extend({
...
+      return this.owner.allowance;
+    if (this.repoOwnerAllowance)
+      return this.repoOwnerAllowance;
+    this.fetchRepoOwnerAllowance.perform();
+  }),
+
+  repoOwnerAllowance: reads('fetchRepoOwnerAllowance.lastSuccessful.value'),
+
+  fetchRepoOwnerAllowance: task(function* () {
+    const allowance = this.store.peekRecord('allowance', this.owner.id);
```
artursmirnov, line 83: `peekRecord` takes a model name and record id as arguments. And id assumed to be of the same type as the model name. Therefore, `peekRecord` with this implementation will return an `allowance` record, which `id` is equal to `this.owner.id`, which seems wrong to me. `owner` and `allowance` have different `id`s, aren't they? 
According to the current data structures, I'd suggest the following:
```suggestion
    const allowance = this.store.peekAll('allowance ').findBy('owner.id', this.owner.id);
```

app/models/repo.js
```diff
@@ -68,7 +69,23 @@ const Repo = VcsEntity.extend({
...
+    if (allowance)
+      return allowance;
+    yield this.store.queryRecord('allowance', { login: this.owner.login, provider: this.provider });
```
artursmirnov, line 86: The `yield`ed value needs to be returned in order to get set as task's value:
```suggestion
    return yield this.store.queryRecord('allowance', { login: this.owner.login, provider: this.provider });
```
Currently, the `fetchRepoOwnerAllowance.lastSuccessful.value` will be undefined after fetch.

app/models/repo.js
```diff
@@ -80,12 +97,16 @@ const Repo = VcsEntity.extend({
     const allowance = this.allowance;
     const isPrivate = this.private;
 
-    if (allowance && allowance.subscription_type === 1)
+    if (allowance && allowance.subscriptionType === 1)
       return true;
-    if (!allowance)
+    if (!allowance && !this.repoOwnerAllowance) {
+      return true; // pending allowance call
```
artursmirnov, line 103: It seems to me that this should return `false`. Until we know exactly that user is allowed to build, we shouldn't allow them to do so. Let's say the allowance request hung for a while and user had a chance to trigger a build (for example, via Restart) - they will likely get an error from API. If it's necessary to have this data already fetched at this point, you can trigger the task run in corresponding route - `beforeModel`, `model` and `afterModel` hooks are blocking view rendering until their returned promises are resolved.
Does it make sense to you?

mirage/config.js
```diff
@@ -516,6 +516,16 @@ export default function () {
     }
   });
 
+  this.get('/owner/:provider/:login/allowance', function (schema, request) {
+    let owner = schema.users.where({ login: request.params.login }).models[0];
```
artursmirnov, line 520: This won't work for organizations. Is it expected?

tests/acceptance/repo/allowance-test.js
```diff
@@ -14,76 +14,83 @@ module('Acceptance | repo allowance', function (hooks) {
...
+      concurrency_limit: 666
+    }),
+
+    this.server.create('allowance', {
+      id: 3,
+      subscription_type: 2,
+      public_repos: true,
+      private_repos: false,
+      concurrency_limit: 2
+    }),
```
artursmirnov, line 55: You shouldn't specify `id`s manually, Mirage's factories will generate them automatically and take care of uniqueness.

tests/acceptance/repo/trigger-build-test.js
```diff
@@ -14,6 +14,14 @@ module('Acceptance | repo/trigger build', function (hooks) {
   setupMirage(hooks);
 
   hooks.beforeEach(function () {
+    this.server.create('allowance', {
+      id: 1,
```
artursmirnov, line 18: Same here with `id`s

tests/acceptance/repo/view-migrated-test.js
```diff
@@ -19,13 +19,25 @@ module('Acceptance | repo/view migrated', function (hooks) {
 
     await signInUser(user);
 
+    this.server.create('allowance', {
+      id: 1,
```
artursmirnov, line 23: And here

artursmirnov (changes requested):
Hey @piotrmilcarz, nicely done! Thank you for this improvement, it looks really good to me 👍 I have just a few comments, please take a look when you have a chance.

app/models/repo.js
piotrmilcarz: Api provides: allowance.id === owner.id
so should be right :-)

app/models/repo.js
piotrmilcarz: great hint! `beforeModel` seems to be created for such purposes

mirage/config.js
piotrmilcarz: right, I extended it for looking for orgs as well.

app/models/repo.js
artursmirnov: Yes, it definitely is :)

tests/acceptance/repo/allowance-test.js
artursmirnov: This comment isn't fixed. Although, let's keep it as is for now. It seems to be working with generated ids too, so this is basically just a confusing thing. I'd appreciate if you could clean this up sometime later, don't want to hold this PR because of it.

tests/acceptance/repo/trigger-build-test.js
artursmirnov: This too

artursmirnov (approved):

PR #2545, opened 2020-09-18
Fix for addon purchase process

pavel-d:
if user tries to change plan after canceling addon purchase, previously selected addon is displayed on the order summary page


artursmirnov (approved):
Pretty straight forward, thank you @pavel-d for catching and fixing this issue 👍

PR #2546, opened 2020-09-18
Negative balance fix

piotrmilcarz:


tests/acceptance/repo/settings-test.js
```diff
@@ -26,13 +26,16 @@ module('Acceptance | repo settings', function (hooks) {
       login: 'org-login',
     });
 
+    this.server.create('allowance', {subscription_type: 1});
+    this.server.create('allowance', {subscription_type: 1});
```
artursmirnov, line 30: Is it a duplication? If it's intended, you could use
```suggestion
    this.server.createList('allowance', 2, {subscription_type: 1});
```
Does it make sense?

piotrmilcarz: It's not a duplication. Here I need 2 records of allowance. Thanks for hint level pro! :-)

artursmirnov (approved):

PR #2555, opened 2020-10-12
Poll subscription on plan change

AndriiMysko:
Hello @artursmirnov , can you please take a look at this?

### DIFF
--- a/app/components/billing/payment.js
+++ b/app/components/billing/payment.js
@@ -5,7 +5,6 @@ import { or, not, reads } from '@ember/object/computed';
 import { computed } from '@ember/object';
 import { typeOf } from '@ember/utils';
 import config from 'travis/config/environment';
-import Ember from 'ember';
 
 export default Component.extend({
   stripe: service(),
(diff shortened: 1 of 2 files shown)


app/components/billing/payment.js
```diff
@@ -85,7 +86,11 @@ export default Component.extend({
       } else {
         yield this.subscription.changePlan.perform(this.selectedPlan.id);
       }
-      yield this.accounts.fetchV2Subscriptions.perform();
+      if (!Ember.testing) {
+        yield this.pollSubscription.perform();
+      } else {
+        yield this.accounts.fetchV2Subscriptions.perform();
+      }
```
artursmirnov, line 93: This doesn't seem necessary to me, because `pollSubscription` task basically does the same. Although, this differentiation in tests makes them unreliable, because they may pass when actual behavior fails. Does it make sense?

artursmirnov (approved):
Looks good to me, thank you @AndriiMysko

PR #2556, opened 2020-10-12
V2subscription fixes

piotrmilcarz:


tests/integration/components/billing/warning-message-test.js
```diff
@@ -55,7 +55,7 @@ module('Integration | Component | warning-message', function (hooks) {
       name: 'Startup',
       startingPrice: 3000,
       startingUsers: 10,
-      addon_configs: [
+      addonCsonfigs: [
```
artursmirnov, line 58: Looks like a typo to me. Why didn't it break anything?

artursmirnov (changes requested):

tests/integration/components/billing/warning-message-test.js
piotrmilcarz: Yep, it's a typo. In this `other addons` test probably not used.

artursmirnov: If it's not used, would it make sense to remove it then?

piotrmilcarz: Indeed, there is a sense.

artursmirnov (approved):

PR #2558, opened 2020-10-13
Sign up message fix

piotrmilcarz:


app/models/v2-subscription.js
```diff
@@ -42,33 +42,27 @@ export default Model.extend({
...
-          processed.usedCredits += addon.current_usage.addon_usage;
-          processed.remainingCredits += addon.current_usage.remaining;
-        }
+    const addonReduce = (type) => (processed, addon) => {
+      if (addon.type === type) {
+        processed.totalCredits += addon.current_usage.addon_quantity;
+        processed.usedCredits += addon.current_usage.addon_usage;
+        processed.remainingCredits += addon.current_usage.remaining;
+        processed.validDate = Date.parse(addon.current_usage.valid_to);
+        processed.purchaseDate = Date.parse(addon.current_usage.purchase_date);
```
artursmirnov, line 51: The concern here is that this data isn't aggregated, each previous addon data will be overwritten by the subsequent one. Is this expected?

AndriiMysko: Hello, the fix that was committed just now is the expected behaviour.

artursmirnov (approved):

PR #2561, opened 2020-10-23
SCA card payment fix

AndriiMysko:
Hello @artursmirnov 
Just a quick fix for the SCA card payments, please take a look.

P.S. pollSubscription removal is intentional, it is no longer needed

### DIFF
--- a/app/components/billing/authorization.js
+++ b/app/components/billing/authorization.js
@@ -46,7 +46,8 @@ export default Component.extend({
 
   retryAuthorization: task(function* () {
     if (this.requiresSourceAction && this.retryAuthorizationClientSecret) {
-      yield this.stripe.handleStripePayment.perform(this.retryAuthorizationClientSecret);
+      this.stripe.handleStripePayment.perform(this.retryAuthorizationClientSecret);
+      yield this.accounts.fetchV2Subscriptions.perform();
     }
   }).drop(),
 
@@ -61,6 +62,7 @@ export default Component.extend({
         });
         const { client_secret: clientSecret } = yield this.subscription.chargeUnpaidInvoices.perform();
         yield this.stripe.handleStripePayment.perform(clientSecret);
+        yield this.accounts.fetchV2Subscriptions.perform();
       }
     } catch (error) {
       this.flashes.error('An error occurred when creating your subscription. Please try again.');
--- a/app/components/billing/payment.js
+++ b/app/components/billing/payment.js
@@ -1,5 +1,5 @@
 import Component from '@ember/component';
-import { task, timeout } from 'ember-concurrency';
+import { task } from 'ember-concurrency';
 import { inject as service } from '@ember/service';
 import { or, not, reads } from '@ember/object/computed';
 import { computed } from '@ember/object';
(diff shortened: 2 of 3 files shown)


app/components/billing/authorization.js
```diff
@@ -46,7 +46,8 @@ export default Component.extend({
 
   retryAuthorization: task(function* () {
     if (this.requiresSourceAction && this.retryAuthorizationClientSecret) {
-      yield this.stripe.handleStripePayment.perform(this.retryAuthorizationClientSecret);
+      this.stripe.handleStripePayment.perform(this.retryAuthorizationClientSecret);
```
artursmirnov, line 49: It feels to me that `yield` is still required here to wait for the payment task to perform, isn't it?

app/components/billing/payment.js
```diff
@@ -133,7 +133,8 @@ export default Component.extend({
             token: token.id,
             lastDigits: token.card.last4
           });
-          yield subscription.save();
+          const { clientSecret } = yield subscription.save();
+          this.stripe.handleStripePayment.perform(clientSecret);
```
artursmirnov, line 137: Same here, `yield` seems to be required here.

app/models/v2-subscription.js
```diff
@@ -22,7 +22,7 @@ export default Model.extend({
   organization: belongsTo('organization'),
   coupon: attr('string'),
   clientSecret: attr('string'),
-  paymentIntent: attr('string'),
+  paymentIntent: attr(),
```
artursmirnov, line 25: Is this change intentional? What's the type of `paymentIntent`?

artursmirnov (changes requested):

app/models/v2-subscription.js
AndriiMysko: This change is intentional, yes, it's an object

app/components/billing/payment.js
AndriiMysko: When using `yield` here, subscription does not update after the payment is performed, while doing it this way does work.

artursmirnov: Would it make sense to investigate the root issue in this case? I'm afraid it could lead to some nasty bugs, because something seems to be wrong here.

sgolu6: @artursmirnov could you please clarify 'something seems to be wrong' ?
cc @shairyar

artursmirnov: @sgolu6 `handleStripePayment` is an asynchronous task that handles Stripe payment (sends request to Stripe and waits for response). This method is _aimed_ to run asynchronously and be awaited to get result. In the current implementation, the code just sends the payment request to Stripe and continues to subscription update without waiting for results from Stripe. If it doesn't work as intended, then there might be some kind of a hidden error that causes it. And as long as this is quite sensible part (handling the payment), I though, it wouldn't hurt to be careful with it.

shairyar (approved):

artursmirnov (approved):

app/components/billing/payment.js
artursmirnov: As discussed internally with @shairyar, this is acceptable and we can leave it this way for now.

shairyar: Thank you @artursmirnov

PR #2562, opened 2020-10-28
Plan migration fix

AndriiMysko:
Hello @artursmirnov !
A hotfix for migrating from a legacy plan to a new one.
I've also tried to fix the problem mentioned in the SCA fix, which turned out to be early task cancellation issue and is now working properly. Please take a look.

### DIFF
--- a/app/components/billing/account.js
+++ b/app/components/billing/account.js
@@ -23,6 +23,9 @@ export default Component.extend({
   isManual: bool('subscription.isManual'),
   isManaged: bool('subscription.managedSubscription'),
   isEducation: and('isSubscriptionsEmpty', 'isEducationalAccount'),
+  isSubscription: computed('isManaged', 'hasV2Subscription', 'isTrialProcessCompleted', 'isEduProcessCompleted', function () {
+    return (this.isManaged || this.hasV2Subscription) && this.isTrialProcessCompleted && this.isEduProcessCompleted;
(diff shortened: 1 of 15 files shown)


app/components/billing/account.js
```diff
@@ -31,6 +34,12 @@ export default Component.extend({
 
   showPlansSelector: false,
   showAddonsSelector: false,
+  isTrialProcessCompleted: computed(function () {
+    return !this.isTrial;
+  }),
+  isEduProcessCompleted: computed(function () {
+    return !this.isEducation;
+  }),
```
artursmirnov, line 42: Shouldn't these properties track changes on dependent ones? Or are they just initialization values?
```suggestion
  isTrialProcessCompleted: not('isTrial'),
  isEduProcessCompleted: not('isEducation'),
```

artursmirnov (approved):
Hey @AndriiMysko thank you for the fix and for revisiting the async behavior issue! It looks very good to me, I have just one little concern below. If that's intended, please let me know and I'll merge the PR 👍

app/components/billing/account.js
AndriiMysko: They are just initialisation values, that's correct!

artursmirnov: Thanks, I thought it would be the case 👍

PR #2563, opened 2020-10-29
Add more metrics tracking for new billing

AndriiMysko:


AndriiMysko:
Hello @artursmirnov !
Just a small addition of more metrics tracking for the new billing system, please take a look.

app/components/billing/payment.js
```diff
@@ -78,13 +78,26 @@ export default Component.extend({
 
   updatePlan: task(function* () {
     if (this.selectedPlan.isFree) {
+      this.metrics.trackEvent({
+        action: 'Plan Reverted to Free Plan',
+        category: 'Subscription',
+      });
```
artursmirnov, line 84: Can users cancel in the modal that this action opens? If yes, then this data may become inaccurate.

AndriiMysko: You're right, thank you!

artursmirnov (approved):
It looks great to me, thank you @AndriiMysko 👍

PR #2567, opened 2020-10-31
Fix text for tooltip on free plan

shairyar:
Fix text for tooltip on free plan


michal-at-travisci (commented):
Looks as requested, thank you

PR #2587, opened 2020-12-10
Permanently remove people section

weban:


Montana:
Possibly add `language support` in place of this? Thoughts @weban?

PR #2597, opened 2021-02-19
Trusted confirmed user account

makemp:


app/controllers/account/settings.js
```diff
@@ -56,6 +58,17 @@ export default Controller.extend({
     return repositories.filter(repo => !repo.emailSubscribed);
   }),
 
+  userHasNoEmails: computed('auth.currentUser.emails', function () {
+    if (!this.auth.currentUser.emails || this.auth.currentUser.emails.length === 0)
```
weban, line 62: why not return !this.auth.currentUser.emails || this.auth.currentUser.emails.length === 0

weban (changes requested):
Please check my comments

PR #2612, opened 2021-05-24
update plans page - Rich's request 

Montana:
this is on staging to take a peek beforehand.


travis-architect (approved):
Looks good but the test is failing

Montana:
In GreenKeeper (as I thought about 3 weeks ago) it says `plans-remove` branch is not listed. 

We can do:

```yaml
branches:
  only:
    - master
    - /^greenkeeper.*$/
  ```

GreenKeeper depending on the config will be scoped to `master` only, for example, what if I give the CI permission to access all branches, or whitelist branches with GreenKeeper in their name. For example, say if the project was using `yarn` and I had a `if` statement for `yarn`, then Travis would look at the `branches` which would involve GreenKeeper, it would look something like:

```yaml
after_success:
- 'if [ "$TRAVIS_PULL_REQUEST" = "false" ]; then yarn deploy; fi'
- 'if [ "$TRAVIS_PULL_REQUEST" = "false" ]; then yarn alias; fi'
branches:
  only:
  - master
  - "/^greenkeeper/.*$/"
  ```

- Montana Mendy

PR #2616, opened 2021-06-03
Hide Plan Usage tab from users without credits

AndriiMysko:


app/templates/components/profile-nav.hbs
```diff
@@ -161,7 +161,7 @@
               </LinkTo>
             </li>
           {{/if}}
-          {{#if this.showSubscriptionTab}}
+          {{#if this.showPlanUsageTab}}
```
travis-architect, line 164: In order to hide the Plan Usage tab for Hybrid plans, this check should also be applied on [line#129](https://github.com/travis-ci/travis-web/blob/0c0dd341003e444c0eca05b9809ce85e8cfcf079/app/templates/components/profile-nav.hbs#L129)

travis-architect (changes requested):

app/templates/components/profile-nav.hbs
AndriiMysko: Missed that, sorry

travis-architect (approved):

PR #2656, opened 2022-02-10
Travis ci spotlight fusioncharts Component

chinmaychaudhari-accolite:
- Added Fusioncharts Timeseries Component;
- API is called and integrated.
- Date Parameter is hardcoded needs to be changed when integrated with main component.


Sureshsangra19:
why are we removing package-lock.json file ?

chinmaychaudhari-accolite:
We did not remove the package.json.lock
It has been modified and need to be added

PR #2774, opened 2023-07-14
userroles prd

gbarc-dt:


murtaza-swati:
The build is failing and is not because of this PR, the build started to fail after merging the master in this branch. Needs to fix this project's build separately.

murtaza-swati (approved):

PR #2821, opened 2024-05-17
Build status img

nikolaTrichkovski:


app/services/status-images.js
```diff
@@ -27,12 +27,16 @@ export default Service.extend({
     }
 
     let slug = repo.get('slug');
-
+    let token = this.auth.assetToken;
```
murtaza-swati, line 30: @nikolaTrichkovski 
any specific reason to change the type of this variable from `const` to `let`?
I think lets use `const` for this and the following `branchQuery` variables as these are not getting reassigned, what do you think?

murtaza-swati (changes requested):

app/services/status-images.js
nikolaTrichkovski: you are right, this is an oversight. thanks for the comment

murtaza-swati (approved):

PR #2833, opened 2024-06-10
Organization settings access fix

nikolaTrichkovski:


DominikAlberski (approved):

app/components/profile-nav.js
```diff
@@ -69,9 +69,9 @@ export default Component.extend({
   hasPlanCreatePermissions: reads('model.permissions.plan_create'),
   hasBillingViewPermissions: reads('model.permissions.billing_view'),
   hasInvoicesViewPermissions: reads('model.permissions.plan_invoices'),
-  hasSettingsReadPermissions: reads('model.permissions.settings_read'),
+  hasSettingsReadPermissions: reads('model.permissions.settings_create'),
```
murtaza-swati, line 72: @nikolaTrichkovski the variable name is `hasSettingsReadPermissions` and its value is set to `reads('model.permissions.settings_create')` which might get confusing bus settings.read != settings.write. I think it would be better to rename the variable accordingly. Please make sure to change its usage everywhere it's used if you rename the variable.

murtaza-swati (changes requested):

app/components/profile-nav.js
nikolaTrichkovski: @murtaza-swati good point I will change the variable name

PR #2868, opened 2024-08-09
PDTQ-163 - [Pricing] travis-web - everywhere on plans page format number of credits to be more readable

murtaza-swati:
Formatted number of credits to separate thousands parts, e.g. 30 000 (many EU countries, US too) or 30,000 (EU countries/accounting formats)

changes made in:
- plan selection
- after plan is selected
- when showing pool of credits vs used credits on plan page
- addons selection page
- auto-refill boxes

### DIFF
--- a/app/helpers/format-number.js
+++ b/app/helpers/format-number.js
@@ -1,12 +1,16 @@
 import { helper } from '@ember/component/helper';
 
-export function formatCurrency(value) {
+export function formatNumber(value) {
   const parsedValue = Number(value);
   if (isNaN(parsedValue)) {
     return value;
   }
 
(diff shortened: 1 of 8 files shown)


DominikAlberski (approved):
I'm not convinced that we should make any localization efforts. From my experience, Travis is only in English, so we should decide on a preferred number format and use that consistently.

murtaza-swati:
> I'm not convinced that we should make any localization efforts. From my experience, Travis is only in English, so we should decide on a preferred number format and use that consistently.

we are not doing localization efforts here, we are only checking browser's language (like in chrome you can goto setting > language > add a language to top). so we are only covering that here.

PR #2941, opened 2024-12-26
Fix validto computed property infinite calls problem

murtaza-swati:
When subscription/v2-plan in DB has valid_to null, the /plan page make infinite calls to v_subscriptions and causing the page unresponsive.

The computed property is causing infinite calls because it depends on subscription.validTo, and within the computation, it makes a call to fetchV2Subscriptions.perform(), which likely modifies subscription.validTo. This causes the computed property to re-evaluate, leading to an infinite loop.

**Fixed Approach with an Observer:**
Ensure fetchV2Subscriptions doesn't modify subscription.validTo directly within the computed property, as this triggers recomputation.
Use computed properties for deriving state, and use observer to avoid side effect like fetching data.

### DIFF
--- a/app/components/billing/summary-v2.js
+++ b/app/components/billing/summary-v2.js
@@ -1,5 +1,5 @@
 import Component from '@ember/component';
-import { computed } from '@ember/object';
+import { computed, observer } from '@ember/object';
 import { reads, or, not, and, bool } from '@ember/object/computed';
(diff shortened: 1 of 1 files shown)


DominikAlberski (approved):

murtaza-swati:
closing in favour of https://github.com/travis-ci/travis-web/pull/2942

PR #2954, opened 2025-01-15
Tbt 137 account env vars

tonoyansergey-devt:
Pull Request contains changes related to recaptch Dockerfile that were needed for testing, before merge we need to remove these commits.

### DIFF
--- a/app/styles/app/modules/forms.scss
+++ b/app/styles/app/modules/forms.scss
@@ -166,9 +166,9 @@ textarea {
       display: flex;
       justify-content: left;
       width: auto;
-      height: 80px;
       align-items: center;
       margin-left: 1px;
+      height: 80px;
 
       button {
         order: 1;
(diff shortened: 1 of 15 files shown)


app/styles/app/modules/forms.scss
```diff
@@ -166,7 +166,6 @@ textarea {
       display: flex;
       justify-content: left;
       width: auto;
-      height: 80px;
```
gbarc-dt, line 169: won't it break something else?

gbarc-dt (commented):
tests!

app/styles/app/modules/forms.scss
tonoyansergey-devt: Returned it back.

murtaza-swati (approved):

DominikAlberski:
If staging tests are finished. Please make an interactive rebase on top of master and drop my commits in the process :-) thx!

PR #2958, opened 2025-01-30
sender usage updates

GbArc:


murtaza-swati (changes requested):
@GbArc does this still requires the internal user record in DB to have login/name present? Because as I understand, the main issue would still persist. From Michal comment on the ticket it is stated as:

"the description is empty? We have this 'special ids' known, we can even have short config in travis-web on how to put user names on them (eg cron job / triggered by bot )"

gbarc-dt:
Yes, you need the login/name in the database, there are already such users so why making more and more workarounds

murtaza-swati:
> Yes, you need the login/name in the database, there are already such users so why making more and more workarounds

previously I remember this bot user had a login present but now it's not there. so that's why we need some defaults here.

gbarc-dt:
> > Yes, you need the login/name in the database, there are already such users so why making more and more workarounds
> 
> previously I remember this bot user had a login present but now it's not there. so that's why we need some defaults here.

ok, if defaults for something not in db are required, i'll put it into api config

murtaza-swati (approved):

DominikAlberski (approved):

